PR #23, opened 2015-03-12
comma-fix

jzshen:
Added an extra argument to fetch_taxonomies to determine whether the page should show a comma (list = true) or not (list = false)


julianguyen:
Awesome!

Issue #17, opened 2015-02-28
more user-friendly in terms of viewings


jzshen:
currently the allies who are able to view your posts are all adjusted manually. this should be done automatically and users should be able to configure if they dont want certain or any users to see their post.

julianguyen:
This breaks viewing shared triggers on profile pages.

PR #25, opened 2015-03-29
Post types

jzshen:
Issue #17


spec/controllers/trigger_controller_spec.rb
```diff
@@ -62,5 +62,30 @@
 		  		expect(response).to redirect_to(trigger_path(other_new_trigger.id))
 			end
 		end
+
```
julianguyen: These tests look great, but they are really model tests. I would move it to the spec for trigger model! Also make sure the indentation is consistent.

jzshen: commited new changes

PR #26, opened 2015-05-01
Post types

jzshen:
Added changes with color-box and customization.
Process:
1) Click custom for viewers which renders colorbox.
2) Tick friends who you wish to have the post visible too
-In the back end: an attribute to that user will turn to true (view_permission)
3) Submit form and the attribute will default back to false


julianguyen:
@jzshen I don't recall whether you said this pull request isn't working. Can you update?

PR #29, opened 2015-05-10
Continuing CSS cleanup

juliepagano:
See commit messages for details about each change.

I think I caught and fixed all UI changes caused by adding the normalize.css, but it is possible I missed something because I'm new to the application. Please let me know if I missed anything.

Example up on heroku at:
http://enigmatic-tundra-5010.herokuapp.com/


julianguyen:
Looks good! Thanks Julie!

PR #34, opened 2015-07-03
Support groups

gCrew:
Added Calendar and Timepicker to meetings. Updated groups' show view to display the date of the meeting. There is a bug tho. Need to refresh the meetings page to use the timepicker.


julianguyen:
Will merge and take a look at the bug!

PR #43, opened 2015-07-21
Added sliding effects to subtitles and changed the color of "buttons"

gCrew:


julianguyen:
There is repetitive code! Please put this in one place or put the code into the respective js asset files!

gCrew:
@julianguyen for some reason, if I put the code only in application.js, the sliding effect works only the first time. Therefore, I had to put it in each file. I will put the code in js asset files instead tho

PR #48, opened 2015-09-24
Responsive header

emitche:
This makes the header and the text/pattern block beneath the header responsive.


emitche:
May be able to make these changes with Foundation classes. Going to look into that instead.

PR #61, opened 2015-11-15
Fix migration chain

jolim24601:
At Hacker Hours to contribute to my first open source 
project. Tried to migrate the db locally but ran into
the following error:

ERROR:  column "meetingid" of relation "meeting_members" already exists
STATEMENT:  ALTER TABLE "meeting_members" RENAME COLUMN "meetingid" TO "meetingid"

Deleted all but the most recent migration and replaced it
with the contents of the schema. 

Note: The schema version changed. We think this means
someone forgot to commit the schema at an earlier point.


jolim24601:
@julianguyen Hey check this out and let me know what you think!

julianguyen:
@jolim24601 Awesome work! Looks good to me! +1

julianguyen:
@jolim24601 If you're interested, you should submit a blurb about yourself in contributors.html.erb! Write about what mental health means to you and why you're helping out with if me :)

PR #69, opened 2015-11-24
Header adjustments

emitche:
Adjust header to account for smaller screens.


app/assets/javascripts/header.js
```diff
@@ -0,0 +1,60 @@
+$(document).ready(function(){
```
emitche: Moved header styles and scripts to their own separate files.

app/views/shared/_header.html.erb
```diff
@@ -1,5 +1,17 @@
 <div id="header">
```
emitche: Began addressing need for different ordering of header components based on screen size.

emitche:
I began revising the header to accommodate smaller screens. Some of this was making sure certain parts that were interactive would automatically be displayed instead on the smallest sizes.

Though I made progress, this does need polish.

PR #71, opened 2015-12-05
Update docs to include ifme_test DB in installation instructions.

wdhorton:
Following the installation instructions, when I ran rspec I got the following error: "FATAL:  database "ifme_test" does not exist (ActiveRecord::NoDatabaseError)." I saw that in installation we create an ifme_development database but not an ifme_test. I updated the README to include creating an ifme_test DB as well.


wdhorton:
Never mind, PR #72 is a better solution

PR #72, opened 2015-12-05
update readme

chenghw:
If the developer has not set up test environment and is receiving a fatal error. Added into readme the code to setup a test environment database.


julianguyen:
Thank you so much @chenghw! Awesome work!

PR #73, opened 2015-12-05
add annotate, rails-erd and better_errors to gemfile

danielglatstein:
added annotate to help make viewing tables attributes easier, rails-erd to see association, and better_errors for browser errors


julianguyen:
@danielglatstein Thank you! Awesome work :) :100:

PR #75, opened 2015-12-05
Delete empty README.rdoc.

wdhorton:
The README.rdoc file seems redundant given the documentation is in README.md. I just deleted the .rdoc file.


julianguyen:
Awesome catch @wdhorton

PR #78, opened 2015-12-05
removed deprecated @page_title variables from groups controller

reubenson:
Hi! Thanks for sharing ifme with the thoughtbot open source group! I enjoyed looking through your project, but did encounter difficulties contributing to the ifme project. In particular, I was confused why the forking the current repo generated javascript-related UI issues (unable to access drop-down tabs when navigating within the website; only works on page load), which did not happen on the live production site. Additionally, not using active record to do model associations makes it harder to understand how the codebase works. However, I was able to start making a small contribution to the groups controller be deleting the usage of the unnecessary/un-used instance variable, @page_title

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -6,7 +6,6 @@ class GroupsController < ApplicationController
   # GET /groups.json
   def index
     @groups = GroupMember.where(userid: current_user.id).all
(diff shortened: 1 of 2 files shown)


julianguyen:
@reubenson Issue you pointed with JS related UI issues has been fixed!

PR #79, opened 2015-12-05
Add ActiveRecord relations to Users and Allies

wdhorton:
I started to refactor the models to use ActiveRecord relations. To start, I renamed the old Ally model into Allyship (and the table from allies to allyships), since it tracks the relationship, and the allies are really instances of the User class. I got some ideas from this blog post: http://collectiveidea.com/blog/archives/2015/07/30/bi-directional-and-self-referential-associations-in-rails/

The tests still pass, and mostly all I had to do was update it with the new table, column, and model names.

It might seem like a lot of work for no improvement, but you can start to see the advantages in the refactor of the ApplicationController, where I reduced the method get_accepted_allies(userid) to:

`
def get_accepted_allies(userid)
  return User.find(userid).allies.where(status: AllyStatus::ACCEPTED)
end
`

If we can continue to refactor, you wouldn't necessarily even need these methods in ApplicationController at all, you could just call allies.where(status: AllyStatus::ACCEPTED) directly in the User model. It could really help simplify your fetch_taxonomies method in ApplicationController, which is really long right now.


wdhorton:
I realized that User.find(userid).allies.where(status: AllyStatus::ACCEPTED) wouldn't work, so I updated the User model to have an accepted_allies method and completely removed get_accepted_allies from ApplicationController

julianguyen:
Thanks @wdhorton for the fixes! Can you pull the latest from master? Thanks :)

fpcyan:
because I just went through this,

to update your fork, you pull from this repo then push it to your fork.

julianguyen:
:+1: @fpcyan

wdhorton:
It should be up-to-date now! @fpcyan thanks for the tip

julianguyen:
Does this break any unit tests related to allies?

PR #86, opened 2015-12-08
Refactoring allies

wdhorton:
This is a continuations of the refactor of allies (and now allyships) to use ActiveRecord relations. Overall I removed a lot of methods from ApplicationController and, when necessary, replaced them in the User model. I removed most of the references to UserRelations because they were only referenced in two places in the project, but if you want to add back some of those helper methods (like are_allies?, etc.) they should be placed on the User model (like User#is_ally_of?(other_user)).

In the process I also took out many of the conditions in the AlliesController #add and #remove methods, because things like adding and deleting the inverses of Allyships are now handled by the Allyships model itself.

Most importantly, all tests are passing.


fpcyan:
squash yr commits into 1 commit! Apparently that's a style thing for collaborations.

julianguyen:
@wdhorton Excellent changes! :100:  Thank you so much! Sorry I didn't get a chance to look at this PR earlier, been super busy!

PR #105, opened 2016-04-30
Update possible errors in README

cdeborja:
I was having issues getting bundle install to work with my Mac, so after looking through some sites, I saw  that other people were having issues with getting libv8 and therubyracer installed, so I added links to possible methods to help others install the gems


julianguyen:
Great catch @cdeborja ! Thanks!

PR #106, opened 2016-04-30
Add help for libv8 and therubyracer errors

lksevans12:
Hi! I ran into some errors when I bundled because I have El Capitan OS X 10.11.5 installed -- which is the most current version. Another one of my peers also ran into the same problem with 10.11.4, so we thought we would add where we found our solutions for these issues to the Readme. Hopefully other users will be able to benefit from these links. 

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,17 @@ rvm reinstall ruby
 rvm gemset pristine
 ```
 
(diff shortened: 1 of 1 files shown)


julianguyen:
Excellent work @lksevans12! :) thank you!

PR #108, opened 2016-04-30
Fixtextarea

lksevans12:
Hi! There was a bug with the ckeditor. Basically, when the page loads, the ckeditor didn't show up, the normal bootstrap text area showed up -- however, if you refresh it, the ckeditor would load. I was able to figure out that turbolinks was disrupting the javascript on the page, so thats why the ckeditor was only showing up on refresh. Originally, I wanted to remove all the turbolinks, but since I'm not familiar with the app, I don't know what else it would disrupt. I ended up adding a new JS page in the ckeditor file, which somewhat overrides the turbolinks for the added class. I also added ckeditor as a class to the textarea tags.


julianguyen:
@lksevans12 great catch! Thanks

PR #109, opened 2016-04-30
Fix textarea, not displaying 100% on small screens

nonothetoad:
Hey Julia, 
I fixed a UI bug where the textarea was not displaying 100% for screens sizes that are < 1020px. The reason why, was because the css rule to make textarea width 50% was below the min-width media query. I move the rule above the query and it seems to fix the issue.

I added some screenshots, so you can see what I was talking about.   

After the fix.
screen shot 2016-04-30 at 1 47 23 pm

screen shot 2016-04-30 at 2 02 00 pm

screen shot 2016-04-30 at 2 06 43 pm

Befor the fix.
screen shot 2016-04-30 at 1 49 17 pm

I could also change the css to have a max-width just for mobile sized screens if you want the inputs not to display 100% at 1020px and maybe have it change at a smaller screen size like 620px. I was not total sure if that is important or not. Now with the fix, I think it looks pretty good.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -465,6 +465,10 @@ label {
 	display: inherit;
 	margin-bottom: 20px;
 }
+textarea {
+	width: 50%;
+	height: 100px;
+}
 
 input,
 textarea,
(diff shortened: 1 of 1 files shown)


julianguyen:
@nonothetoad Thanks so much for the fix! 💃

nonothetoad:
You're very welcome!

PR #110, opened 2016-04-30
make comments appear in order of creation

lksevans12:
Hi! We thought that it made more sense to have the comments displayed in order of creation, so I just added an order query to the comments in the trigger controller.


julianguyen:
@lksevans12 Awesome work again Laura! :)

Issue #116, opened 2016-05-03
JS Error when user is not signed in


julianguyen:
![screen shot 2016-05-03 at 9 53 01 am](https://cloud.githubusercontent.com/assets/3010728/14991175/ed48e676-1114-11e6-9eef-8ecb65e0640b.png)

When the user is not signed in, the following error is thrown `VM9058:1 Uncaught SyntaxError: Unexpected token <`

Not that a Pusher (notification service) object is not created unless the user is signed in.

PR #118, opened 2016-05-03
Js error not signed in

siyegen:
For #116 

Changes the dataType of the ajax call to json, so that the controller returns the json ```format.json { head :no_content }``` instead of the html and confirms that we got something back.

Tried to make as few changes as possible, so let me know if you had a different way you'd want to approach this.

### DIFF
--- a/app/assets/javascripts/notifications.js
+++ b/app/assets/javascripts/notifications.js
@@ -3,26 +3,28 @@ $(document).on("page:load ready", function() {
   var pusher;
 
   $.ajax({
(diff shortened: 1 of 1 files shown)


julianguyen:
@siyegen LGTM! Do notifications still work too?

siyegen:
They should, I used the sample user and confirmed the pusher creation still
happened.

julianguyen:
Awesome! Thanks again 👍 👍 👍

PR #124, opened 2016-05-11
Remove uploads folder

ehashman:
I think these files were checked in by accident?


julianguyen:
@ehashman Looks like it! Good catch!

PR #133, opened 2016-05-17
Fix rspec deprecation for requiring rspec/autorun

calebthompson:
RSpec produced the following deprecation warning, which is resolved by not
requiring `rspec/autorun` in spec/spec_helper.rb:

    Requiring `rspec/autorun` when running RSpec via the `rspec` command is deprecated. Called from /Users/caleb/.rvm/gems/ruby-2.3.0@ifme/gems/activesupport-4.2.5/lib/active_support/dependencies.rb:274:in `require'.


julianguyen:
LGTM! Thanks! :)

PR #135, opened 2016-05-17
Remove protected_attributes gem

calebthompson:
[Protected attributes] were an extraction from rails 3.2 to allow continued
use of an old style of mass assignment protection. They were replaced by the
strong_parameters gem which is a part of Rails 4+. The protected_attributes
gem won't be supported past the forthcoming rails 5.0 release, and it's not
doing much for us so let's drop it.

[Protected attributes]: https://github.com/rails/protected_attributes

I also included some opportunistic refactorings as I came along them.

This PR is best reviewed with ?w=1 as there are many, many tabs being removed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,8 +50,6 @@ gem 'spring',        group: :development
 
 gem "ruby-graphviz"
 
-gem 'protected_attributes'
-
 gem 'rails_12factor', group: :production
 
 gem 'ckeditor'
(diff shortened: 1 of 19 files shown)


calebthompson:
https://github.com/julianguyen/ifme/pull/135/files?w=1

julianguyen:
💯  awesome!

PR #137, opened 2016-05-17
Fix bug where a user couldn't log in with oauth

stellacotton:
I realized when I was trying to set up my local environment that there was a regression in the oauth login.

User.find_for_google_oauth2 originally returned a user object, which oauth used
to authenticate when redirecting the user to their homepage. In commit
e68e37a, it started returning a true/false value from update!. This
commit switches it back to returning a user and adds specs.


stellacotton:
@julianguyen I wasn't sure if I should also open a bug issue on the issue tracker as well?

julianguyen:
@stellacotton Thanks for tracking this! Yeah open a bug issue too!

PR #141, opened 2016-05-20
Configure all application secrets from a single file and put them in ENV

stellacotton:
This uses the gem Figaro and allows for quicker development setup since
all api keys and secrets can be updated in a single place.
The config/application.yml file and the config/smtp.yml file are created
locally and in .gitignore by default to prevent accidentally checking in
sensitive information. This commit also passes the Pusher environment
variable through to JavaScript by putting it in a metatag so that it's
no longer hardcoded inside the JavaScript.

These files specifically had configuration data extracted:
app/assets/javascripts/notifications.js
app/mailers/application_mailer.rb
config/api.yml
config/environments/development.rb
config/initializers/pusher.rb
config/secrets.yml
config/cloudinary.yml
config/initializers/devise.rb


julianguyen:
💯 💯 💯 💯 💯 💯  Excellent work @stellacotton. Thanks so much <3

PR #148, opened 2016-05-24
Removing line that sets @group in show method

kkelleey:
@group is already set in the private set_group method

Also addings specs to confirm that this doesn't change functionality and making the group factory valid


julianguyen:
@kkelleey Great catch! Thanks Karen! :) 👍

PR #149, opened 2016-05-24
Cleaning up the set_group method

kkelleey:
This is how I've seen the not found handled in other places. It cleans
things up a little bit


julianguyen:
LGTM! Awesome work 💯

PR #150, opened 2016-05-24
Fixing indentations in groups_controller

kkelleey:
Also adding space below 'private' declaration and removing blank line near the end


julianguyen:
💯

PR #151, opened 2016-05-24
Refactoring groups#index

kkelleey:
* Adding HABTM association between groups and users so that you can call
current_user.groups
* Changing @groups to be an array of groups instead of group_members and
changing all references to group.groupid to group.id within the @groups loop 
* Making group_member factory valid


julianguyen:
Thanks for all this 👍 👍👍

PR #152, opened 2016-05-24
Stella/downgrade google api

### LINKED ISSUE #140: Unable to set refill date for medications
When I go to create a new medication with a refill date, it raises an error `NameError Exception: uninitialized constant MedicationsController::Google`

It looks like a version bump on the google-api-client from 0.8 to 0.9 causes breaking changes to the existing API interface: https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md  😿

If 0.8 was working ok and there weren't any features needed in 0.9, it might be worth rolling back to 0.8 and dealing with the upgrade at a later time. I'm happy to roll it back if you don't think it's worth trying to upgrade the gem @julianguyen?

stellacotton:
This downgrades the Google API to fix #140. It also extracts out the Google Calendar upload logic out of the controller into a service object and sets the name of the app in the config/application.yml instead of using the same set in the en.yml. Since this introduces the "service object" as an architectural pattern in the app, definitely open to feedback on whether that seems like the right move!

I also added a test for the Google calendar uploader, but its pretty bare bones and involves a lot of stubbing for the relatively complicated Google API interface.

I didn't add any of the functionality for asking the user if we should save the medication to their google calendar. I figured it might be better to just get this fix up separately.


stellacotton:
Oh and @julianguyen just a heads up, if you're fine with extracting out the name of the Google application into the config file, we'd need to add a new ENV variable in production too. Sorry I'm ENV-ing left and right in these PRs! 😀

julianguyen:
Great changes! 💯

PR #154, opened 2016-05-25
Refactoring groups#index

kkelleey:
* Adding available groups method to user model
* Changing available_groups instance variable to an array of groups
 instead of an array of group_members and updating references to group
 in avalable_groups loop accordingly in the view
* Changing HABTM relation between groups and members to has_many
 :through because it is recommended to use has_many through if the join
table has extra attributes (as we do here with status)
* Fixed some indentation on groups/index (probably should have done that as a separate PR, it looks like I made more changes than I did)


julianguyen:
LGTM @kkelleey. Excellent work as always :)

Also! If you want to add a little blurb about yourself on the Contributors page please do :)

PR #158, opened 2016-05-26
Refactoring joined groups part of groups/index

kkelleey:
* Eager loading group members so that they are easily accessible within
 the view
view
* Fixing indentation
* Adding a few helper methods for complicated links, etc


julianguyen:
Excellent! 💯  Thank you for this @kkelleey!

PR #160, opened 2016-05-26
Refactoring groups/index view, part 2

kkelleey:
* Eager loading group_members in available_groups
* Fixing indentation in available_groups loop
* Adding join_group_link view helper
* Adding render_group_member_partial helper
* Adding user_is_leader local variable in groups loop. Before this, the db was being hit with the same query three times per loop. Now it is only hit once


julianguyen:
💯  Seriously, your refactoring skills are amazing. Thank you <3

kkelleey:
No problem, I have some extra time and I enjoy refactoring 😄

PR #166, opened 2016-05-27
Refactoring groups#show

kkelleey:
* Adding has and belongs to association between group and meetings
* Adding led_by? method to group
* Adding specs for instance variable assignments


julianguyen:
Excellent as always! 💯

PR #168, opened 2016-05-27
Refactoring second pass

kkelleey:
* Changing render_group_member_partial to accept group_members instead
of group
* Cleaning up groups/show view some more and adjusting controller
instance variables accordingly

This is kind of an intermediate refactor, I'd like to add a few more
helper methods for links and cut down on some of the instance variables
in the controller, but I didin't want to do too much in one commit


julianguyen:
LGTM!

PR #169, opened 2016-05-27
Adding leaders relation to meeting model

kkelleey:
This will allow us to easily call meeting.leaders or eager load
group.meetings.includes(:leaders)


julianguyen:
👍

PR #170, opened 2016-05-27
Eager loading meeting leaders in controller

kkelleey:
* Updating view to use eager loaded leaders instead of querying
* Adding default order to group.meetings so that it doesn't have to be
 specified in the controller
* Setting user_is_meeting_leader as a variable at the top of the
meetings loop since it is needed more than once


julianguyen:
LGTM!

PR #171, opened 2016-05-27
Adding has_many leaders to group

kkelleey:
* Updating groups controller
* Updating view (@group_leaders are now instances of user instead of
 group_member)
* Adding spec for Group.leaders
* Removing redundant private method leaders
* Removing controller test for @group_leaders because it's covered by
 model spec


julianguyen:
LGTM!

PR #172, opened 2016-05-28
Cleaning up delete/leave/join group links

kkelleey:
* Tweaking link helpers to accept additional attributes so that helpers
 can be used for both index and show
* Adjusting locales so that .join .delete .leave are translated in the
 same way they were before the helpers were used
* Removing deletable instance variable since it's only used once and it
 doesn't hit the db


julianguyen:
LGTM!

PR #173, opened 2016-05-28
Removing unnecessary instance variables

kkelleey:
After some experimenting with rendering the page and watching the
console, I figured out (I think) that Rails doesn't evaluate instance
variables in the controller if it doesn't have to. This means that
when you set @group_members = @group.members, the query for group
members isn't made until @group_members is used in the view.
Furthermore, it looks like if @group.members is used more than once in
the view, the query to get the members is only made once. Therefore,
setting the @group_members variable doesn't have any advantage over
simply using @group.members in the view.


julianguyen:
Whoa really great catch! :O 💯

PR #179, opened 2016-05-28
Further refactoring groups/show by adding helpers

kkelleey:
* Cleaning up meetings loop. `<h1 class="meeting_name"</h1>` and the meeting link were being used in both if and else blocks, the only thing that was really changing conditionally was the edit_meeting link. I moved the logic for that to the edit_meeting_link helper. 
* Changing t('groups.show.delete') to t('groups.cancel_meeting) because
 the page was displaying 'Delete' because en.yml had two translations
for delete, so Delete was being displayed instead of Cancel Meeting


julianguyen:
LGTM

PR #181, opened 2016-05-29
More cleanup

kkelleey:
* Renaming group_is_deletable helper method
* Adding one liner ifs where possible
* Adding member_pluralized_for helper
* Fixing more indentation

### DIFF
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -7,54 +7,38 @@
 <% if @groups.any? %>
   <h1 class="group_type">Joined Groups</h1>
   <% @groups.each do |group| %>
-    <% user_is_leader = user_is_leader_of?(group) %>
+    <div class="group <%= 'no_margin_bottom' if group == @groups.last %>">
+      <h1 class="group_name">
+        <%= link_to group.name, group %>
+        (<%= t('.leader') if user_is_leader_of? group %>)
+      </h1>
 
-    <% if @groups.last == group %>
-      <div class="group no_margin_bottom">
-    <% else %>
-      <div class="group">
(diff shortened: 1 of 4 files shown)


app/views/groups/index.html.erb
```diff
@@ -7,54 +7,38 @@
...
+      </div>
 
-    <%= strip_tags(group.description[0..80]) %>
-    <% if group.description.length >= 80 %>
-      <%= " [...]" %>
-    <% end %>
+      <%= strip_tags(group.description[0..80]) %>
+      <%= " [...]" if group.description.length >= 80 %>
 
-    <% if user_is_leader || (!user_is_leader || group.group_members.count > 1) || (user_is_leader && group.group_members.count == 1) %>
```
kkelleey: I don't think we need this if. I can't think of a case where none of those would be true.

Gemfile.lock
```diff
@@ -414,8 +414,5 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   underscore-rails
 
-RUBY VERSION
```
julianguyen: Why remove this?

kkelleey: Oops, didn't mean to commit that. My Gemfile.lock keeps changing this for some reason, so I usually have to `git checkout Gemfile.lock` to avoid committing it. I should probably figure out why that's happening. 

Good catch 👍

julianguyen: Sound good >.< Other than that commit LGTM :) Please update!

kkelleey:
Updated

julianguyen:
Thank you ^_^

PR #185, opened 2016-05-29
Refactoring groups_controller_spec

kkelleey:
* Adding StubCurrentUserHelper to stub the current user with either a
stubbed user or a created user
* Adding specs for when user doesn't exist


julianguyen:
LGTM!

PR #188, opened 2016-05-30
Alphabetize members in member partial

### LINKED ISSUE #186: Alphabetical ordering of members list in Groups and Meetings
Hey @kkelleey! Since you've been working a lot with the Groups and Meetings code lately, I was wondering if you could do this. Thanks :)

screen shot 2016-05-29 at 6 52 57 pm

Test1 Lastname should appear before Test2 Lastname

kkelleey:
Resolves #186

To do this, I added the has_many :members association to both Group and
Meeting and changed the notifications/_members partial so that it
can receive an instance of either group or meeting as the local variable
called group. Using this instance, the partial can get anything it needs
since both group and meeting respond to #leaders and #members. Instead
of passing a second :is_meeting_member local, we can check whether group
is an instance of Group.

In order to implement the actual alphabetization part, I added the
default ordering on name to the members association to both group and
meeting models.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -19,7 +19,7 @@ class GroupsController < ApplicationController
   def show
     @page_title = @group.name
 
-    if @group.users.include? current_user
+    if @group.members.include? current_user
       @meetings = @group.meetings.includes(:leaders)
     end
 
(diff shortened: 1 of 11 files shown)


julianguyen:
LGTM! Thanks for working on this :)

PR #190, opened 2016-05-31
More cleanup of meeting_info partial

kkelleey:
* Pulling some translations up from groups.show (for time and location I
removed the colon from the translation and added it to the view)


julianguyen:
LGTM! :)

PR #192, opened 2016-06-01
Updated README: Link to a postgres tutorial for OSX

nagmak:
Added a link to the postgres tutorial I used. Let me know if I'm doing this right =D I've never created a pull request before

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ Install via [Homebrew](http://brew.sh/)
 
 `brew install postgresql`
 
+After that, follow [this guide](http://exponential.io/blog/2015/02/21/install-postgresql-on-mac-os-x-via-brew/) for a more detailed setup
+
 #####Linux
 
 Bulid from the source using yum, apt-get, etc.


starbelly:
@nagmak Definitely doing it right : ) Though, it looks like you are are a collaborator, so I don't think you needed to do the PR, unless you wanted to open a discussion on the link you're providing. I do think that this documentation is concise enough to where it makes more sense to add this to the README itself or perhaps in a separate INSTALL.md file.

nagmak:
Yay :) As a collaborator I can make commits directly to the repo? So PR would be for more major changes or bugs then. Thanks for the feedback on that!

starbelly:
@julianguyen Might be a good idea to establish guidelines for collaborators at this point. Do we do a PR for everything? Obviously this is somewhat of a negligible change, but maybe it shouldn't have just gone through. I assumed it would be fine since there were no rules or guidelines on the matter in place, in addition to the fact that @nagmak is a collaborator. Yet, still we need to be clear about all this, otherwise we could end up with a bit of a mess at some point.

maestromusica:
I think that pull requests should be done for every change in the code. Just to make sure that 'another pair of eyes' will go through the changes to make sure no unwanted/dangerous side effects will be introduced to the codebase.

julianguyen:
Agree with you two, please make PRs, you can push directly if you are undoing something. Sorry I did not make that clear. I will update the docs to clarify that!

PR #194, opened 2016-06-01
Add feature specs

kkelleey:
This PR is two commits

The first adds associations to factories instead of using hard coded ids and updates the model specs that failed because of the change

The second configures warden so that it can be used in feature specs and adds two simple feature specs for groups/index and groups/show, which are just simple smoke tests that should hopefully catch any future changes that might cause the page to error


julianguyen:
LGTM! Like the use of the warden!

PR #202, opened 2016-06-03
Adding meeting_memberships association to group_member

kkelleey:
* Adding a callback to destroy all associated meeting memberships when a
group_membership is destroyed

I usually hesitate to add callbacks because they can get out of control
quickly but in this case it is really just a dependent: :destroy. I
couldn't do a real dependent: :destroy because it is a nested
association


julianguyen:
LGTM! A callback here makes sense!

PR #203, opened 2016-06-04
Tidy up Gemfile

starbelly:
- Tidy up Gemfile and specify version for all gems
- Bump up ruby to 2.3.1
- Bump up rails to 4.2.6
- Bump bootstrap-datepicker to 1.6.1.1
- bump byebug up to 9.0.5
- bump guard up to 2.14.0
- Remove reference to rake, deps will handle
- Remove therubyracer (not being used in code)
- Remove reference to bundler, deps will handle


starbelly:
@julianguyen Before we merge on this. Are there any gems you can see that we can do without?

julianguyen:
LGTM, is there a specific reason why you bumped up the version for Ruby and Rails?

starbelly:
@julianguyen Just for bug fixes in rails, didn't see any good reason to stay on 4.2.5

julianguyen:
That's fine then. I'm just wondering how it will affect existing contributors, especially those running Ruby on Rails in Windows.

julianguyen:
Also random: do you recommend updating to Rails 5 before the official release?

starbelly:
@julianguyen Yes I do. I upgraded a friend from 4 to 5 and it was painless.

PR #206, opened 2016-06-05
More refactoring of groups#leave

kkelleey:


app/controllers/groups_controller.rb
```diff
@@ -164,35 +164,34 @@ def join
...
-
-    group_name = Group.where(id: params[:groupid]).first.name
+    member_id = params[:memberid] || current_user.id
+    group_member = GroupMember.find_by(userid: member_id, groupid: params[:groupid])
 
     # Cannot leave When you are the only leader
-    is_leader = GroupMember.where(userid: memberid, groupid: params[:groupid], leader: true).count
-    are_leaders = GroupMember.where(groupid: params[:groupid], leader: true).count
-    if is_leader == 1 && are_leaders == is_leader
+    if group_member.group.leader_ids == [member_id]
```
stellacotton: It's probably worth adding a feature spec to `user_leaves_groups_spec` to cover this case since I don't think it's covered elsewhere?

stellacotton: Also, I know we only use `group_member.group` twice in this method, but it might make the flow a little clearer if we assign `group` to a variable and operate on it directly instead of going through the `group_member.group`.

stellacotton:
Lovely refactoring! The controller is much clearer now. I added one tiny comment and a test suggestion, but overall looks good to me!

kkelleey:
Added the group variable :)

I'm going to add a controller spec; I can't add a feature spec because the 'Leave' link isn't rendered if the user can't leave the group.

stellacotton:
Ah, so this is just in case they do somehow get to this page (or the other leader leaves out from under them)?

kkelleey:
Yeah exactly

PR #207, opened 2016-06-06
WIP: Initial refactor of schema and migrations

starbelly:
- Use t.references to create indexes and columns for associations
  -  Added foreign keys for referential integrity
  -  Use symbols instead of strings in migrations
  - Remove add index (references takes care of that, defaults to btree)
  - Rename userid columns to user_id
  - Rename groupid columns to group_id
  - Rename meetingid columns to meeting_id
  - Rename user2 in allyships to ally (ally_id)
  - Rename comment_by columns to user_id
  - Update test units to work with new column names and foreign key
constraints
  - Re-factor migrations back out into seperate files


starbelly:
@julianguyen This is **not** ready to me merged, I _think_ (keep reading). 

  This is a breaking change, meaning when this is pushed out to production, the existing database will have to be backed up,  blown away, and re-migrated, and then re-imported. In fact, a custom script might be in order to pull data back in. 

 This is only the beginning though : ) I've seen lots of areas in the schema that need to be worked on (next on my list is multi-value-attribute columns that I've seen). Anywho, we can merge this now, and then introduce yet another breaking change in the next few days or so. I figured, probably best to wait until I'm all done in this area. 

 None the less, we need to discuss the changes before merging anything.

starbelly:
@dleve123 What are your thoughts Mr. man?

starbelly:
@julianguyen Good luck.

PR #208, opened 2016-06-06
Refactoring groups#destroy

kkelleey:
Adding dependent: :destroy associations so @group.destroy will destroy
associated meeting and meeting_members

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -202,14 +202,7 @@ class GroupsController < ApplicationController
   # DELETE /groups/1
   # DELETE /groups/1.json
   def destroy
-    # Destroy group members
-    GroupMember.where(groupid: @group.id).destroy_all
-
-    # Destroy meetings and its members
-    Meeting.where(groupid: @group.id).all.each do |meeting|
-      MeetingMember.where(meetingid: meeting.id).destroy_all
-      meeting.destroy
-    end
+    @group.destroy
 
     # Delete notifications for this group
     Notification.where("uniqueid ilike ?", "%new_group%").all.each do |notification|
@@ -218,7 +211,6 @@ class GroupsController < ApplicationController
       end
     end
 
-    @group.destroy
     respond_to do |format|
       format.html { redirect_to groups_path }
       format.json { head :no_content }
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -12,10 +12,10 @@
 class Group < ActiveRecord::Base
   validates_presence_of :name, :description
 
-  has_many :group_members, foreign_key: :groupid
+  has_many :group_members, foreign_key: :groupid, dependent: :destroy
   has_many :members, -> { order 'name' }, through: :group_members, source: :user
-  has_many :meetings, -> { order 'created_at DESC' }, foreign_key: :groupid
-  has_many :meetings, foreign_key: :groupid
+  has_many :meetings, -> { order 'meetings.created_at DESC' },
+           foreign_key: :groupid, dependent: :destroy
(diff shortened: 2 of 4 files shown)


app/controllers/groups_controller.rb
```diff
@@ -202,14 +202,7 @@ def leave
...
   def destroy
-    # Destroy group members
-    GroupMember.where(groupid: @group.id).destroy_all
-
-    # Destroy meetings and its members
-    Meeting.where(groupid: @group.id).all.each do |meeting|
-      MeetingMember.where(meetingid: meeting.id).destroy_all
-      meeting.destroy
-    end
+    @group.destroy
```
kkelleey: Who should be allowed to delete a group? Should we make sure that current_user is a leader of the group?

app/models/group.rb
```diff
@@ -12,10 +12,10 @@
 class Group < ActiveRecord::Base
   validates_presence_of :name, :description
 
-  has_many :group_members, foreign_key: :groupid
+  has_many :group_members, foreign_key: :groupid, dependent: :destroy
   has_many :members, -> { order 'name' }, through: :group_members, source: :user
-  has_many :meetings, -> { order 'created_at DESC' }, foreign_key: :groupid
-  has_many :meetings, foreign_key: :groupid
+  has_many :meetings, -> { order 'meetings.created_at DESC' },
```
dleve123: Was this previously busted and you just noticed it? Regardless, I would think it would be valuable to capture this potential regression with some unit tests to characterize this behavior.

dleve123:
This generally LGTM from a code perspective! Definitely a solid refactoring to the model-layer!

app/models/group.rb
kkelleey: Are you talking about the `meetings.created_at` change or the removal of the duplicate has_many :meetings association?
For the first, I'm pretty sure it was working find before this, but when I added both dependent destroys to both Group has_many :meeetings and Meeting has_many :meeting_members  and tried calling group.destroy, I got ambiguous column errors because it was trying to also join meeting_members. So technically group.members was working as expected but it wouldn't have worked if you tried additional joins. The Group#destroy unit test in this PR should catch that regression.

dleve123: Yeah, I am talking about the former. Because meeting_members is dependent on meetings, and hence is included in the deletion cascade? If so, makes sense, but weird nonetheless.

kkelleey: Yeah exactly.

app/models/meeting.rb
```diff
@@ -21,7 +21,7 @@ class Meeting < ActiveRecord::Base
 
   has_many :members, -> { order 'name' }, through: :meeting_members,
                                           source: :user
-  has_many :meeting_members, foreign_key: :meetingid
+  has_many :meeting_members, foreign_key: :meetingid, dependent: :destroy
```
kkelleey: I wonder if we could make this `dependent: :delete_all` ? We definitely want `dependent: :destroy` for the `Group has_many :meetings` association above so that meetings are instantiated and associated meeting_members get destroyed, but I can't think of a reason we would want to instantiate meeting_members before destroying them.

dleve123: The [relevant docs for `dependent: delete_all`](http://api.rubyonrails.org/v4.1.8/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many) state that:

> :delete_all causes all the associated objects to be deleted directly from the database (so callbacks will not be executed).

So in the event that callbacks are added to the `MeetingMember` model, changing this to `dependent: delete_all` will silently ignore them, which I imagine will be problematic. So unless `GroupsController#destroy` is shown to be a performance hotspot, I would suggest leaving as is.

PR #209, opened 2016-06-07
Adding validation messages in groups/form

kkelleey:
screenshot 2016-06-06 19 37 18
Before this commit, if you tried to submit new group without a
description, there was a JavaScript validation that prevented the form
from being submitted, but there was no error message, so it was
confusing why nothing was happening when 'Create Group' was clicked. By
removing this piece of JS and displaying the rails validation errors,
the user will be able to see why the group isn't sucessfully being
created.

### DIFF
--- a/app/views/groups/_form.html.erb
+++ b/app/views/groups/_form.html.erb
@@ -1,8 +1,7 @@
 <%= form_for(@group) do |f| %>
   <% if @group.errors.any? %>
     <div class="error_explanation">
-      <h2><%= pluralize(@group.errors.count, 'error') %>
-        prohibited this group from being saved:</h2>
+      <h2>We can't save this group because: </h2>
       <ul>
         <% @group.errors.full_messages.each do |msg| %>
           <li><%= msg %></li>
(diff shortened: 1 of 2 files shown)


app/views/groups/_form.html.erb
```diff
@@ -1,9 +1,15 @@
 <%= form_for(@group) do |f| %>
-<% if @group.errors.any? %>
-  <div class="error_explanation">
-    <%= t('.error_explanation') %>
-  </div>
-<% end %>
+  <% if @group.errors.any? %>
+    <div class="error_explanation">
+      <h2><%= pluralize(@group.errors.count, 'error') %>
+        prohibited this group from being saved:</h2>
```
kkelleey: I copied this from RailsGuids, so if we want different phrasing, I can update it

http://guides.rubyonrails.org/active_record_validations.html#displaying-validation-errors-in-views

aayushpi: IMO, I’d probably humanize this by saying something in 'we' language: We can’t save this group because:

*
*

We could also just go through all the errors later and revise the language.

kkelleey: Good idea. Updated.

aayushpi:
LGTM 👍 Thanks ~

Issue #205, opened 2016-06-04
Getting tooltips to show on mobile without clicking link


julianguyen:
I recently added stats on Categories, Moods, and Strategies on user profiles. In your own profile, you can hover over each stat and a tooltip with a list of viewers will show up. On mobile web, you can't hover... so the only way you can see it is if you tap the text... but tapping the text will click the link. 

![screen shot 2016-06-04 at 11 31 32 am](https://cloud.githubusercontent.com/assets/3010728/15801411/7e979ae0-2a48-11e6-8cbe-e8510c6e7a17.png)

Instead, we want to be able to tap over anywhere in the background div and have the viewers tooltip show up.

julianguyen:
Please look at viewers_hover in application_controller.rb :)

andela-cdaniel:
Hi, good day. 

I got to know of this project from `yourfirstpr` on `twitter` and I'd like to have a go at this issue. 

I've already hacked at it from the browser and was able to come up with this:

<img width="1280" alt="screen shot 2016-06-05 at 10 30 44 am" src="https://cloud.githubusercontent.com/assets/7131935/15804678/3279547a-2b09-11e6-86ed-9708411dd7c3.png">
<img width="1280" alt="screen shot 2016-06-05 at 10 31 05 am" src="https://cloud.githubusercontent.com/assets/7131935/15804679/3284114e-2b09-11e6-9430-162cde146ffb.png">
<img width="1280" alt="screen shot 2016-06-05 at 10 31 24 am" src="https://cloud.githubusercontent.com/assets/7131935/15804680/329ed718-2b09-11e6-908b-c5fe4afb2ac7.png">

The idea behind it is to only show the modified `tooltip` when on mobile and make it dismissible. Doing it this way ensures that the default behaviour of clicking a link and having it redirect to a different page is preserved, furthermore, the user will not have to bother about clicking the text or the background encompassing the text.

PR #211, opened 2016-06-10
Feat (user interface): Show viewer tool-tip on mobile

andela-cdaniel:
Show viewer tool-tip on mobile when background element encompassing the text is clicked
- Create javascript file to handle the viewer tool-tip on mobile

Implements #205

### DIFF
--- a/app/assets/javascripts/mobile_tooltip.js
+++ b/app/assets/javascripts/mobile_tooltip.js
@@ -1,0 +1,41 @@
+(function () {
+  /*
+  **  Tooltip for the profile page on mobile devices
+  **  This script should ideally run on the profile page
+
+  **  Behaviour
+      ========================
+      - Enables the viewers tooltip to be shown on mobile devices when the
+        parent element is clicked.
+      - Toggle behaviour is implemented so that the tooltip can be both shown
+        and dismissed
+
+  */
+
+  $(document).on("page:load ready", function () {
+    // Allow this script to only be run on the profile page
+
+    if (window.location.pathname !== "/profile") return false;
+
+    var toggleCount = 0;
+
+    // Ensure the script is activated on mobile devices only
+    // Max-width: 768px
+
+    if ($(window).width() <= 768) {
+      $(".some_focus_text").on("click", function () {
+        var child = $(this).children(".yes_title").eq(0),
+            title = child.prop("title");
+
+        child.tooltip({
+          content: title,
+          position: { my: "top-10 left+25", at: "top left", of: $(this) }
+        });
+
+        toggleCount++;
+
+        toggleCount % 2 === 1 ? child.tooltip("open") : child.tooltip("close");
+      });
+    }
+  });
+}());


stellacotton:
@dleve123 Would you be able to review this PR? Otherwise I can send out the bat signal in Slack for someone with JS experience to take a look. 😄

dleve123:
I'm not really going to be able to review until later this week at the
earliest, so probably makes sense to tag someone else in for this one!

app/assets/javascripts/mobile_tooltip.js
```diff
@@ -0,0 +1,41 @@
...
+        parent element is clicked.
+      - Toggle behaviour is implemented so that the tooltip can be both shown
+        and dismissed
+
+  */
+
+  $(document).on("page:load ready", function () {
+    // Allow this script to only be run on the profile page
+
+    if (window.location.pathname !== "/profile") return false;
```
sienatime: I don't think this is strictly necessary... `some_focus_text` is only used in the profile page, and if we ever have another use case for mobile tooltips on a different page, we should just keep using that class (or add a new one). I think this falls enough into a "general utility" category that we don't have to restrict its usage with an early return like this.

app/assets/javascripts/mobile_tooltip.js
```diff
@@ -0,0 +1,41 @@
...
+            title = child.prop("title");
+
+        child.tooltip({
+          content: title,
+          position: { my: "top-10 left+25", at: "top left", of: $(this) }
+        });
+
+        toggleCount++;
+
+        toggleCount % 2 === 1 ? child.tooltip("open") : child.tooltip("close");
```
sienatime: Okay, I experimented with this, and this works most of the time, but consider this situation:

- There are 3 elements with the class `some_focus_text`
- I click on the first element -> toggleCount increases to 1 -> first tooltip shows
- I click on the second element -> toggleCount increases to 2 -> nothing happens, because toggleCount % 2 is 0, and it tries to close something that isn't there
- I click on the first element again -> toggleCount increases to 3 -> nothing happens, because toggleCount % 2 is 1, but the tooltip is already open so opening it again does nothing
- I click on the first element a third time -> toggle Count increases to 4 -> the first tooltip finally closes

What would be better is not tracking how many times we click any element with the class `some_focus_text` but rather ask the element being clicked on if it has a tooltip open or not already, and do the opposite action of that. After some poking around, I found that the `yes_title` span gets an `aria-describedby` attribute with the ID of a tooltip if that tooltip is already open. So you could do something like:

```
var tooltipId = $(child).attr('aria-describedby');

if (tooltipId) {
  child.tooltip("close");
} else {
  child.tooltip("open");
}
```

and then things open/close as expected even given the above scenario.

sienatime:
Hi! I took a look. I had some suggestions after spending some time with the jQuery UI position() API http://api.jqueryui.com/position/ Let me know if you have any concerns.

app/assets/javascripts/mobile_tooltip.js
```diff
@@ -0,0 +1,41 @@
...
+    // Max-width: 768px
+
+    if ($(window).width() <= 768) {
+      $(".some_focus_text").on("click", function () {
+        var child = $(this).children(".yes_title").eq(0),
+            title = child.prop("title");
+
+        child.tooltip({
+          content: title,
+          position: { my: "top-10 left+25", at: "top left", of: $(this) }
```
sienatime: Okay I had a tough time trying to figure out why the position of the tooltip was dead center instead of any sort of top left alignment.

screen shot 2016-06-17 at 2 37 57 pm

It turns out that these arguments need to be flipped (it's horizontal, vertical and not vertical, horizontal) so this would become 
```
position: { my: "left+25 top-10", at: "left top", of: $(this) }
```
which yields this position, which makes a lot more sense but IMO doesn't look great (kind of confusing as to which item we're talking about, I clicked on the 2nd one)

screen shot 2016-06-17 at 2 51 29 pm

I think based on how this looks on desktop, I'd like to propose

```
position: { my: "left+40 top+25", at: "left top" }
```
(note no more `of: $(this)`) which yields this

screen shot 2016-06-17 at 2 53 56 pm

Compare to desktop:

screen shot 2016-06-17 at 2 54 50 pm

But of course, open to other suggestions as well.

julianguyen: LGTM

stellacotton:
Hey @andela-cdaniel, do you think you'll be able to tackle some of the changes that @sienatime suggested? We can also pass it on to someone else if you aren't able to work on it. 😊

julianguyen:
What's the status on this?

andela-cdaniel:
@everyone, sorry for my prolonged absence. I'll work on integrating the changes suggested by @sienatime and update the pr.

julianguyen:
@andela-cdaniel Thanks so much :) Also please email join.ifme@gmail.com your email so we can add you to our Slack :P

julianguyen:
Any updates on this @andela-cdaniel? (:

julianguyen:
Going to merge this and do some refactoring myself

PR #214, opened 2016-06-15
Meeting info view was expecting number, but was getting an association

stellacotton:
This was intermittently failing on CI, but represents a real issue


kkelleey:
Oops, I was the one who wrote that, good catch!

Issue #129, opened 2016-05-16
Add email reminders for medications


julianguyen:
Add a field in Medications for users to send email reminders
Also send email reminders for refill dates

Issue #112, opened 2016-05-02
Change form layout in Medications


julianguyen:
Currently it looks crowded. Any layout change that looks consistent with the other forms in the app is fine :)

![screen shot 2016-05-02 at 4 41 35 pm](https://cloud.githubusercontent.com/assets/3010728/14971209/069dd446-1085-11e6-9ab1-de1a77346b36.png)

julianguyen:
Wanna start working on @aayushpi? :)

aayushpi:
Is there a preferred order that you’ve found helpful, or users have suggested?

julianguyen:
Nope! But I think an order that makes sense when you're reading off your prescription bottle makes most sense.

aayushpi:
![image](https://cloud.githubusercontent.com/assets/2206776/15720879/3976c1ac-27eb-11e6-90ba-97af5c3eacf3.png)
Pass 1.

Still working on this

julianguyen:
Looks great @aayushpi! Hmm but should be using more of the screen width? It looks a little isolated floating in the center? Hmm would it be weird if were left aligned instead?

aayushpi:
It’s not done yet.

1. Screen width: IMO no, there’s a lot of research about fields that are longer than they typically need to be being left incomplete. Our data set is small, so the fields should be as such.

2. Floating in the center looks odd now, but the style isn’t done. Eventually, I still want this centralized but should solve the isolation that comes off.

Also, this may be mobile optimized at this point: needs testing.

aayushpi:
LOTS of styling changes incoming, this page is going to be pretty... ish

julianguyen:
Ah thanks for clarifying! Very interesting that longer fields are often left incomplete >.<

Looking forward to seeing them! Thanks Aayush :)

julianguyen:
@tiffanysun This was something Aayush was working on but didn't get a chance to finish. Would you want to pick up this first? :)

tiffanysun:
@julianguyen sounds good! i'll start here

kindlingscript:
Here's what I've done in Sketch so far, followed with how it looks in my local environment:

![if_me](https://cloud.githubusercontent.com/assets/13127241/19299459/0b6b07d6-901a-11e6-9f8e-a2eefc148aab.png)
[...]

PR #215, opened 2016-06-16
User can activate email reminders for medications

stellacotton:
This PR addresses #129.

A few things to consider:
- It doesn't look super beautiful, but with #112 coming down the pipeline, I figured there was more to be done when the medications tab is fancied up a bit.
- The emails will start to go out once we set up a scheduler on heroku for the two rake tasks `bundle exec rake scheduler:send_take_medication_reminders` & `bundle exec rake scheduler:send_refill_reminders`. I smoke tested these manually in my local environment (which is configured to send email out via gmail) but we can also test them on staging running those same commands from the command line.
- Refill reminders will go out a week before the refill date. If we want to make this functionality more configurable, we'll probably want to migrate away from storing these dates as strings and move them to timestamps.
- I'm open to feedback on the actual text of the emails themselves! They're pretty bare bones right now. 😄 

screenshot 2016-06-15 18 18 25

screenshot 2016-06-15 18 18 43

### DIFF
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -1,5 +1,24 @@
 class NotificationMailer < ApplicationMailer
-	def notification_email(recipientid, data)
+  def take_medication(reminder)
+    @medication = reminder.medication
+    @user = @medication.user
+    mail(
+      to: @user.email,
+      subject: "Don't forget to take #{@medication.name}!"
+    )
+  end
+
+  def refill_medication(reminder)
+    @medication = reminder.medication
+    @user = @medication.user
+    mail(
+      to: @user.email,
+      subject: "Your refill for #{@medication.name} is coming up soon!"
+    )
+  end
+
+
+  def notification_email(recipientid, data)
 		@data = JSON.parse(data)
 		@recipient = User.where(id: recipientid).first
 		subject = 'if me | '
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -1,4 +1,6 @@
 class NotificationMailer < ApplicationMailer
+   default from: ENV["DEFAULT_FROM_ADDRESS"]
+
   def take_medication(reminder)
     @medication = reminder.medication
     @user = @medication.user
--- a/app/models/take_medication_reminder.rb
+++ b/app/models/take_medication_reminder.rb
@@ -1,0 +1,5 @@
+class TakeMedicationReminder < ActiveRecord::Base
+  belongs_to :medication
+  validates_inclusion_of :active, in: [true, false]
+  scope :active, -> { where(active: true) }
+end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -31,7 +31,8 @@ Rails.application.configure do
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
   config.action_mailer.default_url_options = { host: 'localhost:3000' }
-  
+  config.action_mailer.default_options = { sender: 'test@test.com' }
+
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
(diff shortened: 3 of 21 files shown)


app/models/take_medication_reminder.rb
```diff
@@ -0,0 +1,5 @@
+class TakeMedicationReminder < ActiveRecord::Base
+  belongs_to :medication
+  validates_inclusion_of :active, in: [true, false]
+  scope :active, -> { where(active: true) }
+end
```
stellacotton: The reminder to take a medication and to refill a medication look basically identical right now and I did think about having them be one reminder model. But I think that once we start enhancing functionality, their scheduling patterns will inherently be pretty different and it felt like we'd almost immediately have to split them apart.

config/environments/test.rb
```diff
@@ -30,10 +30,13 @@
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
+  config.action_mailer.default_url_options = { host: 'localhost:3000' }
+  config.action_mailer.default_options = { sender: 'test@test.com' }
```
stellacotton: These enable the specs to run without relying on the real-life mail data that is configured separately (and not checked into the repo) in `application.yml`.

app/mailers/notification_mailer.rb
```diff
@@ -1,5 +1,26 @@
 class NotificationMailer < ApplicationMailer
-	def notification_email(recipientid, data)
+   default from: ENV["DEFAULT_FROM_ADDRESS"]
```
stellacotton: Added the default here to be consistent with the other mailer.

kkelleey:
> I'm open to feedback on the actual text of the emails themselves! They're pretty bare bones right now. 😄

@aayushpi could you maybe review the email text? (Since you provided helpful wording feedback on one of my other PRs)

I can try to review some of the code, although I'm not too familiar with mailers

spec/services/medication_reminders_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  describe "#send_refill_reminder_emails" do
+    context "no reminders" do
+      it "sends no emails" do
+        MedicationReminders.new.send_refill_reminder_emails
+        expect(deliveries.count).to eq(0)
+      end
+    end
+
+    context "reminders exist and it is one week before the refill" do
+      let!(:user) { FactoryGirl.create(:user1) }
```
kkelleey: Could this be a build_stubbed instead of a create? It looks like it's only used for its id, so it doesn't need to be persisted (I think).

Another option would be to add
`association :user, factory: :user1` to the medication factory, but we might not always need the associated user.

app/models/take_medication_reminder.rb
kkelleey: Could we make them both inherit from a Reminder model?

kkelleey:
Looks good! I added a few small suggestions/comments. Maybe someone else could take a look at it too since it's a pretty big PR

spec/services/medication_reminders_spec.rb
stellacotton: Good idea, I'll switch it over to build_stubbed.

app/models/take_medication_reminder.rb
stellacotton: I thought about it when I originally was going to use Single Table Inheritance for the two reminders. But I think with them being separate tables, it ends up being a Reminder concern with just 
`validates_inclusion_of :active, in: [true, false] ` and `scope :active, -> { where(active: true) }` which felt like overkill?

spec/services/medication_reminders_spec.rb
stellacotton: Ah, except I forgot that the emails themselves need the user to exist because we use information on the user and medication models in the email text.

app/models/take_medication_reminder.rb
kkelleey: That makes sense. Then I agree with keeping it as is.

sienatime:
I took a look, I think it looks great. I don't know things about schedulers but assuming the setup of that is coordinated along with this getting deployed, should be cool.

kkelleey:
thanks @sienatime!

kkelleey:
Merged #215

PR #216, opened 2016-06-17
More refactoring of groups#leave

kkelleey:
The ABC of the #leave method is still too high -- 29.5, but this is an
improvement

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -170,32 +170,21 @@ class GroupsController < ApplicationController
 
     # Cannot leave When you are the only leader
     if group.leader_ids == [member_id]
-      respond_to do |format|
-        format.html do
-          redirect_to groups_path,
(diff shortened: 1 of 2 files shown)


app/controllers/groups_controller.rb
```diff
@@ -170,33 +170,22 @@ def leave
...
+        flash[:notice] = "You have left #{group.name}"
       else
-        respond_to do |format|
-          format.html { redirect_to groups_path,
-                        notice: 'You have removed ' + group_member.user.name +
-                        ' from ' + group.name }
-          format.json { head :no_content }
-        end
+        flash[:notice] = "You have removed #{group_member.user.name} from \
+                         #{group.name}"
```
stellacotton: Yay! So happy to see the switch to string interpolation. It's a lot more readable.

stellacotton:
lgtm!

PR #217, opened 2016-06-17
added instructions for rbenv and added a line about copying database.…

sienatime:
…example.yml

Hi! I use rbenv instead of RVM for managing my Ruby versions so I added instructions for updating your Ruby version if you use rbenv. I also needed to copy `database.example.yml` to get up and running so I added that as well.


stellacotton:
lgtm, thanks @sienatime!

PR #218, opened 2016-06-17
adding ruby version

KaraAJC:
bitmoji
Upon updating local Ruby version via RBNV, a new .ruby-version file was created. Adding this to the repository for posterity.

Research on this preference lends itself to be safe to add this file to the repo: http://stackoverflow.com/questions/16087120/should-i-check-in-ruby-gemset-and-or-ruby-version


stellacotton:
Thanks @KaraAJC! This is compatible with rvm and rbenv and it would be good to lock the Ruby version explicitly, so lgtm!

Issue #221, opened 2016-06-17
Add cursor: pointer to elements that are clickable


sienatime:
I'm a big fan of adding the CSS `cursor: pointer` to anything that is clickable in order to indicate that you can interact with that element. If you add a function via Javascript to an element that isn't an anchor tag, you don't get this for free, so I propose an app-wide audit of places that could use a `cursor: pointer` rule. I found a few:

- Account/profile (logo) dropdown menu open/close
- in Groups, list of members link
- Notification "Clear"
- Notification dismiss (X) (probably could be added to `.fa-times` generally)
- link to Categories/Moods from create a story (home)
- Categories/Moods modal links to category or mood
- links to tagged categories/moods/strategies in story feed (under class `story`)

There could be others.

sienatime:
Opened PR https://github.com/julianguyen/ifme/pull/222

PR #222, opened 2016-06-17
[Issue #221] add cursor: pointer on elements that are clickable

sienatime:
Things that are clickable should look clickable :D


kkelleey:
This looks good to me

PR #223, opened 2016-06-18
Clean up groups#join, part 1

kkelleey:


app/controllers/groups_controller.rb
```diff
@@ -134,32 +134,32 @@ def update
...
-    group = Group.where(id: params[:groupid]).first.name
-
-    group_leaders.each do |leader|
-      data = JSON.generate({
+    group.leaders.each do |leader|
+      data = JSON.generate(
         user: current_user.name,
         groupid: params[:groupid],
-        group: group,
+        group: group.name,
```
stellacotton: This bug must have made for a pretty amusing group subject line. 😀

stellacotton: Ah, never mind I totally missed that originally "group" was the name. Whoops!

stellacotton:
lgtm!

Issue #224, opened 2016-06-24
Accessibility Issue - Multiple h1 tags in footer


johnfrenchxyz:
I've noticed that, in the footer, the "if me" and "connect" sections are both `<h1>` tags with additional styles applied to them. Having more than one `<h1>` tag on a page can make a non-css view difficult to understand, or cause a screen reader encounter unpredictable behavior.
 [
You can read more about best-practices here.](http://webaim.org/techniques/semanticstructure/#maincontent)

Locally, I have converted the `<h1>` tags to `<span>` tags and updated the scss (but have not yet compiled it to CSS, as I am slightly new to the rails file structure - is this done automatically with the asset pipeline?).  

When testing this fix in Chrome's developer tools - fixes the problem without altering the style of the footer. I have cloned the repo locally and created a new branch for these changes, but it appears as though I need to be added as a contributor before I can push the changes to my branch and submit a pull request.

stellacotton:
Hey John, an easy way to make a PR if you aren't a contributor is to fork the app to your own repo. Github has some instructions here: https://help.github.com/articles/fork-a-repo/

Since you've already pulled the repo down, after you fork the repo you can skip the steps in the instructions in Step 3.1 - 3.4 about pulling down the clone and go straight to Step 3.5 where it talks about setting up the remote.

Hope this is helpful!

johnfrenchxyz:
@stellacotton thanks! I'll look into that now.

johnfrenchxyz:
Just submitted a PR #225

PR #225, opened 2016-06-24
Fixes accessibility issue of having multiple h1 tags in the footer

johnfrenchxyz:
#224 

Hopefully fixes the above issue.  I have changed the `h1` tags in the footer to `span` tags and updated the footer.scss and _footer.html.erb respectively, but I have not compiled the scss, as I was not sure if this was done automatically in the asset pipeline, but the scss will need to be compiled.  Otherwise, I hope this fixes the issue - but eventually the "if me" branding at the top should probably be converted to an `h1` instead of a `span`.

I hope this helps!

### DIFF
--- a/app/assets/stylesheets/application/footer.scss
+++ b/app/assets/stylesheets/application/footer.scss
@@ -126,12 +126,13 @@
 	}
 }
 
-h1.footer_heading {
+.footer_heading {
 	font-weight: 300;
 	font-size: 30px;
 	padding: 0;
 	margin-top: 0;
 	margin-bottom: 20px;
+	display: block;
 	color: $header-text-color;
 	text-shadow: 0px 0px 20px $secondary-text-color;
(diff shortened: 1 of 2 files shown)


sienaigg:
Hello!! Thank you for the PR. If you make these `div`s instead of `span`s, you won't need the `display: block` as divs are block-level elements by default, and then that is one less CSS rule to have :D But yes I agree, I think the whole site needs an audit of header tags, there are quite a few `h1` tags that should probably be demoted...

julianguyen:
LGTM. Thanks @johnfrenchxyz and @sienaigg ^_^ I'll merge this and we can open up a new issue to do an audit of header tags for accessibility

johnfrenchxyz:
Awesome! I'll keep helping with the accessibility issues as I have time. Thanks, everybody.

PR #229, opened 2016-06-30
Refactoring groups#update

kkelleey:
The update method has a lot of responsibilities. I would like to split it into separate methods for each responsibility eventually, but this is a first step so that it is more clear what all is going on.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -87,40 +87,40 @@ class GroupsController < ApplicationController
   def update
     @page_title = "Edit " + @group.name
     respond_to do |format|
-      if @group.update(group_params) && !params[:group][:leader].nil?
-        group_members = GroupMember.where(groupid: @group.id).all
-        group_members.each do |member|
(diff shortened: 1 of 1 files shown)


app/controllers/groups_controller.rb
```diff
@@ -87,40 +87,40 @@ def create
...
     @page_title = "Edit " + @group.name
     respond_to do |format|
-      if @group.update(group_params) && !params[:group][:leader].nil?
-        group_members = GroupMember.where(groupid: @group.id).all
-        group_members.each do |member|
-          group_member_id = GroupMember.where(groupid: @group.id, userid: member.userid).first.id
-          if params[:group][:leader].include? member.userid.to_s
-            GroupMember.update(group_member_id, groupid: @group.id, userid: member.userid, leader: true)
+      if @group.update(group_params) && params[:group][:leader].present?
+        @group.group_members.each do |group_member|
```
kkelleey: I renamed the variable 'group_member' so that it is clear that it's an instance of GroupMember and not User

julianguyen:
LGTM! Thanks @kkelleey! 👍

Issue #180, opened 2016-05-28
Unit tests for JS


julianguyen:
There are currently no JS unit tests, which is problematic. I've set up Jasmine (https://github.com/julianguyen/ifme#unit-tests), so anyone is welcome to help write tests! :)

A sample test can be found here: https://github.com/julianguyen/ifme/commit/b6f4fb081cc2365142f0150564964698977b17be

julianguyen:
We would like to have individual test specs for
* add_comment.js
* application.js
* charactercount.js
* delete_comment.js
* form_processing.js
* header.js
* medications.js
* meetings.js
* notifications.js
* search.js
* strategies.js

If you need to refactor any of this code, go right ahead :)

julianguyen:
Hey @piecedigital, here's the issue for JS unit tests. Please sync up with @maestromusica to figure out which specs you want to work on :)

stellacotton:
For any new WriteSpeakCode folks, looks like @piecedigital and @maestromusica have done some awesome work on so far:
- header.js
- moments.js
- strategies.js

and the rest could use some love.

nchambe2:
Hey @julianguyen I would be interested in continuing work on this issue.

piecedigital:
@nchambe2 Feel free to work on this issue. As @stellacotton noted the new list should look like this:

* add_comment.js
* application.js
* charactercount.js
* delete_comment.js
* form_processing.js
* medications.js
* notifications.js
* search.js

nchambe2:
Hey @piecedigital, 
I plan to work on this issue today, along with another user @revolaution.

julianguyen:
Hey @nchambe2 @revolaution, could you send join.ifme@gmail.com your emails so I can add you to our Slack channel? Thanks :)

piecedigital:
@nchambe2 Alright. If you have any questions just ask me on Slack. Same name :smiley:

PR #230, opened 2016-07-01
Refactoring HeaderSpec tests - issue #180 

nchambe2:
This commit contains the following changes: 

- Change .toBeTruthy / .toBeFalsy assertions to .toBe(false) && .toBe(true)

- Refactor test names to use the following pattern : has  expected behavior When situation that behavior depends on

- Replace use of == with ===

- Refactor tests that have multiple assertions into multiple tests with singular assertions 

- Remove comments 

- Pass dependencies as arguments to functions to increase ease of testing

### DIFF
--- a/app/assets/javascripts/header.js
+++ b/app/assets/javascripts/header.js
@@ -15,7 +15,7 @@ var onReadyHeader = function() {
     }
   });
 
-  $('.expand_moment_button').mouseover(expandMomentMouseover);
+  $('.expand_moment_button').mouseover(click_flag,expandMomentMouseover);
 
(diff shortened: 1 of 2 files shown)


julianguyen:
LGTM 👍 Can someone else take a look too?

sienatime:
I think this all looks great! This can be a separate issue, but I want to get away from the `click_flag` thing, as opening/closing elements can be handled just by toggling classes or inspecting whatever `display` is instead of tracking how many times they have been clicked and doing math...

sienatime:
Added new issue for removing `click_flag`: https://github.com/julianguyen/ifme/issues/231

PR #232, opened 2016-07-03
Refactoring groups#update

kkelleey:
Extracting update_leader helper method. ABC of update method is now ~16

The update_leader method is still too big, I'll work on breaking that up next

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -85,48 +85,14 @@ class GroupsController < ApplicationController
   # PATCH/PUT /groups/1
   # PATCH/PUT /groups/1.json
   def update
-    @page_title = "Edit " + @group.name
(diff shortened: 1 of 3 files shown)


sienatime:
Cool! Thoughts on putting the update logic in its own PORO/class? Kinda nice to just get all that stuff out of controllers. Otherwise LGTM!

kkelleey:
I agree that it doesn't belong in the controller, I just wanted to move it out of the update method for starters. I *think* I might be able to abstract some sort of GroupNotifier that takes an array of people to be notified (accepted_allies in groups#create, group.leaders in groups#update, etc) and sends the appropriate notifications. That's the dream.

kkelleey:
^^ Dream accomplished. #234

There will be merge conflicts between these, but I'll try to resolve them once the first is merged.

PR #233, opened 2016-07-03
Adding latest version of schema

kkelleey:
When I ran the latest migration, the schema was updated. It looks like the the schema changes weren't committed after 20160512174202_set_limits_on_text.rb ran.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -26,20 +26,20 @@ ActiveRecord::Schema.define(version: 20160518220139) do
 
   create_table "bootsy_image_galleries", force: :cascade do |t|
     t.integer  "bootsy_resource_id"
-    t.string   "bootsy_resource_type", limit: 255
+    t.string   "bootsy_resource_type"
     t.datetime "created_at"
     t.datetime "updated_at"
   end
 
   create_table "bootsy_images", force: :cascade do |t|
-    t.string   "image_file",       limit: 255
+    t.string   "image_file"
     t.integer  "image_gallery_id"
     t.datetime "created_at"
     t.datetime "updated_at"
   end
 
   create_table "categories", force: :cascade do |t|
-    t.string   "name",        limit: 255
+    t.string   "name"
     t.text     "description"
     t.datetime "created_at"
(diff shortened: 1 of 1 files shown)


sienatime:
I actually don't get these changes when I update the schema... I wonder if there's a version difference? Can someone else try? Also the columns mentioned in the migration you pointed out don't seem to corroborate with the changes here. Thanks!

Rails 4.2.6
Rake 11.1.2

stellacotton:
If there's been a Rails upgrade on the app, it may be related to this: https://github.com/rails/rails/issues/19001.

If the columns on the database don't have the limit set (which it seems like maybe not), I think merging in the schema change without the limits is the right thing to do since it's probably driving everyone a little nuts and it is technically accurate. 

On the other hand, these columns probably could use some sensible limits but I don't think its critical.

kkelleey:
Hmm... weird that yours keeps the the limits @sienatime, but good catch that the columns don't match the migration that I referenced. I should have checked more closely.

Good find @stellacotton. I tried branching a new branch from master and running `rake db:drop db:create db:migrate` as they suggested, and it removed the limits from the schema again. I'm using Rails 4.2.6 and Rake 11.1.2. 

Digging deeper... [Here is the PR](https://github.com/rails/rails/pull/14579) that introduced the change in Rails. Interestingly, they made the change because specifying the limit has no performance advantage in Postgres (in fact its usually is slower).

If this isn't happening for anyone else, we can close this.

stellacotton:
It is happening to me and also continues to happen when I run `rake db:drop db:create db:migrate` as well. It seems like removing the limit from the schema is the right thing since the columns don't actually have the limit in the db, but I can't tell if that's just going to mean that @sienatime will have the opposite problem. :)

sienatime:
Ah haaaa, if I do `rake db:drop db:create db:migrate` I get them... I tried a `rake db:reset` like one of the threads suggested and no dice. So yeah this is probably fine to merge. 👍 Thanks for the additional investigation @stellacotton!

PR #234, opened 2016-07-04
Extracting GroupNotifier service object

kkelleey:
This change removes duplicated code in the join, update and create
methods of the groups_controller by extracting a service object to
notify the relevant members of each group.

Testing the GroupNotifier was a little messy because of how much
interaction there was with collaborators, so I ended up using a lot of
mocks.


sienatime:
Sweet, this looks great. I guess you could wrap each of the collaborators in another class so that you could inject the intermediate classes in the test instead of mocking the actual Thing but.. probably overkill for now. I merged your other PR so you can deal with the merge conflicts now :)

kkelleey:
👍  Thanks for reviewing! Merge conflicts fixed

PR #235, opened 2016-07-05
Adding title helper

kkelleey:
Before this commit, @page_title was being set in the controller so that the
value was accessible in the application layout. IMO the logic for
setting the title really belongs in the view instead of the controller.
[This Railscast](http://railscasts.com/episodes/30-pretty-page-title?autoplay=true) shows a clever way to do this through an application helper method. This commit
implements this helper and uses it for all groups views. I was thinking
we could slowly update the rest of the views, but for now the layout
falls back to @page_title if title isn't set in the view.

### DIFF
--- a/app/views/layouts/_title.html.erb
+++ b/app/views/layouts/_title.html.erb
@@ -1,0 +1,28 @@
+<%= t(:app_name) %>
+<% if current_page?(new_user_session_path) ||
+    (params[:controller] == "devise/sessions" && action_name == "new") %>
+  - Sign in
+<% elsif current_page?(new_user_registration_path) ||
+  (params[:controller] == "devise/registrations" && action_name == "create") %>
+- Sign up
+    <% elsif current_page?(new_user_password_path) ||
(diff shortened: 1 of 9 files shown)


app/views/layouts/application.html.erb
```diff
@@ -1,194 +1,167 @@
...
-					<%= yield %>
-				</div>
-
-				<%= render partial: "shared/footer" %>
-			</div>
-		</div>
-	</body>
+  <head>
+    <title>
+      <%= render 'layouts/title' %>
```
kkelleey: This is the only real change in this file, the rest is just auto-indenting

app/views/layouts/_title.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= t(:app_name) %>
+<% if current_page?(new_user_session_path) ||
```
kkelleey: A follow-up to this PR is to add <%title %>s to all of these devise views so that we can get rid of all of this logic

julianguyen:
LGTM! Great work as always @kkelleey ^^

PR #236, opened 2016-07-05
Using built-in :authenticate_user! before_action

kkelleey:
The user_signed_in method duplicated the functionality of
the authenticate_user! helper provided by Devise. The authenticate_user!
method also also provides a flash 'You need to sign in before continuing',
so I think it's better to just use that.

This change initially caused the controller specs to fail because the
stub_current_user wasn't effectively stubbing the authentication, so I
tweaked the method using the example
[here](https://github.com/plataformatec/devise/wiki/How-To:-Stub-authentication-in-controller-specs)


julianguyen:
Great refactor @kkelleey 👍

PR #237, opened 2016-07-06
Adding after_destroy callback to destroy notifications

kkelleey:
Ideally we could have a dependent: :delete_all association on group, but
I couldn't quite figure out how to define the has_and_belongs_to
association because the groupid is in the json string in the data field,
so I implemented it manually, defining group#notifications and an
after_destroy callback.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -19,7 +19,21 @@ class Group < ActiveRecord::Base
   has_many :leaders, -> { where(group_members: { leader: true }) },
            through: :group_members, source: :user
 
+  after_destroy :destroy_notifications
+
   def led_by?(user)
     leaders.include? user
+  end
+
+  def notifications
+    Notification.where('uniqueid ilike ?', '%new_group%').select do |n|
+      JSON.parse(n.data)['groupid'].to_i == id
+    end
+  end
+
+  private
+
+  def destroy_notifications
+    notifications.each(&:destroy)
   end
 end
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -127,13 +127,6 @@ class GroupsController < ApplicationController
   def destroy
     @group.destroy
 
-    # Delete notifications for this group
-    Notification.where("uniqueid ilike ?", "%new_group%").all.each do |notification|
-      if JSON.parse(notification.data)["groupid"].to_i == @group.id.to_i
-        notification.destroy
-      end
-    end
-
     respond_to do |format|
       format.html { redirect_to groups_path }
       format.json { head :no_content }
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -10,4 +10,5 @@
 
 class Notification < ActiveRecord::Base
   validates_presence_of :userid, :uniqueid, :data
+  belongs_to :user, foreign_key: :userid
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,6 +49,7 @@ class User < ActiveRecord::Base
   has_many :groups, through: :group_members
   has_many :meeting_members, foreign_key: :userid
   has_many :medications, foreign_key: :userid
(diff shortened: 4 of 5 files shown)


app/models/group.rb
```diff
@@ -19,7 +19,21 @@ class Group < ActiveRecord::Base
...
            through: :group_members, source: :user
 
+  after_destroy :destroy_notifications
+
   def led_by?(user)
     leaders.include? user
   end
+
+  def notifications
+    Notification.where('uniqueid ilike ?', '%new_group%').select do |n|
```
kkelleey: It seems like there should be a way to directly query for this instead of doing one query and then selecting, but I couldn't figure it out.

stellacotton: I'm not super familiar with the notifications, but it sounds like there is a json store in uniqueid that looks something like `comment_on_strategy_1`? 

Can we include the id in the query?
```query_string = "new_group_#{groupid}"
`Notification.where('uniqueid ilike ?', '%query_string%')```

**Comment by stellacotton** [comment]
One question about the query but also a meta-question since it would be amazing to not have that query at all... 😀

Are there any plans to refactor the way the notifications are structured? I like this PR because it moves us in a direction of having more Rails-y associations but the schema behind notifications is very brittle (both from a performance standpoint and from an extensibility standpoint).

File: app/models/group.rb (hunk 1 above)
Comment (kkelleey) [inline app/models/group.rb]: It looks like we've been storing uniqueid as "new_group_#{current_user.id}", not "new_group_#{groupid}" (see services/group_notifier.rb:48) :(

Here is an example of an instance of Notification
```
#<Notification:0x007ff28744ad50
  id: 2,
  userid: 2,
  uniqueid: "new_group_member_4",
  data:
   "{\"user\":\"Karen\",\"groupid\":1,\"group\":\"Students with Anxiety\",\"type\":\"new_group_member\",\"uniqueid\":\"new_group_member_4\"}",
  created_at: Mon, 04 Jul 2016 16:30:28 UTC +00:00,
  updated_at: Mon, 04 Jul 2016 16:30:28 UTC +00:00>
```

I played around with it a little bit in the console and this works: 

```ruby
query_string = "groupid\":#{groupid},"
Notification.where('data ilike ?', "%#{query_string}%")
```

I don't like the `\":` piece, and it seems like it could potentially match data strings that it's not supposed to (the `,` is included at the end of the query string to ensure that `groupid: 1` doesn't also match `groupid: 11`, for example), although this does seem pretty unlikely considering the contents of the rest of the string. 

I think both solutions are about the same in terms of elegance, the former is more efficient while the latter is more rigorous.

However, I don't even think the latter allows us to achieve a group has_many :notifications association either because it still needs a foreign key.

kkelleey:
I agree that it would be nice to refactor notifications. I haven't worked much with them outside of the context of groups, so I'd have to look into how else they are used.

As a continuation of my other comment, it would be nice if it had a group_id column, but that wouldn't be relevant to comment notifications. Or we could change the format of uniqueid so that it is `new_group_#{groupid}`, as you had said, but then it might not necessarily be unique? (Not sure if that's important)

julianguyen:
"Another possible improvement: make data a jsonb type instead of text, which would make the query more elegant, but we still wouldn't have the foreign key for the association." <--- I like this refactor idea 👍 

"Or we could change the format of uniqueid so that it is new_group_#{groupid}, as you had said, but then it might not necessarily be unique? (Not sure if that's important)" <--- I think we should keep it unique just in case.

Your changes LGTM and I agree with @stellacotton  it's a step in the right direction. Does anyone object to merging it?

stellacotton:
Nope, no objections! 

Sorry I haven't had a chance to dig into the possible future refactoring, but definitely interested in chatting about it further if we think it makes sense to do more. If this code isn't going to change much more than this and there aren't new notification features in the pipeline, we can probably just push off doing additional data model changes to when that happens.

PR #239, opened 2016-07-11
Fixing page title in layout

kkelleey:
I didn't realize that there were more instances of @page_title when I made the title helper in #235

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -13,6 +13,7 @@ class CategoriesController < ApplicationController
       @categories = Category.where(:userid => current_user.id).all.order("created_at DESC").page(params[:page]).per($per_page)
     end
     @page_new = new_category_path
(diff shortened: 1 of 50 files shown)


app/controllers/categories_controller.rb
```diff
@@ -41,14 +38,11 @@ def show
...
-    @page_title = "New Category"
   end
 
   # GET /categories/1/edit
   def edit
-    if @category.userid == current_user.id
-      @page_title = "Edit " + @category.name
-    else
+    if @category.userid != current_user.id
       respond_to do |format|
```
kkelleey: There were a few like this... I ended up with and empty if block once I removed the `@page_title` declaration, so I reversed the logic of the if statement and got rid of the else

julianguyen:
LGTM! Thanks for the fix @kkelleey ^^

PR #241, opened 2016-07-15
Refactoring update_leaders method

kkelleey:
The ABC of the update_leaders method was too high before this commit.
The method is still pretty long, but the ABC is now below 15.

This version also does one update_all for adding leaders and one for
removing, while the old one looped through each member and updated
individually.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -147,23 +147,9 @@ class GroupsController < ApplicationController
   end
 
   def update_leaders
-    return if params[:group][:leader].nil?
+    updated_leader_ids = params[:group][:leader]
+    return if updated_leader_ids.nil?
 
-    updated_leader_ids = params[:group][:leader].map(&:to_i)
-    current_leader_ids = @group.leader_ids
-
-    leaders_to_add = updated_leader_ids - current_leader_ids
-    leaders_to_remove = current_leader_ids - updated_leader_ids
-
-    update_leader_status(leaders_to_add)
-    update_leader_status(leaders_to_remove, false)
-  end
-
-  def update_leader_status(leaders, leader = true)
-    return if leaders.empty?
-    @group.group_members.where(userid: leaders).update_all(leader: leader)
-    pusher_type = leader ? 'add_group_leader' : 'remove_group_leader'
-    GroupNotifier.new(@group, pusher_type, current_user)
-                 .send_notifications_to(@group.leaders)
+    LeaderUpdater.new(@group, updated_leader_ids.map(&:to_i)).update
   end
 end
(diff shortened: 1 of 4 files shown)


kkelleey:
I feel like this isn't a perfect solution. I don't like that the notifications are being sent from the update_leader_status method, it seems like they should be separate. Also it's confusing that the  methods are called update_leaders and update_leader_status.

I might try to extract a service object here, which keep more logic out of the controller.

app/services/leader_updater.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def remove_former_leaders
+    group_memberships_for(former_leader_userids)
+      .update_all(leader: false)
+  end
+
+  def notify_leaders(user_ids, pusher_type)
+    leaders_before_update = User.where(id: @leader_ids_before_update)
+    User.where(id: user_ids).each do |user|
+      GroupNotifier.new(@group, pusher_type, user)
```
kkelleey: @julianguyen before this, user was being set to current_user, but I think that's not what we want? It's possible that I accidentally set it to current_user somewhere along the line. 

With the way it was  before, the messages were all saying `"#{whoever was current user} has been made a leader"`, but I think it should be `#{some member who now has leader = true} has been made a leader"`

The code above reflects the latter now, let me know if that's not what you want.

app/controllers/groups_controller.rb
```diff
@@ -147,18 +147,9 @@ def group_member_params
...
-        pusher_type = 'add_group_leader'
-      else
-        group_member.update(leader: false)
-        pusher_type = 'remove_group_leader'
-      end
+    updated_leader_ids = params[:group][:leader]
+    return if updated_leader_ids.nil?
 
-      GroupNotifier.new(@group, pusher_type, current_user)
-                   .send_notifications_to(@group.leaders)
```
kkelleey: Are the notifications supposed to be sent to the people who were leaders before the changes were made or after?

app/services/leader_updater.rb
julianguyen: Good catch @kkelleey!

app/controllers/groups_controller.rb
julianguyen: @kkelleey: It should be sent to people who are leaders after the changes.l

julianguyen:
I agree update_leaders and update_leader_status methods should be combined. I liked that you put things out into a service. 

LGTM, I can merge.

app/controllers/groups_controller.rb
kkelleey: ok! I'll have change the LeaderUpdater

kkelleey: It's updated now

Issue #245, opened 2016-07-25
Travis CI needs to run Jasmine tests


julianguyen:
Thanks @sienatime for noticing that Travis CI wasn't actually running the Jasmine tests. Oof. 

@starbelly originally set up Travis CI here: https://github.com/julianguyen/ifme/pull/198/files

I modified the Rakefile to run both rspec and Jasmine, but it is not working. https://github.com/julianguyen/ifme/commit/b21d3f31a283b4c97be2018d8ef126d1e793d987

Help is appreciated :)

julianguyen:
Fixed by @starbelly! Thank you ^^

starbelly:
@julianguyen No problem : )

PR #246, opened 2016-07-26
 Get jasmine test units working in travis-ci and passing

starbelly:
Fixes issue #245 

  This will get your jasmine test units going in travis-ci : ) Note: You had everything right in your Rakefile but travis-ci defaults to running 'rake' and not 'rake travis' so just needed to add that to .travis.yml. Three tests in your header spec were not passing though. This was because the css opacity value was a float with 15 digits of precision and not a string. I modified those tests to round to the first decimal place using Math.round and changed the expectation from a string to a float. 

Cheers


julianguyen:
Thank you soooooo much @starbelly 💯 💯 💯

PR #247, opened 2016-07-26
New setup script

kkelleey:
This PR has two commits. The first copies the setup script that I found [here](https://gist.github.com/gvarela/9370838) into bin/setup_new as is. The second customizes that script to more closely fit our app. I separated them like this so that it would be clear what was mine and what I got from the gist.

This is meant to be kind of a WIP, if we decide it looks ok I'll replace bin/setup with it and update the README. 

This is meant to be kind of a WIP, if we decide it looks ok I'll replace bin/setup with it and update the README.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,29 +1,162 @@
-#!/usr/bin/env ruby
-require 'pathname'
+#!/usr/bin/env bash
+# taken from https://gist.github.com/gvarela/9370838
 
-# path to your application root.
-APP_ROOT = Pathname.new File.expand_path('../../',  __FILE__)
+{
+  RUBY_VERSION=$(cat .ruby-version)
+  RUBY_VERSION=${RUBY_VERSION:=2.3.1}
+  PROJECT_NAME="ifme"
+} >/dev/null 2>&1
 
-Dir.chdir APP_ROOT do
-  # This script is a starting point to setup your application.
-  # Add necessary setup steps to this file:
+## bootstrap function called at bottom of file comment out verification steps you don't need
+init (){
+  verify_ruby_version &&
+    verify_bundled_gems
+  copy_example_config_files
+  verify_postgres_running &&
+    verify_database_migrations
+}
 
-  puts "== Installing dependencies =="
-  system "gem install bundler --conservative"
-  system "bundle check || bundle install"
 
-  # puts "\n== Copying sample files =="
-  # unless File.exist?("config/database.yml")
-  #   system "cp config/database.yml.sample config/database.yml"
-  # end
+has_postgres (){
+  hash psql >/dev/null 2>&1
+}
 
-  puts "\n== Preparing database =="
-  system "bin/rake db:setup"
+postgres_is_running (){
+  ps aux | grep [p]ostgres >/dev/null 2>&1
+}
 
-  puts "\n== Removing old logs and tempfiles =="
-  system "rm -f log/*"
-  system "rm -rf tmp/cache"
+has_rvm (){
+  hash rvm >/dev/null 2>&1
+}
 
-  puts "\n== Restarting application server =="
-  system "touch tmp/restart.txt"
-end
+update_rvm (){
+  rvm get latest
+}
+
+rvm_install_ruby_version () {
+  rvm install $RUBY_VERSION
+}
+
+has_rbenv (){
+  hash rbenv >/dev/null 2>&1
+}
+
+rbenv_install_ruby_version () {
+  rbenv install $RUBY_VERSION
+  rbenv local $RUBY_VERSION
+}
+
+has_ruby_version (){
+  INSTALLED_RUBY_VERSION=$(ruby -e 'puts RUBY_VERSION')
+  [ "$INSTALLED_RUBY_VERSION" == "$RUBY_VERSION" ]
+}
+
+install_ruby_version (){
+  if has_rvm
+  then
+    echo "Installing Ruby Version $RUBY_VERSION with rvm"
+    update_rvm
+    rvm_install_ruby_version
+  elif has_rbenv
+  then
+    echo "Installing Ruby Version $RUBY_VERSION with rbenv"
+    rbenv_install_ruby_version
+  else
+    echo 'Please install rbenv or rvm to manage your ruby versions.'
+    exit 1
+  fi
+}
+
+verify_ruby_version (){
+  has_ruby_version || install_ruby_version
+}
+
+has_bundler (){
+  hash bundle >/dev/null 2>&1
+}
+
+verify_bundled_gems () {
+  has_bundler || install_bundler
+  bundle_check || bundle_install || {
+    check_pg || pg_message
+    echo 'Unable to install all dependencies from Gemfile. Make sure `bundle install` succeeds before continuing.'
+    exit 1
+  }
+}
+
+install_bundler () {
+  echo 'Installing bundler...'
+  gem install bundler || {
+    echo 'Unable to install bundler.'
+    exit 1
+  }
+}
+
+bundle_install () {
+  echo 'Installing dependencies specified in Gemfile...'
+  bundle install
+}
+
+check_pg () {
+  gem list -i pg >/dev/null 2>&1
+}
+
+pg_message () {
+  printf "Unable to install pg gem.\n\n
+  If you are using homebrew, try `brew update` and `brew install
+  postgresql`.\n\n
+  If you are using Ubuntu, try `sudo apt-get install
+  libpq-dev`"
+}
+
+bundle_check () {
+  bundle check >/dev/null 2>&1
+}
+
+verify_postgres_running () {
+  echo "Verifying postgres is running..."
+  has_postgres || {
+    echo 'Please install postgresql.'
+    exit 1
+  }
+
+  postgres_is_running || {
+    echo 'Please ensure postgresql is running'
+    exit 1
+  }
+}
+
+verify_database_migrations () {
+  db_setup || {
+    echo 'Unable to setup the databases for you. Please ensure your
+    database.yml is configured for your system and run `rake db:setup`'
+    exit 1
+  }
+}
+
+db_setup () {
+  echo 'Creating the database, loading the schema and initializing it with the seed
+  data'
+  bundle exec rake db:setup >/dev/null 2>&1
+}
+
+copy_example_config_files () {
+  echo "Copying example config files..."
+  if [ ! -f config/application.yml ]
+  then
+    cp config/application.example.yml config/application.yml
+  fi
+  if [ ! -f config/smtp.yml ]
+  then
+    cp config/smtp.example.yml config/smtp.yml
+  fi
+  if [ ! -f config/database.yml ]
+  then
+    cp config/database.example.yml config/database.yml
+  fi
+}
+
+# start it up!!
+init
+echo 'Ready!'
+echo 'Run `rails s` to start the server'
--- a/README.md
+++ b/README.md
@@ -64,153 +64,28 @@ Getting Started
 
 The app uses  **Ruby 2.3.1** and **Rails 4.2.6**. Please stick to these versions.
 
-### Installing Programs
+### Installing Postgres
 
-The steps below should be straightforward for Linux and OSX users. Windows users please refer to this [guide](https://gist.github.com/KelseyDH/11198922) for tips on setup.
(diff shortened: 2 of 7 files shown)


kkelleey:
@stellacotton could you take a look at this if you get a chance?

kkelleey:
Looks like the bundle/bin business has to do with binstubs. I'll investigate further later

bin/setup_new
```diff
@@ -0,0 +1,173 @@
...
+postgres_is_running || {
+echo 'Please ensure postgresql is running'
+exit 1
+  }
+}
+
+verify_database_configuration () {
+  if [ ! -f config/database.yml ]
+  then
+    cp config/database.travis.yml config/database.yml
```
stellacotton: I think our db example file is called `config/database.example.yml`?

bin/setup_new
```diff
@@ -0,0 +1,173 @@
...
+    (gem list -i bundler || gem install bundler) && bundle install
+  } 2>&1
+}
+
+verify_env_file (){
+  if [ ! -f .env ]
+  then
+    echo "Generating secret..."
+    SECRET=$(bundle exec rake secret)
+    echo "SECRET_KEY_BASE=$SECRET" >> .env
```
stellacotton: I think we want to append this to `config/application.example.yml` instead of `.env` since we're using Figaro for our secrets management.

bin/setup_new
```diff
@@ -0,0 +1,173 @@
...
+    bundle exec rake db:setup db:test:prepare
+  fi
+}
+
+copy_example_config_files () {
+  echo "Copying example config files..."
+  cp config/application.example.yml config/application.yml
+  cp config/smtp.example.yml config/smtp.yml
+  cp config/database.example.yml config/database.yml
+}
```
stellacotton: It might be nice to check and see if these exist already before copying them over?

eg on line 139:
`if [ ! -f config/database.yml ]`

bin/setup_new
```diff
@@ -0,0 +1,173 @@
...
+      echo 'Please install homebrew. Install instructions at http://brew.sh/'
+    else
+      echo "Please install Ruby version $RUBY_VERSION"
+    fi
+    exit 1
+  fi
+}
+
+has_postgres (){
+  hash psql >/dev/null 2>&1
```
stellacotton: My bash is a little iffy- do you know what it's doing when it's calling `hash [program name]`? I know the `>/dev/null 2>&1` is for quieting the output but I'm not totally getting what's happening in the rest of the command.

stellacotton:
Really excited about this setup script! It will be so nice for new contributors to get up and running with minimal setup. :D

I haven't tried to run it locally yet, since I don't want to blow away my current application.yml but I can try it out if we switch the `copy_example_config_files` to be idempotent.

bin/setup_new
kkelleey: I didn't know the answer to that either, so I did some research and according to [this stackoverflow](http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script)  `hash` is the recommended way to check that a program exists when `bash` is your shell/hashbang (as it is in our case). It is similar to `which`, except it is better to use `hash` because:

* `which` is an external process, so it is cheaper to use hash because it is builtin
* Many operating systems have a `which` that doesn't even set an exit status
* Many operating systems make `which` do custom and evil stuff like change the output or even hook into the package manager.
* `hash` has the side-effect that the command's location will be hashed (for faster lookup next time you use it), which is usually a good thing since you probably check for its existence in order to actually use it.

bin/setup_new
kkelleey: Good catch. I'm so used to using dotenv that I didn't realize that we actually needed the dotenv gem for that to work.

Appending to the config file makes me worry that the secret will be appended more than once if for some reason if the setup script is run more than once (which would probably only happen if the script were unsuccessful and needed to be rerun). I guess appending `SECRET_KEY_BASE: "#{generated secret key}"` more than once wouldn't break anything, I assume the last one would just override any previous declarations.

[This post](http://daniel.fone.net.nz/blog/2013/05/20/a-better-way-to-manage-the-rails-secret-token/) suggests hardcoding the secret key for dev and test environments, since it's really only production that needs a secure key.

```
# config/initializers/secret_token.rb

MyApp::Application.config.secret_token = if Rails.env.development? or Rails.env.test?
  ('x' * 30) # meets minimum requirement of 30 chars long
else
  ENV['SECRET_TOKEN']
end
```

Or since Figaro allows environment-specific declarations we could add hard-coded dev and test secrets to `config/application.example.yml`

```
development:
  SECRET_KEY_BASE: "#{something_that_is_30_characters}"
```

bin/setup_new
kkelleey: good catch!

kkelleey: Actually now that you mention it I have this `cp` in copy example_config_files, which is run before this, so this check is redundant.

bin/setup_new
kkelleey: 👍

kkelleey:
Updated.

Waiting to change the secret key part until we agree on a solution.

An interesting note on this, I read through the [PR for bin/setup in Rails](https://github.com/rails/rails/pull/15189), and one of DHH's comments says

>If you already have the database setup, then you already have the app
running, then bin/setup is not intended for you. We have rake db:migrate
and bundle to deal with those. I think bin/setup as "clean slate" is what
we're trying to achieve with this.

This is interesting to keep in mind. I think adding the `if [ ! -f config/database.example.yml ]` is still a good idea to protect us from messing up our own environments as we develop the script, but we should assume this will only be used for initial setups as much as possible.

bin/setup_new
```diff
@@ -135,36 +135,36 @@ exit 1
...
 }
 
-verify_database_migrations () {
-  verify_database_configuration
-
-  bundle exec rake db:abort_if_pending_migrations
-  if [ $? -eq 1 ]
-  then
-    echo 'Running migrations...'
-    bundle exec rake db:setup db:test:prepare
```
kkelleey: Removing db:test:prepare as it is [deprecated in Rails 4.1](https://github.com/rails/rails/pull/13528). (The test schema is maintained automatically).

stellacotton: Ooh, I didn't know that! 💃

bin/setup_new
stellacotton: Yeah, it looks like we're already hardcoding the secret in the `config/application.example.yml` to `"examplesecretkey123123"` so I think you're right- lets just keep that as-is and remove the secret stuff from the setup script. 

I don't think we need to break it out by environment since basically all these configuration values are dev-only anyway and I think we're using the Heroku env UI to manage production secrets.

bin/setup_new
stellacotton: Oh awesome, that's a great writeup.

I'm still a little confused though- when I run `hash rvm` or `hash psql` on my machine, I don't get any output (compared to when I run `which psql` or `command -v pqsl`, where I get the path). I wonder if its because I'm using zsh?

kkelleey: I think it's not supposed to have any output upon success, just return exit code 0.

```shell
$ hash psql
$ echo $?
0
$ hash something_i_dont_have
hash: no such command: something_i_dont_have
```

bin/setup_new
kkelleey: Sounds good to me, the current one isn't 30 characters so I'll update it so that it is

bin/setup_new
stellacotton: Ahh, got it. I was expecting output in addition to the exit code. Thanks for walking me through this! :D

bin/setup_new
```diff
@@ -0,0 +1,173 @@
...
+  fi
+}
+
+copy_example_config_files () {
+  echo "Copying example config files..."
+  if [ ! -f config/database.yml ]
+  then
+    cp config/application.example.yml config/application.yml
+  fi
+  if [ ! -f config/smtp.example.yml ]
```
kkelleey: Whoa, didn't put the right file names here. Don't run this script yet!

bin/setup_new
kkelleey: Np, I learned a lot too.

bin/setup_new
```diff
@@ -21,17 +20,17 @@ init (){
...
+  [[ "$PLATFORM" == "Darwin" ]]
 }
 
 has_homebrew (){
-  [ is_darwin ] && [ hash brew >/dev/null 2>&1 ]
+  is_darwin && [ hash brew >/dev/null 2>&1 ]
 }
 
 verify_homebrew (){
-  if [ ! has_homebrew ]
```
kkelleey: This wasn't working as intended with the square brackets, and I *think* it was because the `[` command was evaluating the exit code for `! has_homebrew` instead of the result of `! homebrew`, so it was always getting true even if homebrew wasn't installed.

kkelleey:
Just made a bunch of changes after some testing on the VM. Seems like the original version wasn't as functional as I thought, but it's been a nice exercise for my Bash skills to fix it.

I redirected most of the output to dev/null except `bundle install`. I feel like since there are so many gems being installed (and so many potential errors) it would be helpful to see what is going on instead of waiting quite a while with no output.

stellacotton:
Looks awesome! Looks like you decided to go with an info message instead of trying to deal with postgres installation? I think that makes sense, especially since if the install fails, someone probably doesn't want to have to deal with re-trying by re-running the setup script over and over.

bin/setup_new
```diff
@@ -80,24 +87,16 @@ has_ruby_version (){
 }
 
 install_ruby_version (){
-  verify_homebrew
```
kkelleey: I got rid of the homebrew/darwin piece, it seemed like it wasn't really necessary

kkelleey:
Yeah exactly. Hopefully it won't fail for too many people anyway.

Thanks for following along so far :) Just added another change rewriting the rvm piece. I'm not sure the rbenv part is working either, I'm going to take a look at that next.

julianguyen:
Just a progress check, how's this going @kkelleey? :)

kkelleey:
I think it's close, I just need to put the finishing touches on it. Hoping
to do that this weekend/early next week.

kkelleey:
I think the script is mostly good to go, the only thing I can think of that it doesn't handle is setting the postgres password. Do you think we should prompt the user for it?

```bash
echo "Postgres username:"
read username
echo "Postgres password:"
read password

PSQL_USERNAME=$username
PSQL_PASSWORD=$password
```

And then in the default database.yml add (as it suggests in the README)

```yml
development: &default
  ...
  username: <%= ENV["PSQL_USERNAME"] %>
  password: <%= ENV["PSQL_PASSWORD"] %>
```

julianguyen:
@kkelleey: Ooh, that's a good enhancement! Also wondering, would it be possible to not set a password on postgres?

kkelleey:
Hmm I was getting errors for not having it when I tested on an ubuntu VM but maybe it's not always required? Also I just realized the rbenv piece isn't quite working yet, so TODOS

- [ ] Fix rbenv piece
- [ ] Update README to say that they need to have either rbenv or rvm

kkelleey:
Haven't forgotten about this... I just moved and started a new job so I haven't had much extra time recently. Will hopefully have time to work on it next week

julianguyen:
Congrats @kkelleey on the new job! 👍  Sounds good, keep us posted :)

kkelleey:
Should be ready for review now

julianguyen:
Hey @kkelleey can you rebase? :)

bin/setup
```diff
@@ -1,29 +1,162 @@
-#!/usr/bin/env ruby
-require 'pathname'
+#!/usr/bin/env bash
+# taken from https://gist.github.com/gvarela/9370838
 
-# path to your application root.
-APP_ROOT = Pathname.new File.expand_path('../../',  __FILE__)
+{
+  RUBY_VERSION=$(cat .ruby-version)
+  RUBY_VERSION=${RUBY_VERSION:=2.3.0}
```
julianguyen: We've upgraded to `Ruby 2.3.1` :)

julianguyen:
Hey @kkelleey, we recently set up dotenv to load the environment variables (see https://github.com/julianguyen/ifme#mandatory). Let me know if you can update this PR :)
Thanks!

mtrberzi:
I'm really sorry, this got put on the backburner for a while! Would you still like me to help review when this is ready? Let me know.

PR #248, opened 2016-07-27
Redirecting to groups path if nonleader edits group

kkelleey:
Before this commit, anyone could visit the edit group page directly via
URL. Although they are unlikely to do this, we should still redirect if
the authenticated user is not a leader of the group.

Other notes:
* The @group_members variable wasn't being used, I think I had switched
the view to use @group.members earlier and forgot to clean up the
controller action
* I extracted the redirect_to_index helper and added a check on
user_leaves_group feature spec to make sure it works

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -33,7 +33,10 @@ class GroupsController < ApplicationController
 
   # GET /groups/1/edit
   def edit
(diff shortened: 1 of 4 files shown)


julianguyen:
@kkelleey Omg thanks for catching this. Oh boy silliness. 💯

PR #252, opened 2016-08-07
Only run metrics for Ruby and JavaScript

gbonfant:
We don't need to add configuration details for other languages.


julianguyen:
Good catch @gbonfant 💯

PR #253, opened 2016-08-11
Moving logic for page new from controller to view

kkelleey:
This commit is similar to #235

The logic for setting the path for the page_new button should be in the
view, not the controller. This commit adds a page_new helper in
application helper so that the path can be declared from the view


kkelleey:
Code climate is still complaining about some repeated code, but I feel like fixing that the right way would be beyond the scope of this PR.

julianguyen:
LGTM! Awesome work as always @kkelleey :)

PR #256, opened 2016-08-20
Siena/fully qualify i18n

sienatime:
This PR does a few things:
* Turn on exceptions for missing translations in dev and test
* Fixes the indentation for notifications in en.yml in order to fix a failing test
* Removes several app_name keys that were unnecessary
* Uses the fully qualified key names of all currently existing translated keys so that we can more easily identify areas of overlap, and find existing uses of keys
* Removes some whitespace from some files (my editor does that automatically, don't think it should be a problem)

Commits can be squashed after review.


julianguyen:
LGTM 💯

PR #257, opened 2016-08-20
Use pluralization where possible

sienatime:
Use Rails i18n pluralization for labels like this

screen shot 2016-08-20 at 4 42 00 pm
screen shot 2016-08-20 at 4 42 56 pm

Also made sure that the count was included in the string instead of right next to it.


julianguyen:
LGTM!

PR #258, opened 2016-08-21
spec for viewers indicators in a strategy spec; refactored viewer ind…

sienatime:
…icators to be more i18n-friendly

I started this refactor by writing some acceptance tests for viewing a strategy, since that is one of the places where the viewer_indicator partial is used. I'm not crazy about testing a partial in this way but it seemed better than nothing so I would have a baseline to run tests against while I refactored. I was also trying to avoid using a helper since I thought the if/else blocks could be simplified enough for a view.

This also assumes that the controller will take care of permissions issues, which I think is fine (no need to check that, if you are not logged in as the strategy owner, that the ID of the one viewer matches the current_user's ID).


julianguyen:
LGTM. We should also duplicate the test specs for Moments and Meetings since they use viewers_indicator as well :)

sienatime:
Yeah there is definitely a better testing solution than duplicating these tests 3 times... just not sure what it is yet. Testing a partial by itself seems gross but maybe it's.. not? Or maybe I should just use a helper.. ha

PR #261, opened 2016-08-28
Siena/viewers indicator spec

sienatime:
This PR does a few things:

* Enables Selenium for running acceptance specs with Javascript (`js: true`). This is useful for writing specs that test entire end-to-end interactions with the site, and they are run in a real browser.
* Consolidates rails_helper config into spec_helper. It's useful to have all this config in the same place, and in some instances we actually had the same configuration defined twice.
* Uses the database_cleaner gem and database truncation as our cleanup strategy for acceptance specs. This is a recommended best practice for Selenium and Capybara, and you can read more about it from Avdi Grimm here: http://www.virtuouscode.com/2012/08/31/configuring-database_cleaner-with-rails-rspec-capybara-and-selenium/

All of this is in service of the following refactor:
* Moves viewers_indicators logic into a unit-tested helper file
* Adds acceptance specs for creating/viewing/editing a strategy and a moment, where viewers indicators are included, to make sure that the Helper is being used properly


sienatime:
If anyone knows what to do about the `Missing frozen string literal comment.` rubocop failure, let me know. After some googling, think it could be okay to ignore, but maybe we should ignore it at more of a global level, which I'd be happy to include in this PR if that is the most appropriate course of action. We have a looooottttt of rubocop failures so I'm not super inclined to fix this now.

julianguyen:
@tienyuan can you shed any light on the robocop failure Siena is talking about?

spec/controllers/errors_controller_spec.rb
```diff
@@ -1,5 +1,3 @@
-require 'rails_helper'
```
julianguyen: Just wondering why you removed rails_helper?

sienaigg: As far as I could tell, rails_helper and spec_helper served the same purpose, so I moved everything that was in rails_helper but not in spec_helper into spec_helper and got rid of rails_helper.

sienatime: As far as I could tell, rails_helper and spec_helper served the same purpose, so I moved everything that was in rails_helper but not in spec_helper into spec_helper and got rid of rails_helper.

spec/features/user_creates_moment_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  feature 'Creating, viewing, and editing a moment' do
+    it 'is successful' do
+      login_as user
+      visit moments_path
+
+      within '#page_title_content' do
+        expect(page).to have_content 'Moments'
+      end
+
+      expect(page).to have_content 'Delve deep into your moments - events and situations that affect your mental health.'
```
julianguyen: Should we use the translation key here?

spec/features/user_creates_moment_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      page.find('input[value="Update Moment"]').click
+
+      #VIEWING AFTER EDITING
+      within '#page_title_content' do
+        expect(page).to have_content 'My new moment'
+      end
+      expect(page).to have_content 'I am changing my moment why description'
+    end
+  end
+end
```
julianguyen: Thank you so much for writing this spec! 💯

spec/spec_helper.rb
```diff
@@ -30,7 +35,7 @@
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  config.use_transactional_fixtures = false
```
julianguyen: Why set this to false?

spec/features/user_creates_moment_spec.rb
sienatime: Well.. we could.. other feature specs assume everything is in English too so if we're going to take that stance, it would be nice to apply it to all feature specs since there aren't that many right now. I think it's a trade-off, using the key means that if you change the translation, you don't have to update the test, but it also makes the tests a little less readable so it adds time to one's development cycle.

spec/spec_helper.rb
sienatime: Did you take a look at this article I linked to? It explains why http://www.virtuouscode.com/2012/08/31/configuring-database_cleaner-with-rails-rspec-capybara-and-selenium/

spec/features/user_creates_moment_spec.rb
julianguyen: I'm fine with using a key!

julianguyen:
LGTM. Not really sure what this [error](https://codeclimate.com/github/julianguyen/ifme/pull/261) means, but I think we can ignore it?

PR #263, opened 2016-08-28
Add updated wiki screenshots

jfriestedt:
I saw that the Features section of the Wiki still refers to "Triggers", which seems to be deprecated & renamed to "Moments." I replaced all wiki screenshots which reference the deprecated “Triggers” resource to now reference “Moments” instead. I also renamed the screenshots directory to "wiki_screenshots", and renamed all screenshots to have more descriptive file names.

If & when this change is merged, the wiki's text will also need to be updated to reflect the renaming of "Triggers" to "Moments, and the image links will need to be updated with the new directory and file names.


julianguyen:
Thanks for this @jfriestedt 💯

PR #270, opened 2016-09-12
Siena/add missing translations

sienatime:
Pulling out more stuff for translation. This is not all of the views but figured it's better to do this stuff in batches.


julianguyen:
LGTM! Feel free to merge! Thanks Siena :)

sienatime:
code climate doesn't like the TODO but i'm going to take care of 'em real soon, promise.

PR #276, opened 2016-09-24
Display leader names correctly on group show page

damiong28:
When browsing the app I noticed that group leader names were displaying as raw html instead of actual links on the group show page. So it would display as 

`"Led by: <a href=user_show_path>User</a>" `

I modified the user_visits_group_show spec to test that the names would display correctly and added .html_safe to the group show page so that the links would display as intended.

### DIFF
--- a/app/views/groups/show.html.erb
+++ b/app/views/groups/show.html.erb
@@ -19,7 +19,7 @@
     <%= t('groups.show.led_by') %>
   </strong>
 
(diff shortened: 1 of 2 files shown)


julianguyen:
OMG nice catch @damiong28! Thanks for this fix! 💯 

Also feel free to email join.ifme@gmail.com so we can add you to our Slack :D

PR #281, opened 2016-09-26
Tests for allies notifications

chrisrp:
* Adds simplecov
* Adds tests for allies notification
* minor refactoring on notification_mailer

To check the coverage run the tests and check the generated folder:
`coverage/index.html`


app/mailers/notification_mailer.rb
```diff
@@ -1,166 +1,161 @@
 class NotificationMailer < ApplicationMailer
```
julianguyen: Unrelated to this PR, but we should move the text in `en.yml`

julianguyen:
LGTM great work @chrisrp :D
Also! Please add yourself to the contributors page :D

app/mailers/notification_mailer.rb
chrisrp: I thought the same.. once we have the entire class covered I will try to improve this.

julianguyen: That would be fantastic! Thanks Chris :)

PR #283, opened 2016-10-01
Create grid to display items on index pages

### LINKED ISSUE #278: Using grid layout on Categories/Moods/Medications index pages
Each item (Category/Mood/Medication) on the index page takes up the full width of the screen, which is unnecessary given amount of content in the div. Let's use flex boxes to create a grid to display these items! On mobile web view, each item should take up the full width.

screen shot 2016-09-25 at 3 59 53 pm

Wowu:
- Change posts per page ( `$per_page`) global variable to 6
- Remove unnecessary "no_margin_bottom" class additions

Fixes #278

### DIFF
--- a/app/assets/stylesheets/application/pages.scss
+++ b/app/assets/stylesheets/application/pages.scss
@@ -15,6 +15,12 @@
 	}
 }
 
+#pages .page {
+	display: flex;
+	flex-wrap: wrap;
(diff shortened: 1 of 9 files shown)


julianguyen:
The PR looks good to me @Wowu, can you also add screenshots (just so that I confirm that's it's good to go)? Thanks!

PR #284, opened 2016-10-01
Form field to add refill onto Google Calendar

cartothemax:
This pull request will add a form field for users to switch on weather they want to add medication refills to Google Calendar (for those that used omniauth to join).

Originally users that used omniauth to sign up would automatically have medication refills added their Google Calendar. This now adds them only if the user has checked the "Add to Google Calendar" checkbox.

This is my first pull request to the if-me project, so, let me know if I missed anything when doing this work. Thanks!


julianguyen:
LGTM! Thanks @cartothemax!

PR #291, opened 2016-10-02
add support for i18n on the javascript side

sienatime:
Hopefully this will work! I added the middleware configuration so it should just automatically work in all environments without having to run the rake task to generate the translations.js file (`rake i18n:js:export`).


sienatime:
plz ignore code climate warnings for library code 🙏

julianguyen:
Yesss! This is great thank you @sienatime!

PR #292, opened 2016-10-02
Make i18n text in mailer HTML safe

### LINKED ISSUE #286: Raw html is showing in the email footer
screen shot 2016-10-01 at 2 27 34 pm

This might have been impacted by some of the work we are doing for internationalization. These should be real HTML links! The easiest way to see this in development is to comment on something (maybe private comment/shared with only 1 ally?).

leereilly:
A quick hack to fix #286 (raw HTML in the email footer) :metal: 

screen shot 2016-10-01 at 8 10 34 pm


sienatime:
I'd call that more of a fix than a hack, thanks!

Issue #288, opened 2016-10-01
Be able to auto select checkbox of newly created Category/Mood/Strategy in a Moment/Strategy


julianguyen:
Current behaviour:
1) Create a Mood through new Moments page:
![screen shot 2016-10-01 at 3 41 03 pm](https://cloud.githubusercontent.com/assets/3010728/19017439/ff70003c-87ed-11e6-92b7-25f00d1f9dc8.png)

2) Does not get auto-checked once created.
![screen shot 2016-10-01 at 3 41 13 pm](https://cloud.githubusercontent.com/assets/3010728/19017438/ff6f11b8-87ed-11e6-94ef-e24f0ee1ff58.png)

Expected behaviour:
1) Create a Mood through new Moments page
2) Does get auto-checked once created

We want this to work for the following:
- New Category in Moments
- New Mood in Moments
- New Strategy in Moments
- New Category in Strategies
- New Mood in Strategies

yigitozkavci:
I'm working on this! :)

PR #293, opened 2016-10-02
#288 - Check checkboxes of quick-added resources.

yigitozkavci:
After a successful `ajax` response, after adding `label` and `checkboxes` to `DOM`, we check them.

### DIFF
--- a/app/assets/javascripts/notifications.js
+++ b/app/assets/javascripts/notifications.js
@@ -179,6 +179,10 @@ function quickCreate(form, data_type) {
         $('#strategies_list').prepend(json.checkbox);
       }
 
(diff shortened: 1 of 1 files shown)


yigitozkavci:
Yes, I am interested, I will work on unit tests of `notifications.js` and refer to the issue #180. Thanks :)

julianguyen:
Thanks @yigitozkavci for helping out with this too! Please email us at join.ifme@gmail.com to be added to our Slack :P

yigitozkavci:
I will, one more thing: since is's a static data, keeping contributors list in controllers is a very bad practice imo, do you mind if I carry them to a `YAML` file?

PR #298, opened 2016-10-02
Fix positioning of labels in the tiles on the app index page

### LINKED ISSUE #296: Fix positioning of the Moment/Strategy label in the tiles on the app index page
screen shot 2016-10-02 at 12 27 57 pm

We want the Strategy label in this example to be fixed at the bottom of the div it's contained in - so that it lines up with the Moment label in the tile beside it.

Wowu:
Fix #296.


julianguyen:
LGTM!

Issue #131, opened 2016-05-16
Add email reminders for upcoming meetings


julianguyen:
Send Meeting members email reminders for upcoming meetings within 24 hours of the event.

tcdowney:
Hi @julianguyen, is anyone working on this issue?

julianguyen:
Nope! Feel free to work on this! Also email us at join.ifme@gmail.com to be
added to our Slack! Don't forget to add yourself to the Contributors page
too when you're ready.

tcdowney:
Great!  Thanks for getting back with me so quickly.. now time to get this dev environment set up! :full_moon_with_face:

PR #305, opened 2016-10-05
Email reminders to meeting members [#131]

tcdowney:
This adds a send_meeting_reminders task to the scheduler to send an email reminder to all meeting members a day before the meeting occurs (see issue #131).

**Notes:**
Since the `Meeting` models use strings for their `date` and `time` fields it makes it a bit difficult to query.  We should probably have an enhancement later to migrate them over to using a time field.

Given that, the way this is coded if you configure the scheduler on Heroku to run the `send_meeting_reminders` task daily at some time (midnight?) it will send an email to everyone who is attending a meeting the following day 24 hours+ in advance.

If you would like for the emails to be sent closer to 24 hours in advance in regards to the meeting's `time` field it gets a bit trickier since it's a string column and time zones become a bigger factor.

Regardless, I think this code is a decent MVP and can be expanded upon later if desired.  :)

Let me know what you would like changed and please give suggestions for improvement. 👍 

Thanks,
Tim

### DIFF
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -17,6 +17,14 @@ class NotificationMailer < ApplicationMailer
          subject: "Your refill for #{@medication.name} is coming up soon!")
   end
 
+  def meeting_reminder(meeting: , member: )
+    @meeting = meeting
+    @member = member
+
+    mail(to: @member.email,
+         subject: "Your meeting \"#{@meeting.name}\" is tomorrow at #{@meeting.time}!")
+  end
+
   def notification_email(recipientid, data)
     @data = JSON.parse(data)
     @recipient = User.where(id: recipientid).first
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -1,7 +1,7 @@
 class NotificationMailer < ApplicationMailer
-   default from: ENV["DEFAULT_FROM_ADDRESS"]
+  default from: ENV['DEFAULT_FROM_ADDRESS']
 
-   ALLY_NOTIFY_TYPES = ["new_ally_request", "accepted_ally_request"]
+  ALLY_NOTIFY_TYPES = %w(new_ally_request accepted_ally_request).freeze
 
   def take_medication(reminder)
     @medication = reminder.medication
@@ -17,12 +17,17 @@ class NotificationMailer < ApplicationMailer
          subject: "Your refill for #{@medication.name} is coming up soon!")
   end
 
-  def meeting_reminder(meeting: , member: )
+  def meeting_reminder(meeting, member)
     @meeting = meeting
     @member = member
 
-    mail(to: @member.email,
-         subject: "Your meeting \"#{@meeting.name}\" is tomorrow at #{@meeting.time}!")
+    subject = I18n.t(
+      'meetings.reminder_mailer.subject',
+      meeting_name: @meeting.name,
+      time: @meeting.time
+    )
+
+    mail(to: @member.email, subject: subject)
   end
 
   def notification_email(recipientid, data)
--- a/app/views/notification_mailer/meeting_reminder.html.erb
+++ b/app/views/notification_mailer/meeting_reminder.html.erb
@@ -1,7 +1,6 @@
 <p>
-  Hi <%= @member.name %>,
+  <%= t('meetings.reminder_mailer.salutation', name: @member.name) %>
 </p>
 <p>
-Here's a friendly reminder that your meeting tomorrow, "<%= @meeting.name%>", will be taking place at <%= @meeting.time %> at 
-<%= @meeting.location %>!
+  <%= t('meetings.reminder_mailer.body', meeting_name: @meeting.name, time: @meeting.time, location: @meeting.location) %>
 </p>
--- a/app/views/notification_mailer/meeting_reminder.html.erb
+++ b/app/views/notification_mailer/meeting_reminder.html.erb
@@ -1,0 +1,7 @@
+<p>
+  Hi <%= @member.name %>,
+</p>
+<p>
+Here's a friendly reminder that your meeting tomorrow, "<%= @meeting.name%>", will be taking place at <%= @meeting.time %> at 
+<%= @meeting.location %>!
+</p>
--- a/app/views/notification_mailer/meeting_reminder.text.erb
+++ b/app/views/notification_mailer/meeting_reminder.text.erb
@@ -1,4 +1,3 @@
-Hi <%= @member.name %>,
+<%= t('meetings.reminder_mailer.salutation', name: @member.name) %>
 
-Here's a friendly reminder that your meeting tomorrow, "<%= @meeting.name%>", will be taking place at <%= @meeting.time %> at 
-<%= @meeting.location %>!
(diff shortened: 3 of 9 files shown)


app/views/notification_mailer/meeting_reminder.html.erb
```diff
@@ -0,0 +1,7 @@
+<p>
+  Hi <%= @member.name %>,
+</p>
+<p>
+Here's a friendly reminder that your meeting tomorrow, "<%= @meeting.name%>", will be taking place at <%= @meeting.time %> at
```
tcdowney: Let me know what you actually want this to say.  I kinda just made this up and I'm no copywriter.. 😸

tcdowney:
Ew, I introduced a bunch of Code Climate issues.  Will get those fixed.

app/views/notification_mailer/meeting_reminder.html.erb
julianguyen: Can you put this string into `en.yml`? :)

app/mailers/notification_mailer.rb
```diff
@@ -17,6 +17,14 @@ def refill_medication(reminder)
          subject: "Your refill for #{@medication.name} is coming up soon!")
   end
 
+  def meeting_reminder(meeting: , member: )
```
julianguyen: Can we get rid of the whitespace after each argument?

app/mailers/notification_mailer.rb
```diff
@@ -17,6 +17,14 @@ def refill_medication(reminder)
...
+    @meeting = meeting
+    @member = member
+
+    mail(to: @member.email,
+         subject: "Your meeting \"#{@meeting.name}\" is tomorrow at #{@meeting.time}!")
```
julianguyen: Add to `en.yml`

app/views/notification_mailer/meeting_reminder.text.erb
```diff
@@ -0,0 +1,4 @@
+Hi <%= @member.name %>,
+
+Here's a friendly reminder that your meeting tomorrow, "<%= @meeting.name%>", will be taking place at <%= @meeting.time %> at
```
julianguyen: Put into `en.yml` too :P

julianguyen:
@tcdowney LGTM! Thanks for working on this.

> Given that, the way this is coded if you configure the scheduler on Heroku to run the send_meeting_reminders task daily at some time (midnight?) it will send an email to everyone who is attending a meeting the following day 24 hours+ in advance.

> If you would like for the emails to be sent closer to 24 hours in advance in regards to the meeting's time field it gets a bit trickier since it's a string column and time zones become a bigger factor.

I think for now that 24 hrs + in advanced ok until we refactor the date and time fields so that they aren't strings, but rather date/time objects. Can you create an issue for this?

spec/services/meeting_reminders_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      it 'does not send any notification emails' do
+        expect(NotificationMailer).to_not have_received(:meeting_reminder)
+      end
+    end
+
+    context 'when there is a meeting occurring tomorrow' do
+      let(:date) { 1.days.from_now.strftime('%m/%d/%Y') }
+
+      before do
+        expect(mail).to receive(:deliver_now).twice
```
julianguyen: Not 100% sure why this would be received twice, could you clarify? (:

tcdowney: Yeah, it's so it tests that it was delivered to each member.  Would it be more clear if I made this line like this to set up the spy

```
allow(mail).to receive(:deliver_now)
```

and then in the actual it block do it like this?
```
expect(mail).to have_received(:deliver_now).twice

Comment (julianguyen) [inline spec/services/meeting_reminders_spec.rb]: Yeah that's a good idea!

File: app/views/notification_mailer/meeting_reminder.text.erb (hunk 4 above)
Comment (julianguyen) [inline app/views/notification_mailer/meeting_reminder.text.erb]: Good question. @sienatime is working on the i18n project now. I'm thinking we have a language picker that saves the users preferences to local storage? Ideas/thoughts?

Comment (tcdowney) [inline app/views/notification_mailer/meeting_reminder.text.erb]: Local storage sounds like it would work for most cases, but I'm not sure if it would work for reminder-type emails since they are triggered by the scheduler and there is not actually a user session in context.

I think for these types of emails you may need to store the user's preferred locale somewhere persistent that the scheduler has access to -- maybe on the `User` model.  

Then the jobs could do something like this:

```
meetings_tomorrow.each do |meeting|
  meeting.members.each do |member|
    I18n.with_locale(member.locale) do
      NotificationMailer.meeting_reminder(meeting, member).deliver_now
    end
  end
end
```

I've never actually had to make internationalized mailers before, though, so take this with a grain of salt haha.

File: app/mailers/notification_mailer.rb [hunk 6]
```diff
@@ -1,7 +1,7 @@
 class NotificationMailer < ApplicationMailer
-   default from: ENV["DEFAULT_FROM_ADDRESS"]
+  default from: ENV['DEFAULT_FROM_ADDRESS']
 
-   ALLY_NOTIFY_TYPES = ["new_ally_request", "accepted_ally_request"]
+  ALLY_NOTIFY_TYPES = %w(new_ally_request accepted_ally_request).freeze
```
tcdowney: The indentation was off on these lines and it was causing Code Climate to report that the function I introduced was misaligned.  Since I touched these I had to make all of these changes to appease Rubocop.  😭

app/mailers/notification_mailer.rb
tcdowney: Updated: [commit](https://github.com/julianguyen/ifme/pull/305/commits/54ea544f19d4ea901dd13048f36f0efba77333d8#diff-0e5af34ef5e2f201a562b1ea575bad68R24)

app/views/notification_mailer/meeting_reminder.text.erb
tcdowney: i18ned: [commit](https://github.com/julianguyen/ifme/pull/305/commits/54ea544f19d4ea901dd13048f36f0efba77333d8#diff-2009ab0a038a430e0f8cea676f9ecfd4R1)

spec/services/meeting_reminders_spec.rb
tcdowney: Updated these tests: [commit](https://github.com/julianguyen/ifme/pull/305/commits/54ea544f19d4ea901dd13048f36f0efba77333d8#diff-d2ec198f10bf879943751efce0d8cf18R51)

app/views/notification_mailer/meeting_reminder.html.erb
tcdowney: Done :) [commit](https://github.com/julianguyen/ifme/pull/305/commits/54ea544f19d4ea901dd13048f36f0efba77333d8#diff-c9d60aff146070855874060d228ee26cR2)

julianguyen:
LGTM 💯 Fantastic work :)

julianguyen:
Feel free to merge when you're ready! Not sure if it will be reviewed by anyone else. 
Also add yourself to the Contributors page :)

tcdowney:
Thanks for reviewing @julianguyen, I added myself to the contributors list. :)

I actually don't have permission to merge this PR, though.

PR #309, opened 2016-10-06
Refactor mailer ally request and accepted request

chrisrp:
* Extracts ally notifications
* Adds I18n to notifications
* Adds UrlHelper module

The idea is to decompose the notification_mailer to smaller pieces. I am not sure if it is really necessary to group by types of notification with namespaces like `GroupType::EventType` but it is easy to change later if it does not make sense. 

Eg. `AllyNotifications::AcceptedAllyRequest` and `AllyNotifications::NewAllyRequest`.

### DIFF
--- a/app/mailers/ally_notifications/new_ally_request.rb
+++ b/app/mailers/ally_notifications/new_ally_request.rb
@@ -1,11 +1,13 @@
-module AllyNotifications
-  class NewAllyRequest
-    include Rails.application.routes.url_helpers
+# frozen_string_literal: true
 
-    def initialize(recipient, data, allies_url)
+module AllyNotifications
+  # Class helper to build new ally request message
+  class NewAllyRequest
+    include ::UrlHelper
+
+    def initialize(recipient, data)
       @recipient  = recipient
       @data       = data
-      @allies_url = allies_url
     end
 
     def to
@@ -17,8 +19,7 @@ module AllyNotifications
     end
 
     def message
-      I18n.t('mailers.new_ally_request.message', allies_url: @allies_url)
+      I18n.t('mailers.new_ally_request.message', allies_url: allies_url)
     end
   end
 end
-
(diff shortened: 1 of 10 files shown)


app/mailers/ally_notifications/new_ally_request.rb
```diff
@@ -0,0 +1,22 @@
...
+    def to
+      @recipient.email
+    end
+
+    def subject
+      I18n.t('mailers.new_ally_request.subject', ally_name: @data[:user])
+    end
+
+    def message
+      I18n.t('mailers.new_ally_request.message', allies_url: allies_url)
```
julianguyen: How is `allies_url` being passed here?

app/mailers/notification_mailer.rb
```diff
@@ -18,7 +18,7 @@ def refill_medication(reminder)
   end
 
   def notification_email(recipientid, data)
-    @data = JSON.parse(data)
+    @data = HashWithIndifferentAccess.new(JSON.parse(data))
     @recipient = User.where(id: recipientid).first
     subject = 'if me | '
```
julianguyen: Should be moved into i18n

config/application.rb
```diff
@@ -24,7 +24,7 @@ class Application < Rails::Application
 
     config.exceptions_app = self.routes
 
-    config.autoload_paths += Dir[Rails.root.join('app', 'uploaders')]
+    config.autoload_paths << Rails.root.join('lib')
```
julianguyen: Could you explain what this does? Not quite sure (:

julianguyen:
Awesome refactor @chrisrp! 💯 
Left a few comments, otherwise LGTM. Ooh also try to address these https://codeclimate.com/github/julianguyen/ifme/pull/309 the best you can.

app/mailers/ally_notifications/new_ally_request.rb
chrisrp: Because it includes the ` ::UrlHelper`

app/mailers/notification_mailer.rb
chrisrp: Sure, but the way I thought once all the notifications get extracted they will have the full subject.

julianguyen: Sounds good. We can do this for now then, and then in another PR do that.

Issue #308, opened 2016-10-05
Add tests for notifications.js


zachariast:
Add some tests for notifications.js

julianguyen:
Noticed that the clear button doesn't get removed when you initially open the notifications modal, but it does when you click the clear button.

![screen shot 2016-10-07 at 1 01 54 am](https://cloud.githubusercontent.com/assets/3010728/19182757/c770a250-8c29-11e6-9af1-34bec49597e5.png)

zachariast:
@julianguyen i'll check it now

zachariast:
@julianguyen i can't see the problem. When i open notification modal i see this.
![screen shot 2016-10-07 at 11 57 53 am 1](https://cloud.githubusercontent.com/assets/8640923/19184365/6a0202e2-8c85-11e6-9103-922f379b5f52.png)

julianguyen:
@zachariast 
Reproduction steps (locally and on prod):
- Log on
- Go to sub-navigation bar and click "Notifications"
- Modal opens, clear button still present

I think what we want to do is call `$('#notifications #clear_notifcations').hide();` after the AJAX request which fetches the notifications from the server (if it is empty).

tcdowney:
@zachariast @julianguyen 

I was working on adding a Jasmine spec to cover my change on #320 and I noticed a couple of issues.

There looks like there is a syntax error [here on line 43](https://github.com/julianguyen/ifme/blob/master/spec/javascripts/notificationsSpec.js#L43) that's preventing Jasmine from loading this file and running these specs.

Also, would you be opposed to introducing https://github.com/velesin/jasmine-jquery to the project so that we can use test fixtures?  That would also let us use things like the `toBeHidden()` matcher.

julianguyen:
Thanks for pointing this out @tcdowney. Can you fix that syntax error @zachariast please? :)

Not opposed to using jasmine-jquery. We definitely need test fixtures.

nchambe2:
Yes, I have the commit. I got a new computer, and haven't been able to get the project up and running locally again. I'm going to try again, and hope to have the commit pushed before the end of the week.

zachariast:
@julianguyen yes i'll do it. :)

PR #310, opened 2016-10-07
Tests for notification.js

zachariast:
fix for issue #308


zachariast:
It is the first time that i write tests... sooo help,comments,changes would be helpfull :) 👍

spec/javascripts/notificationsSpec.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it("show notifications modal", function() {
+    find('.notifications_button').click
+    expect($('#notifications').hasClass("display_block")).toBe(true);
+  });
+
+  it("close notifications modal", function() {
+    find('h1#close_notifications').click
+    expect($('a.notifications_button').hasClass("fade")).toBe(false);
```
julianguyen: I think testing for the absence of the modal itself would be a more direct way to test this behaviour

spec/javascripts/notificationsSpec.js
```diff
@@ -0,0 +1,56 @@
+describe("Notifications", function() {
+  beforeEach(function() {
+    var elements = [];
+    elements.push("<div id='notifications'></div>");
```
julianguyen: In general, I'm wondering if there's a better way mock DOM. Any ideas @nchambe2 @piecedigital @maestromusica?

julianguyen:
Awesome work @zachariast :D LGTM. When the comment for the close modal test has been addressed, feel free to merge!

spec/javascripts/notificationsSpec.js
nchambe2: @julianguyen Yes, we can use the jasmine-jquery library. It will allow us to load fixtures vs. building the dom in jquery. I spiked it out locally, and can push it this weekend.

julianguyen: That sounds perfect. Thanks so much @nchambe2 for working on this! 💯

zachariast:
@julianguyen I change the notifications test according to the comment, and i fix the behavior of the notification modal.

PR #312, opened 2016-10-08
Feature/Issue 311

brunoocasali:
Hello guys! I've fixed this: https://github.com/julianguyen/ifme/issues/311 item "3"

#### Description of changes and Questions:

- I've added a `.editorconfig` based on rubocop, It's fantastic to keep a consise best practices around of IDE's and Text Editors... [see more about editorconfig](http://editorconfig.org/)
- I've improved the `search_controller` too. With one just question... (the `data_types` is only hardcoded? or users pass it from one of pages?)
- I can improved other controllers like the before_filter in all pages, if you want...

Thanks!


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
```
julianguyen: This is awesome, thanks for adding this 😃

app/controllers/search_controller.rb
```diff
@@ -1,59 +1,36 @@
...
+  end
 
-		if !email.nil?
-			email.strip!
-			@matching_users = @matching_users.where(email: email)
-		end
-		@matching_users = @matching_users.where.not(id: current_user.id).all
-	end
+  def posts
+    data_type = params[:search][:data_type]
```
julianguyen: data_typr isn't hardcoded anywhere but maybe it should be since we make this string comparison in other places in the app

app/controllers/search_controller.rb
```diff
@@ -1,59 +1,36 @@
 class SearchController < ApplicationController
-	before_filter :if_not_signed_in
+  before_filter :if_not_signed_in
```
julianguyen: When you say improving before_filter in other controllers, do you mean this kind of change?

brunoocasali: No, I saw about the last method inside each controller, they are repeating the code. I can declare in just one place, then reuse it!

julianguyen: We should definitely reuse this code instead of redeclaring it in each controller. Thanks for bringing this up!

app/controllers/search_controller.rb
brunoocasali: I'm talking about this: https://github.com/julianguyen/ifme/blob/master/app/views/categories/index.html.erb#L10

Did you understand?!

julianguyen: Yeah so you're saying should make those strings some kind of global static variable? Correct me if I misunderstood.

app/controllers/search_controller.rb
brunoocasali: Wow, I'll generate an issue and work on it!

app/controllers/search_controller.rb
brunoocasali: Actually what I referred to was the following ...
I created an array of all the values that were pre-stated before, it is valid if it was me past is contained in this array. So I asked him if this information was passed through user input, if it were not I could remove that condition and make peace with the rubocop!

.editorconfig
brunoocasali: :blush:

PR #314, opened 2016-10-09
Feature/Issue 313

brunoocasali:
Hello guys! I've fixed this: 

- Moved `:if_not_signed_in` method to public version on `application_controller.rb`
- Remove the unnecessary tab after `private` keyword (rubocop things).


brunoocasali:
My PR will not pass, because I've changed application_controller and others, that for codeclimate have  many problems.

PR #316, opened 2016-10-09
Secure notifications controller

the-bass:
I noticed that, right now, basically every user can delete any notification, so I've added an additional `before_action` to allow users to delete their own notifications only.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,6 +1,7 @@
 class NotificationsController < ApplicationController
   before_filter :if_not_signed_in
   before_action :set_notification, only: [:destroy]
+  before_action :ensure_user_owns_notification!, only: [:destroy]
 
   # DELETE /notifications/1
   # DELETE /notifications/1.json
@@ -64,6 +65,14 @@ class NotificationsController < ApplicationController
           format.html { redirect_to new_user_session_path }
           format.json { head :no_content }
         end
+      end
+    end
+
+    # Stops here if the signed in user is not the owner of
+    # the notification set by :set_notification.
+    def ensure_user_owns_notification!
+      unless @notification.user == current_user
+        render :nothing => true
       end
     end
 end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -8,13 +8,13 @@ class NotificationsController < ApplicationController
   def destroy
     @notification.destroy
     respond_to do |format|
-        format.html { redirect_to :back }
+      format.html { redirect_to :back }
       format.json { head :no_content }
     end
   end
 
   def clear
-    Notification.where(userid: current_user.id).destroy_all if !Notification.where(userid: current_user.id).nil?
+    Notification.where(userid: current_user.id).destroy_all
     render :nothing => true
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -1,19 +1,20 @@
 class NotificationsController < ApplicationController
   before_filter :if_not_signed_in
   before_action :set_notification, only: [:destroy]
+  before_action :ensure_user_owns_notification!, only: [:destroy]
```
julianguyen: 💯

app/controllers/notifications_controller.rb
```diff
@@ -66,4 +67,12 @@ def if_not_signed_in
         end
       end
     end
+
+    # Stops here if the signed in user is not the owner of
+    # the notification set by :set_notification.
+    def ensure_user_owns_notification!
+      unless @notification.user == current_user
```
tcdowney: Personally, there seems to be a lot of indirection with the assignment of `@notification`.  It's being assigned in `set_notification`, then validated here, and finally actually used in only one place: the destroy action.  It's ok right now because this controller is still pretty simple, but I've seen as Rails apps grow, this type of juggling of instance variables between the before actions and the actual actions can get confusing.

What if we were to combine this new before action with the existing `set_notification` one so that the assignment and validation can happen in a single place?  Just my two cents! 👍

esBeee: @tcdowney @julianguyen Well, I didn't want to break the existing pattern, so I thought, it would be better not to touch `set_notification`. And, as I look at it again, I guess it might be worth thinking about a "global solution" (by that I mean a pattern to be followed in all controllers) to check whether the user is owner of the object, since we can find this problem in other controllers, too (take medications#destroy and medications#update as an example). I'm pretty new to contributing to open source, but maybe it's best if I open another issue about that!? Then you could merge your work in [#319](https://github.com/julianguyen/ifme/pull/319) instead of mine here, since you not only addressed this problem but also wrote tests for the other actions, and then think about away to implement/refactor "permission checking" in all controllers in a consistent way.

chrisrp: I'm also pretty new to open source and I would like to give some cents too LOL.

1 - Since both callbacks are: `only: [:destroy]` does it make sense to have both of them? I think retrieving the record + checking the user in the destroy action is not that bad. 
2 - I know it is not related to this pull request but this endpoint should be only for logged users. We could get rid of the `def signed_in` method and put the current_user in a base controller.

Then the action destroy could be something like:

`current_user.notifications.where(id: params[:id]) `
or
`Notifications.where(id: params[:id], user: current_user)`

IMHO -> For now I would do the whole thing in the action (probably using some private helper methods), but without callbacks.

julianguyen: I like your suggestion @tcdowney 💯 . Combining the new before action with the existing set_notification will save us from problems in the future. 

@esBeee Definitely open an issue for a global method to check whether a user is an owner of the object. Great suggestion!

julianguyen: @chrisrp So that signed_in method exists for `notifications.js` (https://github.com/julianguyen/ifme/blob/bd3a6ab6da7a7c1b281bff20c13d0970e8c54362/app/assets/javascripts/notifications.js#L197).

julianguyen:
Hey what's the status on this @esBeee?

timofonic:
@julianguyen @esBeee @tcdowney @chrisrp What's the current status of this pull request?

PR #325, opened 2016-10-11
Feature/AutoDeployment

brunoocasali:
Hello @julianguyen please see this pull request!

I need your help to finish them, you need to provide a secure key: 

```rb
gem install travis # if you not have it
```

Then you need to run: `travis encrypt $(heroku auth:token)` logged with your heroku user (owner of ifme application). So it just need to copy secure key and past at deploy step `"secure"`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,19 @@
 language: ruby
 rvm:
-- 2.3.1
+  - 2.3.1
 addons:
   postgresql: 9.4
   firefox: 47.0.1
 services:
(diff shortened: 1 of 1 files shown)


julianguyen:
LGTM @brunoocasali. Thanks for taking this on. Would you be interested in helping out with setting up continuous deployment? (:

brunoocasali:
@julianguyen of course! I just need to be invited on heroku and it should be done! :+1:

Issue #320, opened 2016-10-09
Use HTTP DELETE instead of GET for destructive NotificationsController#clear action


tcdowney:
Notifications for a user can currently be cleared by visiting the following route:

```
clear_notifications GET      /notifications/clear(.:format)               notifications#clear
```

Since the action is destroying all of a user's notifications, it should be an HTTP `DELETE` request instead of a `GET` since it is neither safe nor idempotent. ([reference](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html))

Moving away from `GET` also has the added benefit of letting us use Rail's CSRF security token in these requests. 🎃  🌝

tcdowney:
I plan on finishing this up after https://github.com/julianguyen/ifme/pull/319 is merged so that there  are some tests covering this functionality before I change it.

julianguyen:
I 💯  support this refactor 👍

PR #327, opened 2016-10-11
Use DELETE instead of GET for Notifications#clear [#320]

tcdowney:
Copied this from issue #320 for context:
> Notifications for a user can currently be cleared by visiting the following route:
> ```
clear_notifications GET      /notifications/clear(.:format)               notifications#clear
```

> Since the action is destroying all of a user's notifications, it should be an HTTP `DELETE` request instead of a `GET` since it is neither safe nor idempotent. ([reference](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html))

> Moving away from `GET` also has the added benefit of letting us use Rail's CSRF security token in these requests. 🎃  🌝

### DIFF
--- a/app/assets/javascripts/notifications.js
+++ b/app/assets/javascripts/notifications.js
@@ -239,9 +239,12 @@ var onReadyNotifications = function() {
     showNotificationsNone();
     changeTitle(0);
 
-    $.ajax("/notifications/clear");
-    $('#notifications #clear_notifcations').hide();
+    $.ajax({
+      url: '/notifications/clear',
+      type: 'DELETE'
+    });
 
+    $('#notifications #clear_notifcations').hide();
   });
 
   /* Quick Moment */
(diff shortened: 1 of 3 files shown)


tcdowney:
I was writing some Jasmine specs for this change, but I noticed an issue related to the existing specs.  See my comment [here](https://github.com/julianguyen/ifme/issues/308#issuecomment-252804757).

I can either add the test later once that file is fixed or just go ahead and fix it if you'd like... just don't want to step on anyones toes!!

julianguyen:
LGTM. Think it's ready to merge too. If you want to fix that syntax error here and merge - feel free to do that. Thanks @tcdowney!

tcdowney:
Thanks @julianguyen, I'll go ahead and merge this.  It looks like @zachariast is going to fix the Jasmine and I don't want to interfere. :) https://github.com/julianguyen/ifme/issues/308#issuecomment-253036131

PR #335, opened 2016-10-12
Fix similar code issue 

andela-oesho:
Things changed with this PR 

Abstracted similar methods occurring in controllers to a concern and removed all duplicate code from the following controllers …

1. categories_controller
2. memories_controller
3. medications_controller
4. moods_controller
5. strategies_controller

Looks Good to me


app/controllers/concerns/page_tool_tip.rb
```diff
@@ -0,0 +1,22 @@
+module PageToolTip
+  def set(collection, model_name)
```
julianguyen: Love this refactor 💯 
Two things:
- Should we rename PageToolTip to CollectionPageSetup, since it's setting up more than just the tool tip?
- Instead of passing in model_name should we pass the model itself?

app/controllers/concerns/page_tool_tip.rb
```diff
@@ -0,0 +1,22 @@
+module PageToolTip
```
brunoocasali: Hi @andela-oesho thanks for your PR!

We have encountering many work related for tests inside the project can you provide tests to it?

brunoocasali: And you can take good thoughts from this [stackoverflow question](http://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4) about Concerns concept!

app/controllers/medications_controller.rb
```diff
@@ -1,20 +1,14 @@
...
 
 class MedicationsController < ApplicationController
+  include PageToolTip
   helper_method :print_reminders
   before_action :set_medication, only: [:show, :edit, :update, :destroy]
 
   # GET /medications
   # GET /medications.json
   def index
-    name = params[:search]
```
brunoocasali: I love this! (remove duplicated code)! Fantastic :DD

doc/contributors/contributors.json
```diff
@@ -170,5 +170,9 @@
   {
     "name":"Bruno Casali",
     "link":"https://brunoocasali.xyz"
+  },
+  {
+    "name":"Susan Esho",
+    "link":"https://github.com/andela-oesho"
   }
 ]
```
brunoocasali: Can you provide a new blank line here?! Git has annoying problems about this... :+1:

app/controllers/concerns/page_tool_tip.rb
andela-oesho: I agree with renaming it to CollectionPageSetup, I would make necessary adjustments.

app/controllers/concerns/page_tool_tip.rb
andela-oesho: Alright thanks noted, I will do a controller test.

doc/contributors/contributors.json
andela-oesho: Alright, noted.

julianguyen:
Don't forget to rebase :P

PR #336, opened 2016-10-13
Strategies Controller & Notification Specs

talum:
For https://github.com/julianguyen/ifme/issues/311

Here's a first pass at adding specs for the strategies controller and notification specs! Some of the cases were harder to cover.


julianguyen:
LGTM @talum! Excellent work 💯  Thanks for taking these specs on :D

PR #337, opened 2016-10-13
[#112] Medication form layout

kindlingscript:
For context, view [issue #112](https://github.com/julianguyen/ifme/issues/112).

The new layout:

### **Larger screen sizes**
screen shot 2016-10-12 at 9 41 33 pm
screen shot 2016-10-12 at 9 41 51 pm

### **iPhone 6 Plus (according to Chrome responsive tools)**
screen shot 2016-10-12 at 9 47 48 pm

### **Smallest on responsive Chrome, resizing window down in width**

- Also tested on Firefox and Safari to the best of my ability


app/assets/stylesheets/application/medications.scss
```diff
@@ -9,6 +9,25 @@
 	height: 150px;
 }
 
+#medication_content {
+  padding: 2.5em 10em;
```
julianguyen: Can you convert the "em" values over to "px", just to be consistent?

app/assets/stylesheets/base/_colors.scss
```diff
@@ -8,7 +8,7 @@ $header-text-color: #fff;
 $link-color: #02779E;
 $subtle-secondary-text-color: #b5839b;
 $overlay-color: rgba(225, 205, 215, 0.2);
-$field-color: rgba(0, 0, 0, 0.2);
+$field-color: rgba(109, 8, 57, 0.4);
```
julianguyen: Ooh changing the field colour 🐙  I like how it looks in this form, but I'm not sure about other forms on the app. Is there too much maroon? Thoughts @hellocam @tiffanysun?

julianguyen:
Thanks for working on this @kindlingscript! Awesome job 💯 
Once the comments are addressed, it should be ready to merge 👍

app/assets/stylesheets/application/medications.scss
kindlingscript: Hi Julia! I used different values for padding on this specific page as without them the larger-width page looks like this (unless you meant something else when saying "use same values for padding/margin," let me know if I read that wrong):

screen shot 2016-10-12 at 10 32 48 pm

Comparing the above screenshot with the one that includes more padding for this page, I agree with what @aayushpi said earlier this year about how utilizing the entire screen width for content display isn't always the best option ("IMO no, there’s a lot of research about fields that are longer than they typically need to be being left incomplete. Our data set is small, so the fields should be as such"). 

Whitespace can be a benefit to effectively utilizing the "real estate" of the page for optimal user experience (from what I've learned so far, at least; I welcome others' opinions, though!), especially on larger screens. I was basically aiming for better readability, to feel more like filling out a typical form than the original, and something that's not too crowded looking or too isolated. 

If it's an absolute must to keep to the same padding/margin on this page, I can work on creating a different design in Sketch and start over (or left-align everything? though that would probably leave _too_ much whitespace on the right side, it's always a balancing act with that... [thinking to myself here, but I experimented anyways and got the following:

screen shot 2016-10-12 at 11 05 30 pm
]). 

Oh, and I can definitely convert the ems to pxs for this project, of course. 👍 😀

julianguyen: Ah ok, let's not use the exact same values as margin/padding then 😆 
I actually like the form in the centre, let's try that out and see what users think.

Thanks for clarifying Rebecca :)

app/assets/stylesheets/base/_colors.scss
julianguyen: I think for now, let's keep in black. And then we can revisit the colours in a later story?

app/assets/stylesheets/application/medications.scss
kindlingscript: Sure thing! I look forward to hearing what the users think later. 😄

app/assets/stylesheets/base/_colors.scss
kindlingscript: All right, I'll change it back for now (and add in my blurb, almost forgot about that!). I'd love to discuss colors in a later story. 👍

Issue #338, opened 2016-10-13
Setting up automated deployment


julianguyen:
Branch: https://github.com/julianguyen/ifme/tree/devops/local-env
- [x] Add `dotenv` gem
- [x] Move `.yml` config files to `.env`
- [x] Update README.md.
- [x] Spread to other collaborators about it.
- [x] Setup CircleC.I.
- [x] Add Circle badges to project.
- [x] Add webhooks checks to pull requests.
- [x] Don't allow directly pushes to master.

brunoocasali:
@julianguyen what about this issue?

There are only two items to solve, I have no permission to avoid pushes on master, and you have the influence needed to tell others about it! ;)

PR #339, opened 2016-10-13
Setup new Development Workspace + Automated Deploy

brunoocasali:
Fixes the issue #338


app/controllers/allies_controller.rb
```diff
@@ -109,15 +109,4 @@ def remove
       format.json { head :no_content }
     end
   end
-
-  private
-
-  def if_not_signed_in
```
julianguyen: Why is this removed?

config/application.yml
```diff
@@ -8,13 +8,13 @@
 # makes 'Your_Gmail_Username' available as ENV["GMAIL_USERNAME"]
 
 #This is the only configuration value required to boot the application
-SECRET_KEY_BASE: "examplesecretkey123123"
+SECRET_KEY_BASE: "945a3eb9dfc2b5ffa394d3e084dc49e08be6a548fc926976bbc9d8a61adc917214e0c85b20288550cfaf0384dcdde3b07e91f851b1b91a7739f4d67a58b882ae"
```
julianguyen: Think we should revert this line

config/application.yml
```diff
@@ -8,13 +8,13 @@
...
 
 # Configuration values for sending email, set these to true, configure your default "from address",
 # and configure smtp.yml in order to send email
 SEND_EMAIL: "false"
 RAISE_DELIVERY_ERRORS: "false"
-DEFAULT_FROM_ADDRESS: "yourmail@gmail.com"
+DEFAULT_FROM_ADDRESS: "brunoocasali@gmail.com"
```
julianguyen: This should be reverted too

app/controllers/allies_controller.rb
brunoocasali: Check the [`application_controller.rb:46`](https://github.com/julianguyen/ifme/blob/master/app/controllers/application_controller.rb#L46) there is a duplicated method!

config/application.yml
brunoocasali: Not this line, the whole file! You changed the gitignore, and I've don't ran a git remove locally... Because this, this file are here!

config/application.yml
brunoocasali: The same here!

config/env/development.example.env
```diff
@@ -0,0 +1,28 @@
+# Application
+SECRET_KEY_BASE: ""
```
brunoocasali: Why on development env is SECRET_KEY_BASE? and on test is CONFIG_SECRET_KEY_BASE ?

brunoocasali: And they need to be:

SOME_KEY=value # without quotes!

config/env/production.env
```diff
@@ -0,0 +1,28 @@
+# Application
```
brunoocasali: This file is not necessary, because this keys are already on Heroku!

config/env/test.example.env
```diff
@@ -0,0 +1,30 @@
...
+CONFIG_SECRET_KEY_BASE: ""
+
+# Cloudinary
+CLOUDINARY_CLOUD_NAME=""
+CLOUDINARY_API_KEY=""
+CLOUDINARY_API_SECRET=""
+
+# Configuration values for sending email, set these to true, configure your default "from address",
+# and configure smtp.yml in order to send email
+SEND_EMAIL: "false"
```
brunoocasali: If you want to use a SEND_EMAIL key, you need to use like: `SEND_EMAIL=false` and declare ENV['SEND_EMAIL'] on destination! ;D

config/env/development.example.env
julianguyen: OMG good catch
Yeah should be SECRET_KEY_BASE

julianguyen: would putting `insert_value` instead of `""` be okay?

config/env/test.example.env
```diff
@@ -0,0 +1,30 @@
...
+RAISE_DELIVERY_ERRORS: "false"
```
brunoocasali: Check the comment above!

julianguyen:
@brunoocasali Do we want to remove the wercker stuff?

brunoocasali:
I don't know what stuff are you talking about! :D If you found something go ahead!

config/env/development.example.env
julianguyen: Does this need to be updated?

brunoocasali: I've updated :D

julianguyen:
Finally merging this! Thanks @brunoocasali for working on this 💯

brunoocasali:
UHUUUUULLLLLLLL \o/ you rocks @julianguyen :metal:

PR #340, opened 2016-10-13
"#180 | Nicolette | Add Jasmine-Jquery Library"

nchambe2:
Add Jasmine Jquery Library


julianguyen:
Is it better to import the library manually or use a gem?

nchambe2:
It would be better to use the gem. I will push up the change.

julianguyen:
Sounds good thanks @nchambe2! :D

julianguyen:
I'm going to merge this actually - doesn't look like there's an official gem or one that has been updated recently.

PR #344, opened 2016-10-21
fix duplicate code

andela-oesho:
Things changed with this PR

Abstracted similar methods occurring in controllers to a concern and removed all duplicate code from the following controllers …

categories_controller
memories_controller
medications_controller
moods_controller
strategies_controller

new changes for this new branch ..... 
renamed Module name to CollectionPageSetup
added test for CollectionPageSetup

Looks Good to me asides the wrecker


julianguyen:
Looks good to me @andela-oesho! Thanks so much for working on this ❤️
Sorry for the delay too!

PR #350, opened 2016-10-29
reduced comment to 1 line and removed unnecessary word

sergeimellow:
working on my tshirt


julianguyen:
@sergeimellow This not a necessary refactor, but it is better to a comment in one line. Can you somehow reduce the words in this sentence so that the Codeclimate test will pass too? :P

Issue #379, opened 2016-11-24
Address depreciation warnings when running rspec


julianguyen:
Run `rspec` and address deprecation warnings in the output.

Example output:
```
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then
use `register_compressor` or `register_transformer`.
https://github.com/rails/sprockets/blob/master/guides/extending_sprockets.md#supporting-all-versions-of-sprockets-in-processors
 (called from block (2 levels) in <class:Railtie> at /Users/julia/.rvm/gems/ruby-2.3.1@ifme/gems/sass-rails-5.0.4/lib/sass/rails/railtie.rb:57)
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then
use `register_compressor` or `register_transformer`.
https://github.com/rails/sprockets/blob/master/guides/extending_sprockets.md#supporting-all-versions-of-sprockets-in-processors
 (called from block (2 levels) in <class:Railtie> at /Users/julia/.rvm/gems/ruby-2.3.1@ifme/gems/sass-rails-5.0.4/lib/sass/rails/railtie.rb:58)
DEPRECATION WARNING:           [Devise] including `Devise::TestHelpers` is deprecated and will be removed from Devise.
          For controller tests, please include `Devise::Test::ControllerHelpers` instead.
 (called from <top (required)> at /Users/julia/ifme/spec/controllers/allies_controller_spec.rb:1)
DEPRECATION WARNING:           [Devise] including `Devise::TestHelpers` is deprecated and will be removed from Devise.
          For controller tests, please include `Devise::Test::ControllerHelpers` instead.
 (called from <top (required)> at /Users/julia/ifme/spec/controllers/application_controller_spec.rb:5)
DEPRECATION WARNING:           [Devise] including `Devise::TestHelpers` is deprecated and will be removed from Devise.
          For controller tests, please include `Devise::Test::ControllerHelpers` instead.
 (called from <top (required)> at /Users/julia/ifme/spec/controllers/errors_controller_spec.rb:1)
DEPRECATION WARNING:           [Devise] including `Devise::TestHelpers` is deprecated and will be removed from Devise.
          For controller tests, please include `Devise::Test::ControllerHelpers` instead.
 (called from <top (required)> at /Users/julia/ifme/spec/controllers/groups_controller_spec.rb:1)
[...]

PR #391, opened 2016-11-27
Susan/fix deprecation errors

andela-oesho:
fix deprecation errors for register_mime_types
#379 fix deprecation errors for old syntax by explicitly enabling `:should` in spec_helper


spec/spec_helper.rb
```diff
@@ -36,6 +36,9 @@
   config.include FactoryGirl::Syntax::Methods
   config.include Devise::Test::ControllerHelpers, :type => :controller
   config.include StubCurrentUserHelper
+  config.mock_with :rspec do |mock_config|
+    mock_config.syntax = [:expect, :should]
```
julianguyen: Looks good to me

.gitignore
```diff
@@ -26,6 +26,7 @@ database.yml
 
 # Ignore coverage results
 /coverage
+/config/application.yml
```
andela-oesho: alright i will update it, i actually deleted my application.yml but i will delete this line of ignoring it

spec/spec_helper.rb
andela-oesho: thanks

Issue #376, opened 2016-11-18
Generate slugs for individual pages for Moments/Strategies/Moods/Categories/Medications/Groups/Meetings


julianguyen:
Accessing an individual page for a Moment looks like this for example `moments/1` where 1 is the id. We want to replace this with `moments/name-of-moment-slug` so that it's more readable. Also make sure to handle duplicate slugs e.g. `moments/cool-sluggy` and `moments/cool-sluggy-2` etc.

- [x] Moments
- [x] Strategies
- [x] Moods
- [x] Categories
- [x] Medications
- [x] Groups
- [x] Meetings

Perhaps this gem is useful? https://github.com/norman/friendly_id

brunoocasali:
@julianguyen this gem is fantastic! I agree with this change!

PR #394, opened 2016-11-29
[#376] Generate slugs using friendly_id

julianguyen:
Using the [friendly_id](https://github.com/norman/friendly_id) gem to generate slugs using the name field in the models below.

* Moments
* Strategies
* Moods
* Categories
* Medications
* Groups
* Meetings


julianguyen:
Any ideas on how to fix `Unsafe model attribute in link_to href` CodeClimate issue @brunoocasali?

brunoocasali:
of course! @julianguyen I'll try something 👍

sienatime:
I think you can fix the href warning by calling `sanitize` on the `link_to`, like so

```the_link = sanitize link_to User.where(:id => @mood.userid).first.name, link_url```

You can read more about why here https://stackoverflow.com/questions/9634868/sanitizing-url-to-prevent-xss-in-rails

File: app/controllers/categories_controller.rb [hunk 1]
```diff
@@ -128,33 +128,37 @@ def quick_create
...
-        if Moment.where(id: moment).exists? && Moment.where(id: moment).first.category.include?(category.id) && Moment.where(id: moment).first.viewers.include?(current_user.id) && are_allies(moment.userid, current_user.id)
-          return true
-        end
-      end
+  def category_params
+    params.require(:category).permit(:name, :description, :userid)
+  end
 
-      return false
+  def is_viewer(moment, strategy, category)
```
sienatime: this seems like a good thing to pull out into its own class. there are a lot of other helper methods that seem to do a similar thing (in application, moments, and mood controllers) so it would be cool to have a Viewer class or something like that. also these variables are actually IDs so it would be great to be more explicit about that (e.g. rename to `moment_id`, etc)

app/controllers/categories_controller.rb
```diff
@@ -128,33 +128,37 @@ def quick_create
...
+    if strategy &&
+       Strategy.where(id: strategy).exists? &&
+       Strategy.where(id: strategy).first.viewers.include?(current_user.id)
+      return true
+    elsif moment && category &&
+          Moment.where(id: moment).exists? &&
+          Moment.where(id: moment).first.category.include?(category.id) &&
+          Moment.where(id: moment).first.viewers.include?(current_user.id) &&
+          are_allies(moment.userid, current_user.id)
```
sienatime: is `moment.userid` correct here? `moment` doesn't seem to be a `Moment`, it seems to be an ID that came in from the params...

app/controllers/categories_controller.rb
```diff
@@ -128,33 +128,37 @@ def quick_create
...
+      return true
     end
+    return false
+  end
```
sienatime: here's a suggested refactor for this method but if you end up with something different after putting it into its own class, that's fine too (i didn't actually run this so might not be 100% correct)
```
  def is_viewer(moment_id, strategy_id, category)
    if strategy = Strategy.find(strategy_id)
      strategy.viewers.include?(current_user.id)
    elsif category && moment = Moment.find(moment_id)
      moment.category.include?(category.id) &&
      moment.viewers.include?(current_user.id) &&
      are_allies(moment.userid, current_user.id)
    else
      false
    end
  end
```

app/models/medication.rb
```diff
@@ -37,4 +39,8 @@ class Medication < ActiveRecord::Base
   def active_reminders
     [refill_reminder, take_medication_reminder].select(&:active?)
   end
+
+  def should_generate_new_friendly_id?
```
sienatime: you could put all of these `should_generate_new_friendly_id?` etc into a  FriendlyId concern if you want: http://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4

config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,88 @@
...
+  #
+  # This is significantly more convenient but may not be appropriate for
+  # all applications, so you must explicity opt-in to this behavior. You can
+  # always also configure it on a per-model basis if you prefer.
+  #
+  # Something else to consider is that using the :finders addon boosts
+  # performance because it will avoid Rails-internal code that makes runtime
+  # calls to `Module.extend`.
+  #
+  # config.use :finders
```
sienatime: i think it's a good idea to leave this commented out ;) based on how confusing I've found friendly id in other applications...

config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,88 @@
...
+  #
+  #  ### Controlling when slugs are generated
+  #
+  # As of FriendlyId 5.0, new slugs are generated only when the slug field is
+  # nil, but if you're using a column as your base method can change this
+  # behavior by overriding the `should_generate_new_friendly_id` method that
+  # FriendlyId adds to your model. The change below makes FriendlyId 5.0 behave
+  # more like 4.0.
+  #
+  # config.use Module.new {
```
sienatime: it seems like you could also just turn this config option on and use `name_changed?` instead of explicitly defining it in all of your models... unclear about the call to `super` but you could try it out

config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,88 @@
...
+  # calls to `Module.extend`.
+  #
+  # config.use :finders
+  #
+  # ## Slugs
+  #
+  # Most applications will use the :slugged module everywhere. If you wish
+  # to do so, uncomment the following line.
+  #
+  # config.use :slugged
```
sienatime: seems like you could enable this to get rid of the `use: :slugged` from all of your models

julianguyen:
@sienatime Thanks for the suggestion, thought it would work, but it didn't :/
I guess we could just ignore that warning?

app/controllers/categories_controller.rb
julianguyen: Ooh you're right!

app/controllers/categories_controller.rb
julianguyen: Renamed this to `can_view_category` and used `is_viewer` from ApplicationController for `strategy.viewers.include?` and `moment.viewers.include?`.

config/initializers/friendly_id.rb
julianguyen: Good catch!

app/helpers/reminder_helper.rb
```diff
@@ -2,7 +2,7 @@ module ReminderHelper
   def print_reminders(data)
     reminders = ''
 
-    if data.active_reminders.any?
+    if data.active_reminders && data.active_reminders.any?
```
julianguyen: Support old strategies that do not support notifications

app/models/medication.rb
```diff
@@ -35,6 +37,7 @@ class Medication < ActiveRecord::Base
   attr_accessor :add_to_google_cal
 
   def active_reminders
+    return unless refill_reminder && take_medication_reminder
```
julianguyen: Support old strategies that do not support reminders

app/models/strategy.rb
```diff
@@ -34,7 +36,7 @@ def array_data
   end
 
   def active_reminders
-    [perform_strategy_reminder].select(&:active?)
+    [perform_strategy_reminder].select(&:active?) if perform_strategy_reminder
```
julianguyen: Support old strategies that do not support reminders

julianguyen:
Thanks for reviewing @sienatime :) Can you re-review?

app/controllers/categories_controller.rb
```diff
@@ -128,33 +127,35 @@ def quick_create
...
-      rescue
-        if @category.blank?
-          respond_to do |format|
-            format.html { redirect_to categories_path }
-            format.json { head :no_content }
-          end
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_category
+    begin
```
brunoocasali: You can remove this `begin` keyword :)

PR #396, opened 2016-12-01
add test smtp from to example dev env

### LINKED ISSUE #220: App hangs between post and action mailer
If there is no internet connection, the app seems to 'hang' on the local install after post request is sent if action mailer isn't able to send. Unsure if there's a 'failsafe' action to catch the error - I'd suggest implementing if it isn't there.

BonMatts:
fixes #220 See comment on issue for details.


julianguyen:
@BonMatts Do we want to put this in test.example.env as well?

BonMatts:
added to test.env as well!

PR #399, opened 2016-12-10
I18n

BonMatts:
Add devise views to i18n

### DIFF
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -1,5 +1,5 @@
 <p>  <%= t('salutation', @email) %>! </p>
 
-<p><%= t('devise.confirmation.link_instructions') %>:</p>
+<p><%= t('devise.mailer.confirmation.link_instructions') %>:</p>
 
-<p><%= link_to t('devise.confirmation.text'), confirmation_url(@resource, confirmation_token: @token) %></p>
+<p><%= link_to t('devise.mailer.confirmation.link_text'), confirmation_url(@resource, confirmation_token: @token) %></p>
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -1,5 +1,5 @@
-<p>Welcome <%= @email %>!</p>
+<p>  <%= t('salutation', @email) %>! </p>
 
-<p>You can confirm your account email through the link below:</p>
+<p><%= t('devise.confirmation.link_instructions') %>:</p>
 
-<p><%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>
+<p><%= link_to t('devise.confirmation.text'), confirmation_url(@resource, confirmation_token: @token) %></p>
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -1,4 +1,4 @@
-<p>  <%= t('salutation', @email) %>! </p>
+<p><%= t('salutation', name: @email) %>!</p>
 
 <p><%= t('devise.mailer.confirmation.link_instructions') %>:</p>
 
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -1,11 +1,11 @@
 <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f| %>
 	<%= devise_error_messages! %>
 
-	<div class="field"><div class="label"><%= f.label "Name" %></div>
+	<div class="field"><div class="label"><%= f.label t('devise.registrations.labels.name') %></div>
 	<%= f.text_field :name %></div>
 
 	<div class="field"><div class="label"><%= f.label :location %></div>
-	<%= f.text_field :location, placeholder: "Optional" %></div>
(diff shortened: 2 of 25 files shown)


BonMatts:
JK LOOKS LIKE I BROKE A BUNCH OF TESTS

sienaigg:
Hi @BonMatts, thanks for tackling this! I'm curious if you were able to test the emails you modified? When I looked at this in the past, I had some trouble getting some of these emails to actually send in my local dev environment, so that had prevented me from moving forward with extreme confidence. It would be great to verify that those still look as expected, but if it proves to be too much trouble, I'm okay with merging this as-is. Thanks again!

config/locales/en.yml
```diff
@@ -39,6 +39,13 @@ en:
  salutation: "Hi %{name},"
  click_here: "click here"
 
+ common:
+   email: "Email"
+   location: "location"
```
sienaigg: Should this be capital-L"Location"?

config/locales/en.yml
```diff
@@ -57,17 +64,13 @@ en:
...
-   error_explanation: 'Please fill out the marked fields!'
    name_hint: 'Categorize recurring people, places, things, activities, and more'
   label:
    one: 'Category:'
    other: 'Categories:'
   index:
    title: 'Categories'
    subtitle: 'Life is messy, categorize the different people, places, and things.'
-   name: 'Name'
    description: 'Description'
```
sienaigg: can this Description be factored out to use the common one?

config/locales/en.yml
```diff
@@ -564,3 +551,74 @@ en:
...
+        profile_picture: "Profile picture"
+        remove_picture: "Remove picture"
+        password_for: "Password for <strong><%= t('app_name') %></strong> account"
+        update: "Update"
+      placeholders:
+        optional: "Optional"
+        about: "Introduce yourself! Why are you here and what are you looking to achieve? Please note this will be seen when users search for you"
+        required: "Required"
+      delete:
+        text: "Is <strong><%= t('app_name') %></strong> not working out for you?"
```
sienaigg: same comment as above. I think it's safe to just replace this with `if me`.

config/locales/en.yml
```diff
@@ -564,3 +551,74 @@ en:
...
+    sign_in: 'Sign in'
+  shared:
+    sign_in_google: 'Sign in with Google'
+    sign_up_google: 'Sign up with Google'
+    password: 'Forgot your password?'
+    confirmation: "Didn't receive confirmation instructions?"
+    unlock: "Didn't receive unlock instructions?"
+  unlock:
+    resend: 'Resend unlock instructions'
+
```
sienaigg: can we get rid of all of the extra whitespace here? just one newline should suffice :)

app/views/devise/registrations/edit.html.erb
```diff
@@ -1,19 +1,19 @@
...
+	<%= f.text_field :location, placeholder: t('devise.registrations.edit.placeholders.optional') %></div>
 
 	<div class="field">
 		<div class="label">
 			<%= f.label :email %>
 		</div>
 		<% if current_user.provider == "google_oauth2" %>
 			<div class="small_font small_margin_bottom">
-				<%= raw t('devise.registrations.google_auth1') %>
+				<%= raw  t('devise.registrations.google_auth1') %>
```
sienaigg: extra whitespace was added here

config/locales/en.yml
BonMatts: Yes - and I started another branch dedicated to cleaning up the en.yml file. :)

BonMatts:
Looking into testing mailers - will let you know what I find!

BonMatts:
With a change to devise config file, I was able to test the reset password mailer. It appears that we are not currently requiring users to confirm their emails, so there is no need to test that mailer.

BonMatts:
Okay, I think I addressed all of your comments @sienaigg

sienaigg:
@julianguyen does this change to config/initializers/devise.rb look okay to you? i'm not that familiar with the mailer setup

julianguyen:
@sienaigg @BonMatts The change to devise.rb looks good to me! 👍

app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -1,5 +1,5 @@
-<p>Welcome <%= @email %>!</p>
+<p>  <%= t('salutation', @email) %>! </p>
```
julianguyen: Should there be spaces after `<p>` and before `</p>`?

app/views/devise/registrations/new.html.erb
```diff
@@ -1,24 +1,24 @@
...
-    <div class="field"><div class="label"><%= f.label 'Name' %></div>
+    <div class="field"><div class="label"><%= f.label t('common.name') %></div>
     <%= f.text_field :name %></div>
 
-    <div class="field"><div class="label"><%= f.label 'A few details' %></div>
-    <%= f.text_field :location, class: 'spacing', placeholder: 'Location' %>
-    <%= f.email_field :email, placeholder: 'Email' %></div>
+    <div class="field"><div class="label"><%= f.label t('devise.registrations.new.label') %></div>
+    <%= f.text_field :location, class: 'spacing', placeholder: t('common.location') %>
+    <%= f.email_field :email, placeholder: t('email') %></div>
```
sienatime: I think this one should be `t('common.email')` since the top-level email is `join.ifme@gmail.com` :upside_down_face:

BonMatts: whoops! good catch!

PR #400, opened 2016-12-15
I18n cleanup

BonMatts:
Large refactoring of the en.yml file, removing repetitious keys, and re-categorizing others. perhaps your wheelhouse, @sienatime ?


app/controllers/application_controller.rb
```diff
@@ -384,14 +384,14 @@ def viewers_hover(data, link)
     viewers = ''
 
     if link
-      viewers += t('shared.viewers_hover.visible_to')
+      viewers += t('shared.viewers.visible_to')
```
sienatime: ahh I thought I had caught all of these, but we need to eliminate this kind of string concatenation of partial English phrases... (this one builds up `Visible only ` + `to you`). I'm not sure what brakeman is complaining about but refactoring this to not concatenate strings would probably help.

sienatime:
I will look at this more thoroughly after your other PR is merged!

app/controllers/application_controller.rb
BonMatts: Cool, I'll look into refactoring it.

app/views/moments/_form.html.erb
```diff
@@ -37,8 +37,8 @@
           </span>
           <% end %>
         <% end %>
-        </div>
-        <span id='category_quick_button' class='link_style small_margin_top'><%= t('moments.form.add_category') %></span>
+                                                                                 </div>
```
sienatime: this whitespace got a little wacky here!

config/locales/en.yml
```diff
@@ -2,77 +2,97 @@ en:
...
- stories: 'Stories'
- welcome: 'Welcome'
  created: '<strong>Created:</strong> %{created_at}'
  edited: '<strong>Created:</strong> %{created_at} <em>(edited)</em>'
  edited_updated_at: '<strong>Created:</strong> %{created_at} <em>(edited %{updated_at})</em>'
- daily_reminder: 'Daily reminder email'
  salutation: "Hi %{name},"
  click_here: "click here"
 
+ support:
```
sienatime: you shouldn't have to add this yourself, i've been using `to_sentence` all over the place as part of the i18n refactor and these are the default values included in active support. was there another reason you needed to add these?

config/locales/en.yml
```diff
@@ -82,64 +102,43 @@ en:
...
-   cannot_delete: 'Please note that you cannot delete groups.'
-   error_save: "We can't save this group because: "
-   error_edit_permission: 'You must be a leader of a group in order to edit it'
+    leaders: 'Leaders'
+    only_leader: "You are the only leader, therefore you can't leave."
+    cannot_delete: 'Please note that you cannot delete groups.'
+    error_save: "We can't save this group because: "
+    error_edit_permission: 'You must be a leader of a group in order to edit it'
   index:
    title: 'Groups'
```
sienatime: would you like to move this to `groups.plural` to keep things more consistent?

app/controllers/application_controller.rb
```diff
@@ -381,24 +381,22 @@ def get_viewers_for(data, data_type)
...
     if data.blank? || data.length == 0
       if link
-        viewers += t('shared.viewers_hover.only_you').downcase
+        viewers = t('shared.viewers.you_link')
       else
-        viewers += t('shared.viewers_hover.only_you')
+        viewers = t('shared.viewers.you')
+      end
+    else
+      viewer_names = data.to_a.map { |user_id| User.find(user_id).name }.to_sentence()
```
sienatime: code climate seems to not like the use of parentheses on `to_sentence` here

app/controllers/application_controller.rb
```diff
@@ -381,24 +381,22 @@ def get_viewers_for(data, data_type)
...
+      if link
+        viewers = t('shared.viewers.many', viewers: viewer_names)
+      else
+        viewers = viewer_names
```
sienatime: here's what rubocop has to say about this conditional assignment jazz... I'll try to help refactor this method this weekend http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/ConditionalAssignment

BonMatts: I started working on a refactor in another branch, if you want to check it out - [controller_cleanup](https://github.com/julianguyen/ifme/tree/controller_cleanup

config/locales/en.yml
BonMatts: nope! just didn't catch that it defaults to English when i was reading up on it. will remove.

BonMatts:
okay, pretty sure I grabbed all the small changes you mentioned,  I  refactored Viewers_hover, and created moments_helper, pulling some logic out of the view. Let me know what you think @sienatime :)

app/models/moment.rb
```diff
@@ -42,4 +42,8 @@ def array_data
     end
   end
 
+  def strategy
```
sienatime: what is this used for? I couldn't find it

app/helpers/viewers_helper.rb
```diff
@@ -1,13 +1,54 @@
...
+
+  def viewers_hover(data, link)
+    result = { class: 'yes_title',
+               title: get_viewer_list(data, link).html_safe }
+    if link
+      content_tag(:span, result) do
+        link_to link.name, (link.class.link + link.id.to_s)
+      end
+    else
+      result[:class] += ' small_margin_right'
```
sienatime: this refactor is a really great improvement! I think it would be even better if we could stop generating HTML in the helpers and just have it return data values (e.g. what goes in the `title` attribute) but we can do that later.

app/helpers/viewers_helper.rb
```diff
@@ -1,13 +1,54 @@
...
+      content_tag(:span, tag(:i, class: 'fa fa-lock'), result)
+    end
+  end
+
+  def get_viewers_for(data, data_type)
```
sienatime: another thing i'm not crazy about is the vagueness of the variable name `data` (i know that was already here) and passing around a data type when we could just call `.class` on whatever the model is... but also, not something we need to refactor right now.

app/models/category.rb
```diff
@@ -13,4 +13,8 @@
 class Category < ActiveRecord::Base
   validates_length_of :description, :maximum => 2000
   validates_presence_of :userid, :name
+
+  def self.link
```
sienatime: this feels like a strange thing to put at the model layer..? but it's okay for now

app/models/moment.rb
BonMatts: It's a bit obscure, but it allowed for a bit of trickery that reduced 
`def get_viewers_for(data, data_type)
    result = Array.new

    if data && (data_type == 'category' || data_type == 'mood' || data_type == 'strategy')
      Moment.where(userid: data.userid).all.order("created_at DESC").each do |moment|
        if data_type == 'category'
          item = moment.category
        elsif data_type == 'mood'
          item = moment.mood
        else
          item = moment.strategies
        end

        if item.include?(data.id)
          result += moment.viewers
        end
      end

      if (data_type == 'category')
        Strategy.where(userid: data.userid).all.order("created_at DESC").each do |strategy|
          if strategy.category.include?(data.id)
            result += strategy.viewers
          end
        end
      end
    end`

to 

`
  private def get_viewers(data, data_type, obj)
    objs = obj.where(userid: data.userid).all.order('created_at DESC')
    objs.each do |ob|
      item = ob.send(data_type)
      return ob.viewers if item.include?(data.id)
    end
  end`

BonMatts: because all other classes, except for strategy has a method that is it's name in singular, I added this alias method to strategy.

BonMatts:
Squashed down those commits :)

app/models/category.rb
sienatime: I think where possible we should be trying to use Rails's path helper methods instead of concatenating URIs all over the place... I don't think a model should know what (partial) path something else might need to display it (it's a separation of concerns thing). But this is several layers away from where we're at right now.

sienatime:
Whooooo boy okay so I merged this without actually checking out the branch and loading it in my local environment, which was a bad idea, since the homepage got messed up in the process

image

I did a git bisect and tracked it down to SHA 7fa0eff, `refactor viewers_hover`

Unfortunately since I already merged it, I had to open another PR to revert the merge for now: https://github.com/julianguyen/ifme/pull/402 which I need someone else to approve.

PR #402, opened 2016-12-18
Revert "I18n cleanup"

sienatime:
Reverts julianguyen/ifme#400

See https://github.com/julianguyen/ifme/pull/400#issuecomment-267796631


sienatime:
no need for this anymore!

PR #404, opened 2016-12-18
fix some i18n keys in registrations/new

sienatime:


sienatime:
reopened as https://github.com/julianguyen/ifme/pull/405

PR #407, opened 2016-12-24
Siena/enable es i18n

sienatime:
This PR accomplishes the following (can be split up if needed):

* Update the devise English translations according to the official gem (https://github.com/plataformatec/devise/wiki/I18n)
* Add Spanish translation files for devise and kaminari gems
* Add the rails-i18n gem (https://github.com/svenfuchs/rails-i18n), which enables internationalization of many Rails components like error messages, dates and times, etc. Locales are set to be English and Spanish (will also help facilitate the locale picker when we have one)
* Add placeholder es.yml file (to be replaced with real translations when we have them)
* Remove the fork of the local_time gem in favor of using rails-i18n's `time_ago_in_words`, which supports localization (local_time is only available in English)
* Move `created_or_edited` to new `TimeAgo` class to take stuff out of application controller wooooooo

screen shot 2016-12-23 at 6 06 15 pm


sienatime:
@julianguyen would love your thoughts when you get a chance :)

spec/services/time_ago_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        expect(subject).to eq("<strong>Created:</strong> #{TimeAgo.formatted_ago(new_moment.created_at)}")
+      end
+    end
+
+    context "creating and editing happened on the same day (but not the exact same time)" do
+      before do
+        new_moment.update(category: Array.new(1, new_category.id))
+      end
+
+      it "returns created_at with (edited) if " do
```
julianguyen: Should there be something after "if" ?

spec/services/time_ago_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    context "editing happened on a different day than it was created" do
+      let(:new_moment) { create(:moment, userid: new_user1.id, created_at: '2014-01-01 00:00:00') }
+
+      before do
+        new_moment.update(category: Array.new(1, new_category.id))
+      end
+
+      it "returns created_at with updated_at if " do
```
julianguyen: Should there be something after "if" ?

spec/services/time_ago_spec.rb
```diff
@@ -0,0 +1,86 @@
+describe TimeAgo do
```
julianguyen: Should we also test the case when the output has a time (and not date)?

sienatime: what is that case? this is the current implementation
```
if time_with_zone < 1.week.ago
        I18n.l(time_with_zone.to_date, format: :long)
      else
        I18n.t('common.time_ago', date: time_ago_in_words(time_with_zone))
      end
```
so if it is less than a week ago, you'll get the relative time, otherwise you'll get just the date (no time). In the previous implementation with the fork, you could get something like "Sunday at 3:54 pm" but the Rails i18n gem doesn't have that same kind of day-of-the-week formatting (we could write it if it's really important). are there places where I've replaced the implementation that are just time data? cuz then yeah we would need to change that. thanks!

spec/services/time_ago_spec.rb
sienatime: oops, refactored some specs and left some grammar behind

sienatime:
Thanks @julianguyen ! I replied to your comments with some questions.

Issue #416, opened 2017-01-08
Add ability to add premade content for Categories, Moods, and Strategies in Moments and Strategies editors


julianguyen:
Currently, the user can only add premade content for Categories, Moods, and Strategies on their respective index pages.

Example:
![image](https://cloud.githubusercontent.com/assets/3010728/21753468/f6aa7ba0-d5a2-11e6-93ef-893c82de0911.png)

We want to add premade content from the Moments and Strategies editor:
![image](https://cloud.githubusercontent.com/assets/3010728/21753466/dc62ad76-d5a2-11e6-9bc4-4732e9bd121d.png)

Either add this new button in the sidebar or in the modal after clicking "Add [...]"

We only want this button to show up there is no existing content.

### Cases:
Moments
* Categories
* Moods
* Strategies

Strategies
* Categories

andela-oesho:
@julianguyen for clarity the add modal button should only appear if moods/strategies/categories.blank?

julianguyen:
@andela-oesho Yep that is correct (for that user)!

PR #418, opened 2017-01-14
add premade content for moments and strategies

andela-oesho:
#416 LGTM i could proceed to add for medications if you give a go ahead
image


julianguyen:
Thanks for working on this @andela-oesho :)

I think we would want an "Add premade content" button for Categories, Moods, and Strategies separately. Also we would want a way for them to preview it. 

Possible behaviour:
1) Hit expand button in sidebar for each section
2) Button appears
3) Clicking the button opens a modal listing the premade content that will show up
4) User clicks another button confirm adding it

PR #439, opened 2017-05-09
UI toggle for i18n

julianguyen:
- [x] Fixed bugs with translations
- [x] Input fields with placeholders sized based on placeholder text length
- [x] Not signed in: language can be toggled on the same 
- [x] Not signed in: language can be toggled on different pages
- [x] Signed in: language can be toggled on the same 
- [x] Signed in: language can be toggled on different pages
- [x] Language toggled when not signed in persists when signed in
- [x] Language toggled when signed in persists when not signed in
- [x] Translations missing in Account page

## Mobile Web
image
image

## Desktop
image
image

### DIFF
--- a/app/assets/javascripts/placeholder_input_size.js
+++ b/app/assets/javascripts/placeholder_input_size.js
@@ -1,0 +1,12 @@
+function setInputSize(id) {
+  var length = $(id).attr("placeholder").length;
+  $(id).attr("size", length);
+}
+
+var onReadyPlaceholderInputSize = function() {
+  if ($("body").hasClass("allies index")) {
+    setInputSize("#search_email");
+  }
+};
+
+$(document).on("page:load ready", onReadyPlaceholderInputSize);
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -43,11 +43,10 @@ class PagesController < ApplicationController
   private
 
   def toggle_when_signed_in(locale)
-    user = User.find(current_user)
-    if !locale && user || locale == user.locale
-      render json: { signed_in_no_reload: user.locale }
+    if !locale && current_user || locale == current_user.locale
+      render json: { signed_in_no_reload: current_user.locale }
     else
-      user.update!(locale: locale)
+      current_user.update!(locale: locale)
       render json: { signed_in_reload: locale }
     end
   end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -44,7 +44,7 @@ class PagesController < ApplicationController
 
   def toggle_when_signed_in(locale)
     user = User.find(current_user)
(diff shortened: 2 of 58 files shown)


spec/features/toggle_language_spec.rb
```diff
@@ -0,0 +1,216 @@
+describe 'ToggleLanguage', js: true do
+  let(:user) { create :user1 }
+
+  after(:each) do
+    page.reset!
```
sienaigg: Maybe add a comment here that mentions this clears localStorage in order to prevent test pollution?

spec/features/toggle_language_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  feature 'Toggling the locale dropdown to change the language' do
+    describe 'When signed out' do
```
sienaigg: this is extremely minor, but I think all of these `describes` should be `contexts`, since you're describing similar functionality but in different contexts... (also they all start with "when," which is a good indicator that you're talking about a context)

spec/features/toggle_language_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+    describe 'when signed out and then signed in' do
+      before(:each) do
+        visit root_path
+      end
+
+      after(:each) do
+        logout(:user)
+      end
+
+      it 'language can be toggled on different pages' do
```
sienaigg: maybe this it block is more like "language persists when signing in"? i might even put that (and the one below) in its own `describe` that's "logging in" and "logging out"... that seems more descriptive/clear to me than "when signed out and then signed in"

spec/javascripts/ToggleLocaleSpec.js
```diff
@@ -0,0 +1,18 @@
+describe("toggleLocale", function() {
```
sienaigg: might be cool to test `whenSignedOut` also...

app/assets/javascripts/placeholder_input_size.js
```diff
@@ -0,0 +1,12 @@
+function setInputSize(id) {
+  var length = $(id).attr("placeholder").length;
+  $(id).attr("size", length);
```
sienaigg: this is a little unfortunate but if you don't want to just make the input box bigger all the time, it'll do the trick...

app/controllers/pages_controller.rb
```diff
@@ -34,6 +42,16 @@ def privacy; end
 
   private
 
+  def toggle_when_signed_in(locale)
+    user = User.find(current_user)
```
julianguyen: Omg true I could just use current_user, lol, my bad XD

spec/javascripts/ToggleLocaleSpec.js
julianguyen: Gonna cover this in another PR, since this one is getting big.

PR #462, opened 2017-06-16
Fix spelling error of notifications class name

clairewild:


julianguyen:
Thanks for catching this <3

Issue #429, opened 2017-03-22
Ability to close modals with ESC key and clicking the backdrop


julianguyen:
![wsc_opensourcehours](https://cloud.githubusercontent.com/assets/3010728/24170861/8127b1ba-0e3f-11e7-8b0b-dcfe8661b3b9.png)

<img width="1105" alt="screen shot 2017-03-21 at 10 53 26 pm" src="https://cloud.githubusercontent.com/assets/3010728/24184060/3d094044-0e89-11e7-860c-b066aaa8e173.png">

Currently, the user can only close the modal by hitting the X button. We want to add the ability to close the modal using the ESC key and by clicking the backdrop i.e. area outside of the modal.

evykassirer:
@clairewild and I made some progress https://github.com/julianguyen/ifme/compare/master...evykassirer:closing_modals?expand=1

still missing:

- a few modals (e.g. Moments -> Anxious)
- using escape on some of the modals in Moments/Strategies (at least one works though)
- using escape on the homepage modal

But progress is being made! I might work more on this, but if anyone wants to copy my diff and keep working go for it :)

clairewild:
https://github.com/clairewild/ifme/blob/master/app/assets/javascripts/notifications.js

Have been working on this and as far as I can tell it works for all the modals.

PR #466, opened 2017-06-21
#429 Fix modals to close on esc and add to contributor page

clairewild:


app/assets/javascripts/notifications.js
```diff
@@ -109,6 +114,17 @@ function quickCreate(form, data_type) {
   });
 }
 
+function closeQuickCreate() {
+  if ($(this).closest('.quick_create').attr('id') === 'category_quick_create' && !$('#category_quick_create').hasClass('display_none')) {
```
julianguyen: This function looks good, but there looks like we have a lot of repetitive logic here the only difference is the id of the closet `.quick_create` attribute. 

We could do something like this:
```
function closeQuickCreate() {
  var quickCreateId = $(this).closest(".quick_create").attr("id");
  if ((quickCreateId === "category_quick_create" || quickCreateId === "mood_quick_create" || quickCreateId === "strategy_quick_create") && quickCreateId.hasClass("display_none")) {
  quickCreateId.toggleClass("display_none");
}
  hideBackdrop();
}
```

clairewild: Just went back and refactored this, I also needed to wrap the quickCreateId in $() to use jquery methods on it.

julianguyen:
Hey @clairewild, any updates on this? :)

clairewild:
Hi @julianguyen still have to write tests, will get around to it this week :-)

clairewild:
Added tests for checking if the modals close, one test for tip_close_notifications doesn't work, would love some feedback and any help!

julianguyen:
Taking a look at this now! Sorry for now reviewing it right away!

spec/javascripts/NotificationsSpec.js
```diff
@@ -29,8 +36,65 @@ describe("Notifications", function() {
     expect($('#notifications_none').hasClass("display_block")).toBe(true);
   });
 
+  it("close tips notifications modal", function() { // This test is failing!!
```
julianguyen: So checked out your fork's branch, and this test is failing because it's hitting

```
$('.tip_close_notifications').click(function() {
    hideTipNotifications.call(this);
  });
```
and then 
```
$('.tip_notifications').click(function() {
    debugger;
    hideTipNotifications.call(this);
  });
```

Which means `hideTipNotifications()` is being called twice.

app/assets/javascripts/notifications.js
```diff
@@ -183,15 +198,30 @@ var onReadyNotifications = function() {
...
   });
 
-  $('.tip_close_notifications').click(function(){
-    $(this).closest('.tip_notifications').toggleClass("display_none");
-    hideBackdrop();
+  $('.tip_close_notifications').click(function() {
+    hideTipNotifications.call(this);
+  });
+
+  $('.tip_notifications').click(function() {
```
julianguyen: It seems to me that we don't need this? Can you explain why we are closing this modal when we click it?
When we remove this, the corresponding test passes!

julianguyen: Oh wait, I see this is needed for closing via the backdrop...

julianguyen: Updating the fixture for tip_notifications to this fixes the test.
```
<div class="tip_notifications display_none">
  <div class="tip_notifications_text"></div>
  <div class="tip_close_notifications"></div>
</div>
```

clairewild:
Fixed my keypress function, and the problem with the tip notifications. :-)

PR #469, opened 2017-06-28
Chore: Cop

HashNotAdam:
This addresses only the most basic style guide violations. Once this has been agreed to, I'll start on the next level of refactoring.


HashNotAdam:
Seems a little rough for Code Climate to try to blame me for 154 new issues; they always existed, I just stopped ignoring them 😆

.rubocop.yml
```diff
@@ -1,12 +1,28 @@
...
+
 Documentation:
   Enabled: false
 
+Rails:
+  Enabled: true
+
+# There is a bug in the Rubocop 0.48 parser that causes it to incorrectly
+# complain about parenthese around blocks used with scopes. This should be
+# removed once it is possible to upgrade to 0.49 or above.
```
julianguyen: Can we just upgrade to >=0.49 then?

.rubocop.yml
```diff
@@ -1,12 +1,28 @@
...
-  Exclude:
-    - 'db/**/*'
-    - 'config/**/*'
-    - 'spec/**/*'
-    - !ruby/regexp /old_and_unused\.rb$/
+# Rubocop should be used for style guide adherence, not testing for code smell
+Rails/OutputSafety:
+  Enabled: false
+Rails/SkipsModelValidations:
+  Enabled: false
```
julianguyen: Sorry to flip-flop on this issue, but maybe we want model validations? It seems a little odd to me to disable it everywhere because of that `update_all`. Should we just add validations to that? Thoughts?

app/controllers/application_controller.rb
```diff
@@ -69,114 +69,106 @@ def are_allies(userid1, userid2)
...
   def most_focus(data_type, profile)
-    data = Array.new
+    data = []
 
-    if profile.blank?
-      userid = current_user.id
-    else
-      userid = profile
-    end
+    userid =
```
julianguyen: Should we use a ternary operator i.e. `userid = profile.blank? ? current_user.id : profile`

app/controllers/application_controller.rb
```diff
@@ -69,114 +69,106 @@ def are_allies(userid1, userid2)
...
-          end
+    result = {}
 
-          max = data.max_by { |v| freq[v] }
-          if freq[max] == 0
-            break
-          end
+    unless data.empty?
+      freq = {}
+      3.times do
```
julianguyen: Nice ⚡️

app/controllers/application_controller.rb
```diff
@@ -261,18 +245,17 @@ def get_stories(user, include_allies)
...
-    if moments.count > 0
-      stories = moments.zip(strategies).flatten.compact
-    else
-      stories = strategies.flatten.compact
-    end
+    moments = Moment.where(id: my_moments.map(&:id)).all.order('created_at DESC')
+    strategies = Strategy.where(id: my_strategies.map(&:id)).all.order('created_at DESC')
 
-    stories = stories.sort_by {|x| x.created_at }.reverse!
+    stories =
```
julianguyen: Ternary operator here too?

app/controllers/medications_controller.rb
```diff
@@ -88,26 +88,22 @@ def destroy
...
   # Save refill date to Google calendar
   def save_refill_to_google_calendar(medication)
-    if current_user.google_oauth2_enabled? && medication.add_to_google_cal
-      summary = "Refill for " + medication.name
-      date = medication.refill
-      CalendarUploader.new(summary: summary, date: date, access_token: current_user.token, email: current_user.email).upload_event
-    end
+    return unless current_user.google_oauth2_enabled? && medication.add_to_google_cal
+
+    summary = 'Refill for ' + medication.name
```
julianguyen: Is it better to use string interpolation?

app/controllers/meetings_controller.rb
```diff
@@ -22,15 +22,19 @@ def show
...
   def comment
-    @comment = Comment.new(:comment_type => params[:comment_type], :commented_on => params[:commented_on], :comment_by => params[:comment_by], :comment => params[:comment], :visibility => 'all')
+    @comment = Comment.new(comment_type: params[:comment_type], commented_on: params[:commented_on], comment_by: params[:comment_by], comment: params[:comment], visibility: 'all')
 
-    if !@comment.save
+    # TODO: it is quite unclear what the expected outcome is here. If nothing
+    #       more should be done, then a return statement should be added. If
+    #       the follow work should be performed, then the responce actions
+    #       should be grouped at the end of the method
+    unless @comment.save
```
julianguyen: The expected outcome is to render `{ no_save: true }` to indicate that the comment was able to be saved. Otherwise, create notifications.

.rubocop.yml
HashNotAdam: I'm afraid that 0.48 is the latest version of Rubocop that Code Climate supports and I had to manually set an engine in .codeclimate just to be able to use that version.

.rubocop.yml
HashNotAdam: The [RuboCop documentation](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/SkipsModelValidations) states that this "checks for the use of methods which skip validations".
I'll 100% support your decision—and this is a completely subjective matter—but my opinion is that this is RuboCop checking for code smell.

We can wrap these calls in "rubocop:disable" statements if you choose to continue throwing errors about this.

I like using reek for code smell and RuboCop for enforcing a style guide. I'm sure there would be others with a different opinion. Either way, I'll follow your guidance on the matter since it is so subjective.

app/controllers/application_controller.rb
HashNotAdam: This is a choice you could make. In this PR I am not doing any real refactoring, just making the code match the style guide.

If you are asking my opinion, I'm one of those people who embraces white space for legibility but it isn't something I could be bothered fighting people on. 😆

app/controllers/application_controller.rb
HashNotAdam: In this case, that would make for a pretty log line and you'll just end up with a new RuboCop error about line length

app/controllers/medications_controller.rb
HashNotAdam: 100%, yes
That is something I'd like to see in a future PR but I was worried about this one being too big already, being 68 commits and all.

.rubocop.yml
julianguyen: Reek would be an interesting tool to use for code smells. I'm fine with keeping this disabled then! Thanks for explaining. I don't like wrapping calls in disable statements.

app/controllers/application_controller.rb
julianguyen: Ah ok! Do you mind updating it to a ternary operator then? :) I prefer to minimize lines when possible (and when the conditional isn't complicated)

.rubocop.yml
HashNotAdam: Great, let's get reek installed in a future revision. It's pretty similar to what Code Climate tells you but you can have it running in dev, in the same way you tell people to run rspec and not rely on Circle CI.

app/controllers/application_controller.rb
julianguyen: Oh wait really? Hmm okay just leave it then!

app/controllers/application_controller.rb
julianguyen: Just leave it then!

app/controllers/medications_controller.rb
julianguyen: Okay no problem! Sounds good!

julianguyen:
Should be ready to merge! :D

PR #490, opened 2017-07-16
Upgrade Rails to 4.2.9

HashNotAdam:
Moving to Rails 5.0 is proving to be troublesome.

Before doing that transition, I upgraded to 4.2.9 and there was a reduction in memory usage of around 4% so it seems valuable to release this update while working through the 5.0 update issues.


HashNotAdam:
Sorry, the project was changed between committing this and the PR approval so I've rebased again

PR #512, opened 2017-07-24
Auth key lifespan

### LINKED ISSUE #499: Medications page is blank after submitting new medication
I got a blank page after I submitted information the New Medication form.

I had entered content only in the required fields (i.e., no comments) and did not mark any of the checkboxes.

The page remains blank when I reload it, but displays properly when I request the page from the address bar (Ctrl+L, Enter). It also appears that a duplicate medication was created when I tried to reload the page.

Using Firefox ESR 52.2.0 (64-bit)

abrophy:
resolves #499 

This is a pretty big PR, I've added new fields to user model for storing the google refresh token as well as token expiration time.

A note on the google refresh token, it appears that google only issues these once per app the first time a user gives an app permission such as google calendar/profile access. In order to force google to issue new refresh keys for existing accounts, I've added the userinfo.profile scope to the google omniauth initializer, this was recommended by the team behind the gem on a similar issue.

Also explicitly specifying approval_prompt config was necessary.

This worked for me in dev, but I'm paranoid about it working the same way in production.

So I'm posting this as a WIP right now.

Before it goes live I'll write up a fallback that explicitly signs a person out, and redirects them to signin if the update_access_token method is hit but no refresh_token exists.

### DIFF
--- a/app/controllers/medications_controller.rb
+++ b/app/controllers/medications_controller.rb
@@ -93,21 +93,24 @@ class MedicationsController < ApplicationController
                        medication.add_to_google_cal == '1'
 
     summary = 'Refill for ' + medication.name
-    date = medication.refill
     begin
       CalendarUploader.new(summary: summary,
-                           date: date,
+                           date: medication.refill,
                            access_token: current_user.access_token,
                            email: current_user.email).upload_event
     rescue
-      sign_out current_user
-      respond_to do |format|
-        format.html { redirect_to new_user_session_path }
-        format.json { head :no_content }
-      end
+      return_to_sign_in
       false
     else
       true
+    end
+  end
+
+  def return_to_sign_in
+    sign_out current_user
+    respond_to do |format|
+      format.html { redirect_to new_user_session_path }
+      format.json { head :no_content }
     end
   end
 
--- a/app/controllers/medications_controller.rb
+++ b/app/controllers/medications_controller.rb
@@ -51,37 +51,36 @@ class MedicationsController < ApplicationController
   # POST /medications.json
   def create
     @medication = Medication.new(medication_params)
+    return unless save_refill_to_google_calendar(@medication)
 
-    if save_refill_to_google_calendar(@medication)
-      if @medication.save
-        respond_to do |format|
-          format.html { redirect_to medication_path(@medication) }
-          format.json { render :show, status: :ok, location: @medication }
-        end
-      else
-        respond_to do |format|
-          format.html { render :new }
-          format.json { render json: @medication.errors, status: :unprocessable_entity }
-        end
+    if @medication.save
+      respond_to do |format|
+        format.html { redirect_to medication_path(@medication) }
+        format.json { render :show, status: :ok, location: @medication }
+      end
+    else
+      respond_to do |format|
+        format.html { render :new }
+        format.json { render json: @medication.errors,
+                             status: :unprocessable_entity }
       end
     end
-
   end
 
   # PATCH/PUT /medications/1
   # PATCH/PUT /medications/1.json
   def update
-    if save_refill_to_google_calendar(@medication)
-      if @medication.update(medication_params)
-        respond_to do |format|
-          format.html { redirect_to medication_path(@medication) }
-          format.json { render :show, status: :ok, location: @medication }
-        end
-      else
-        respond_to do |format|
-          format.html { render :new }
-          format.json { render json: @medication.errors, status: :unprocessable_entity }
-        end
+    return unless save_refill_to_google_calendar(@medication)
+    if @medication.update(medication_params)
+      respond_to do |format|
+        format.html { redirect_to medication_path(@medication) }
+        format.json { render :show, status: :ok, location: @medication }
+      end
+    else
+      respond_to do |format|
+        format.html { render :new }
+        format.json { render json: @medication.errors,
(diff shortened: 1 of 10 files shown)


app/models/user.rb
```diff
@@ -87,11 +88,44 @@ def self.find_for_google_oauth2(access_token, _signed_in_resource = nil)
...
+
+  def update_access_token
+    url = URI("https://accounts.google.com/o/oauth2/token")
+    refresh_token_params = { 'refresh_token' => refresh_token,
+                              'client_id'     => ENV['GOOGLE_CLIENT_ID'],
+                              'client_secret' => ENV['GOOGLE_CLIENT_SECRET'],
+                              'grant_type'    => 'refresh_token' }
+
+    time_of_request = Time.zone.now
+    # TODO handle a failed request
```
julianguyen: Are you going to throw an exception here?

app/models/user.rb
```diff
@@ -87,11 +88,44 @@ def self.find_for_google_oauth2(access_token, _signed_in_resource = nil)
...
+    response = Net::HTTP.post_form(url, refresh_token_params)
+    decoded_response = JSON.parse(response.body)
+
+    new_expiration_time =  time_of_request + decoded_response["expires_in"]
+    new_access_token = decoded_response["access_token"]
+
+    self.update!(
```
julianguyen: Do we need to confirm that the values for new_expiration_time and new_access_token are non-nil?

julianguyen:
Looking good so far. I like the fallback idea 👍

app/models/user.rb
abrophy: yes, but a failed request will have to be handled the same way as expired tokens with no refresh token. Informing the user something went wrong with a flash message and asking them to sign in once more.

So I hadn't wrapped it in a rescue until I had established that process, which is the edge case I spoke of earlier.

app/models/user.rb
abrophy: probably not, I had my doubts about this as well but submitted it this way for thoroughness, thanks for bringing this up, I'll clean it up.

abrophy:
I've got the circleci tests passing, but the controller now needs some refactoring in order to 100% code climate, it is working however and can be merged to fix the bug.

app/controllers/medications_controller.rb
```diff
@@ -51,29 +51,36 @@ def edit
   # POST /medications.json
   def create
     @medication = Medication.new(medication_params)
-    respond_to do |format|
-      if @medication.save
-        save_refill_to_google_calendar(@medication)
+    return unless save_refill_to_google_calendar(@medication)
+
+    if @medication.save
+      respond_to do |format|
```
julianguyen: So the respond_to blocks starting at lines 57 and 62 are similar. We could move this out into a helper method. This would stop Codeclimate from complaining lol.

app/controllers/medications_controller.rb
```diff
@@ -51,29 +51,36 @@ def edit
...
 
   # PATCH/PUT /medications/1
   # PATCH/PUT /medications/1.json
   def update
-    respond_to do |format|
-      if @medication.update(medication_params)
-        save_refill_to_google_calendar(@medication)
+    return unless save_refill_to_google_calendar(@medication)
+    if @medication.update(medication_params)
+      respond_to do |format|
```
julianguyen: Similar comment as above about combining duplicate code!

app/controllers/medications_controller.rb
```diff
@@ -90,15 +97,26 @@ def destroy
 
   # Save refill date to Google calendar
   def save_refill_to_google_calendar(medication)
-    return unless current_user.google_oauth2_enabled? &&
-                  medication.add_to_google_cal == '1'
+    return true unless current_user.google_oauth2_enabled? &&
+                       medication.add_to_google_cal == '1'
```
julianguyen: `add_to_google_cal` should be true/false instead of a string.

HashNotAdam: Bonus points if you add_to_google_cal? 😄

abrophy: thanks for your patience and input! I need to learn to trust the instruments, instead of checking back for reviews all the time,

julianguyen: Not a problem, I think enforcing people to use jslint and rubocop in their workflows will help out a lot with that :D

abrophy: if I drop the string and do a simple comparison it fails, I think it's due to it being a temporary property, being attr_accessor on the model and passed from the form.

I haven't thought of a way to work around this yet, but the issue was raised somewhere else if I can recall about storing the google reminder state on the model and handling it more intelligently after initial creation. Would this perhaps require its own issue or do you want to rope it in here as well?

app/helpers/medication_refill_helper.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module MedicationRefillHelper
+  # Save refill date to Google calendar
+  def save_refill_to_google_calendar(medication)
+    return true unless current_user.google_oauth2_enabled? &&
+                       medication.add_to_google_cal == '1'
```
julianguyen: Did you investigate `add_to_google_cal` being true/false instead?

abrophy:
And there was much rejoicing!

Issue #455, opened 2017-06-06
Secret share Moments


julianguyen:
The ability to share secret links to your Moments posts so that non if me users can view them! Sometimes people don't like creating new accounts. Or you want to have a discussion about something off the platform.

SophieRoseMcDonald:
![](https://github.trello.services/images/mini-trello-icon.png) [(3) Secret Share Moments](https://trello.com/c/jpYBcg0c/27-3-secret-share-moments)

andy5995:
This was included with #623 ? Ready to be closed?

HashNotAdam:
Thanks, @andy5995, we should have closed this already

HashNotAdam:
Thanks, @andy5995, we should have closed this already

PR #513, opened 2017-07-24
[#455] Add Secret Share Button

jnamster:


db/migrate/20170724205314_add_secret_shares_to_moments.rb
```diff
@@ -0,0 +1,6 @@
+class AddSecretSharesToMoments < ActiveRecord::Migration
+  def change
+    add_column :moments, :secret_share_identifier, :string
+    add_column :moments, :secret_share_expires_at, :datetime
```
HashNotAdam: Have you had a discussion about database indices?
When loading the share URL, you are going to need to find on the moments table for a secret_share_identifier. If there is no index on this field, the database is going to need to check every record looking for a match. In development, you'd never notice this but, when you have lots of records in production, this can be extraordinarily slow.

app/controllers/secret_shares_controller.rb
```diff
@@ -0,0 +1,9 @@
+class SecretSharesController < ApplicationController
+
+  def create
+    moment = Moment.friendly.find(params[:moments])
+    moment.update :secret_share_identifier => SecureRandom.uuid
```
HashNotAdam: You should take a look at `update!`; it does the same thing except it throws an exception if the update fails. `update` just silently does nothing if there is a problem.

This will upset the 👮 because it uses Ruby 1.9 syntax. To keep from trouble with the law, use `moment.update!(secret_share_identifier: SecureRandom.uuid)`

I expect this logic will only get more complex with time and so, for testabiliy, you'll want to move it into the model. In the model, you will also need to look into adding a uniqueness validator.

jnamster: Thanks for the feedback. I have looked into update vs update! I have made the change to the controller and committed the change. I am keen to learn more about your thoughts about moving into the model and adding a uniqueness validator.

jnamster:
Noted. Do you think it would be better then to start the label with something like this [#455 WIP] ...?

HashNotAdam:
Ultimately, what ever @julianguyen wants is the most important factor but I would suggest `[#455][WIP]`

app/controllers/secret_shares_controller.rb
HashNotAdam: Really, controllers should be delegators. While the distinction can get a little grey at times, mostly their work is asking different parts of the application to complete tasks and then packaging up the results for the response.

IMO, in this case, I would say that the controller shouldn't know that you create a share by generating a UUID which is stored in the secret_share_identifier of a moment. I think the controller would get the moment and then ask the moment to generate a share. Any logic regarding how that share is created is contained within the model.

I don't want to prescribe any particular naming conventions but just to help the discussion, it could be that the controller calls `moment.generate_secret_share!` which does the `update!` and returns true.

Take a look over [Active Record Validations](http://guides.rubyonrails.org/active_record_validations.html)—there is a section on [validating uniqueness](http://guides.rubyonrails.org/active_record_validations.html#uniqueness).

While the probability of 2 UUIDs clashing is extremely low, it is still possible.

As this solution develops, this validation may change to account for the expiry but you can start with a simple uniqueness validator and generate a test to prove that it works.

app/views/moments/show.html.erb
```diff
@@ -1,5 +1,7 @@
 <% title @moment.name %>
 <%= render :partial => '/moments/info', locals: { data: @moment } %>
+<%= button_to "Secret Share", secret_shares_path(moment: @moment.name) %>
```
julianguyen: This should be i18n key. You'll need to also translate this to Spanish and Portuguese. For Spanish, you can ask in the #es-translations and for Portuguese, #br-translations

db/migrate/20170724205314_add_secret_shares_to_moments.rb
julianguyen: +1 on what Adam said

app/controllers/secret_shares_controller.rb
```diff
@@ -0,0 +1,9 @@
+class SecretSharesController < ApplicationController
+
+  def create
+    moment = Moment.friendly.find(params[:moments])
```
julianguyen: As part of TDD, I would start writing `secret_shares_controller_spec.rb` so that you can test as you go.

spec/controllers/secret_shares_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+describe SecretSharesController do
+  describe "POST create" do
+    it "Creates Secret Share Identifier" do
+      new_user = create(:user1)
+      sign_in new_user
+      new_category = create(:category, userid: new_user.id)
+      new_mood = create(:mood, userid: new_user.id)
+      new_strategies = create(:strategy, userid: new_user.id)
+      new_moment = create(:moment, userid: new_user.id, category: Array.new(1, new_category.id), mood: Array.new(1, new_mood.id), strategies: Array.new(1, new_strategies.id))
```
HashNotAdam: I'd love to have a chat about this tomorrow but there are a couple of things I'd like to point out now.

It doesn't make sense to use `Array.new` if you are only creating 1 item, just use `[]` like `[new_category.id]`. I've been removing any instances of these I find in the Rails 5 upgrade.

Since the keys in the creation of a moment are singular, it tells me you don't actually need them to be arrays, you can just:
```
category: new_category,
mood: new_mood,
strategies: new_strategies
```

Saying all this, I don't think you need to create all those objects. Using `create` actually stores a record in the database so, for the sake of test performance, you only want to do it when you really need to.
I believe you only need to create a user and a moment.

Issue #491, opened 2017-07-16
Write unit test to check that the translation files do not have missing keys


julianguyen:
This will automate determining whether we are missing translation keys across all of our yml files!

This looks like a promising route to take: http://blog.kiskolabs.com/post/908453942/comparing-rails-locale-files-for-missing

SophieRoseMcDonald:
![](https://github.trello.services/images/mini-trello-icon.png) [Write unit test to check that the translation files do not have missing keys #491](https://trello.com/c/tgMZ3Sz5/58-write-unit-test-to-check-that-the-translation-files-do-not-have-missing-keys-491)

PR #514, opened 2017-07-25
[#491] Unit test for compare_locales

SophieRoseMcDonald:
I have been working on the unit test to check that the translation files do not have missing keys [#491].
Please also note, this isn't supposed to be a helper, it should probably be a rake task. I realised this after I had been working on it for a while and will adjust the files.


app/helpers/compare_locales_helper.rb
```diff
@@ -0,0 +1,43 @@
+module CompareLocalesHelper
+  include ApplicationHelper
+
+  LOCALE_1 = "./config/locales/en.yml"
+  LOCALE_2 = "./config/locales/es.yml"
```
julianguyen: We will also have to run these tests against all files including `ptbr.yml` and any future translations. We should use regex to access these files (something like `"./config/locales/*.yml"`)

app/helpers/compare_locales_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+    yaml_2 = YAML.load(File.open(File.expand_path(locale_2)))
+
+    keys_1 = flatten_keys(yaml_1[yaml_1.keys.first])
+    keys_2 = flatten_keys(yaml_2[yaml_2.keys.first])
+
+    missing = keys_2 - keys_1
+    binding.pry
+    file = locale_1.split('/').last
+    if missing.any?
+      puts "Missing from #{file}:"
```
julianguyen: Instead of printing errors (was originally done cause this code was a rake task), we should throw exceptions

julianguyen:
Looking good so far :D

spec/support/compare_locales_support.rb
```diff
@@ -0,0 +1,36 @@
...
+        keys.concat(flatten_keys(hash[key], "#{prefix}#{key}."))
+      else
+        keys << "#{prefix}#{key}"
+      end
+    end
+  end
+
+  def self.compare_locale_hashes(primary_locale, locale_to_compare)
+    primary_keys = flatten_keys(primary_locale[primary_locale.keys.first])
+    keys_to_compare = flatten_keys(locale_to_compare[locale_to_compare.keys.first])
```
HashNotAdam: `primary_locale.values.first` and `locale_to_compare.values.first` is a little more succinct

spec/support/compare_locales_support.rb
```diff
@@ -0,0 +1,36 @@
+module CompareLocalesSupport
+
+  LOCALE_1 = './config/locales/en.yml'
+  LOCALE_2 = './config/locales/es.yml'
+
+  require 'yaml'
```
HashNotAdam: Shouldn't be needed in Rails

spec/support/compare_locales_support.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.compare_locale_hashes(primary_locale, locale_to_compare)
+    primary_keys = flatten_keys(primary_locale[primary_locale.keys.first])
+    keys_to_compare = flatten_keys(locale_to_compare[locale_to_compare.keys.first])
+
+    primary_keys - keys_to_compare
+  end
+
+  def self.compare(primary_locale_file_name, locale_file_name_to_compare)
+    primary_locale = YAML.load(File.open(File.expand_path(primary_locale_file_name)))
+    locale_to_compare = YAML.load(File.open(File.expand_path(locale_file_name_to_compare)))
```
HashNotAdam: `YAML.load_file(primary_locale_file_name)`
`YAML.load_file(locale_file_name_to_compare)`

spec/support/compare_locales_support.rb
```diff
@@ -0,0 +1,36 @@
+module CompareLocalesSupport
+
+  LOCALE_1 = './config/locales/en.yml'
+  LOCALE_2 = './config/locales/es.yml'
```
HashNotAdam: Can just be `config/locales/es` if you use `load_file` described below

spec/compare_locales_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      'edited_updated_at'=>'<strong>Created:</strong> %{created_at} <em>(edited %{updated_at})</em>',
+      'salutation'=>'Hi %{name},',
+      'click_here'=>'click here',
+      'language'=>'Language'
+    }
+  }
+}
+  specify 'spanish locale file has nothing missing' do
+    english = './config/locales/en.yml'
+    spanish = './config/locales/es.yml'
```
HashNotAdam: Probably better that you use the constants you defined on CompareLocalesSupport and DRY.
Since we will be changing how this works in a future version, for now just use `CompareLocalesSupport:: LOCALE_1`

spec/compare_locales_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    expect(CompareLocalesSupport.compare(english, spanish)).to match_array([])
```
HashNotAdam: Just as a "did you know", I think this could also be written as `.to be_empty`

spec/support/compare_locales_support.rb
SophieRoseMcDonald: Noted. Thank you :+1:

spec/support/compare_locales_support.rb
SophieRoseMcDonald: :+1:

spec/support/compare_locales_support.rb
SophieRoseMcDonald: Open to suggestions, if you can think of a better name.

HashNotAdam: I'm not terribly concerned about the name—although `comparison_locale` would add a little consistency—but it's easier to read as `primary_locale.values.first` rather than `primary_locale[primary_locale.keys.first]`

spec/compare_locales_spec.rb
HashNotAdam: I knew I liked that guy

spec/support/compare_locales_support.rb
SophieRoseMcDonald: Yes, you are correct. I am still getting the hang of naming things.

HashNotAdam: So am I ;)

spec/support/compare_locales_support.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module CompareLocalesSupport
+  LOCALE_1 = './config/locales/en.yml'
```
julianguyen: Looks like we don't need these locale constants anymore

PR #526, opened 2017-08-01
Updating default font to latest version for better i18n support

baohouse:
Lato support for Vietnamese is lacking; Roboto has better support for it, as well as having font-weights 100 and 300. San Francisco and Segoe UI (in the latest OS X and Windows OS) will be primary default, but we can fall back to Roboto if they aren't available.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -426,7 +426,7 @@ a.page_search:hover {
 .headline {
   color: $header-text-color;
   font-size: $page-title;
-  text-shadow: 0px 0px 10px $font-family;
+  text-shadow: 0px 0px 10px $header-text-color;
 
   @media screen and (max-width: $medium) {
     font-size: $font30;
--- a/app/assets/stylesheets/base/_fonts.scss
+++ b/app/assets/stylesheets/base/_fonts.scss
@@ -1,5 +1,5 @@
 // This file contains the base fonts for the application.
 
-$font-family: -apple-system, BlinkMacSystemFont,
-  'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell',
-  'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
+@import url(https://fonts.googleapis.com/css?family=Roboto:100,400,300);
+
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -426,7 +426,7 @@ a.page_search:hover {
 .headline {
   color: $header-text-color;
   font-size: $page-title;
-  text-shadow: 0px 0px 10px $font-family;
+  text-shadow: 0px 0px 10px $header-text-color;
```
julianguyen: Good catch!

app/assets/stylesheets/base/_fonts.scss
```diff
@@ -1,5 +1,5 @@
 // This file contains the base fonts for the application.
 
-@import url(https://fonts.googleapis.com/css?family=Lato:100,400,300);
+@import url(https://fonts.googleapis.com/css?family=Roboto:100,400,300);
 
-$font-family: 'Lato';
+$font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', sans-serif;
```
julianguyen: We want to use Lato as the default though. We could fallback to another font if it's in Vietnamese though.

julianguyen: LOL also https://www.fontsquirrel.com/fonts/list/language/vietnamese says Lato supports Vietnamese. LIES.

baohouse: I saw a footnote on google font lato that a previous version has the extended Latin set, which includes Vietnamese. We can keep Lato I think, if only to support Vietnamese. But in the long run we'll need to support language-specific base fonts because you're not going to be able to have one font cover everything. We can do a short term fix but we'll have to architect once there's a need for Japanese, Chinese, Thai or some language that Lato doesn't cover.

baohouse: @julianguyen I found out that Google Fonts uses version 1.0 of Lato (https://github.com/google/fonts/issues/6), with no certainty of when 2.0 will be uploaded. I've instead switched the CDN to jsdelivr which hosts version 2.0.

baohouse:
Changed the scope of this PR to updating the current font version instead of changing the default font entirely.

Issue #456, opened 2017-06-06
Update app from Rails 4.2.6 to Rails 5


julianguyen:

HashNotAdam:
I'd be happy to help out with this one.
Is the plan to get to 5.1 (via 5.0)?
Any known issues that could make it tricky?

julianguyen:
@HashNotAdam Yeah I think getting it to 5.0 is a good start then 5.1 will be more straightforward. Do you think just upgrading to 5.1 is fine? There are no known issues!

HashNotAdam:
It will most likely be completely fine to produce a PR to goes straight to 5.1 but I'll do the update in two steps and just only submit the final position. There will need to be changes to things like Selenium and the documentation regarding Firefox so I think it's better if the community only sees one change there.

PR #528, opened 2017-08-02
[#456][WIP] Upgrade to Rails 5.0.5

HashNotAdam:
Still filled with errors and deprecation notices and needs many updates to documentation but posted to begin a discussion.


.gitignore
```diff
@@ -31,3 +31,6 @@
 
 # Ignore auto-generated translations.js
 /public/javascripts/translations.js
+
+# Secrets should be secret
+/config/secrets.yml
```
julianguyen: Hmmm I think we need this [file](https://github.com/julianguyen/ifme/blob/master/config/secrets.yml) not to be secret though since we're getting `ENV["SECRET_KEY_BASE"]` from the .env files.

HashNotAdam:
Before I spend time on it, I don't suppose someone has already had to work out how to fix this deprecation notice?

```
DEPRECATION WARNING: #to_hash unexpectedly ignores parameter filtering, and will change to enforce it in Rails 5.1. Enable `raise_on_unfiltered_parameters` to respect parameter filtering, which is the default in new applications. For the existing deprecated behaviour, call #to_unsafe_h instead. (called from merge at /Users/arice/dev/ifme/app/controllers/strategies_controller.rb:301)
```

julianguyen:
@HashNotAdam I don't think anyone is taking a look at the deprecation notice. Do you want to tackle it? Otherwise, I can look at that issue.

HashNotAdam:
@julianguyen no no, it has just come up in the upgrade and it's not super clear. I'm sure I can work it out but, if someone had seen it on another project, it could save me some time.

It was the last of the deprecation warnings I need to solve but then I rebased and someone introduced some new ones for me 😆

HashNotAdam:
@julianguyen there is a lot of boring stuff in the recent update to this PR but you should probably be aware of the change to Selenium, as described in the [README.md](https://github.com/HashNotAdam/ifme/blob/e2ea13e347a75c5ceab76ceeacb07dc2a3fe188c/README.md#testing-guidelines)

HashNotAdam:
I have found no relevant difference in memory use between this and the master branch so I'm making it as complete

HashNotAdam:
Except the tests just added to master break in Rails 5 :sob: :sob: :sob:
(╯°□°）╯︵ ┻━┻

HashNotAdam:
Fixes #456

Issue #523, opened 2017-07-31
Adding auto-complete for adding categories/moods to Moments/Strategies


julianguyen:
I started a branch where I'm working on this i.e. `moments-strategies-cleanup`.
I'm also going to be working on cleaning up the partials that are used in this editor.

# Screenshots:
![image](https://user-images.githubusercontent.com/3010728/29008991-f28dad72-7ad3-11e7-9a92-498ea7f7e503.png)
![image](https://user-images.githubusercontent.com/3010728/28957247-02c65f2a-78a6-11e7-90aa-7776eefc9aa9.png)
![image](https://user-images.githubusercontent.com/3010728/28957269-181e42ac-78a6-11e7-9ffa-3daf0b736043.png)

PR #533, opened 2017-08-04
[#523] Auto-complete for Moments/Strategies form

julianguyen:
Read more here: https://github.com/julianguyen/ifme/issues/523

- Refactored notifications.js and moved code into separate JS files
- Add Autocomplete via jQuery UI to checkboxes for Categories/Moods/Strategies in Moments and Categories in Strategies
- Rename Strategies column in the Moments table to be Strategy for consistency, makes it easier to share code with Mood and Category


app/assets/javascripts/autocomplete_checkboxes.js
```diff
@@ -0,0 +1,90 @@
+function hitEnter(event, element, formType, inputType) {
+  if (event.which === 13) {
+    enterOnAutocomplete(element, formType, inputType);
```
nma: the input field has `Press ENTER to add`, but I can't open the quick create window unless I typed a single character. Would this behaviour seem confusing to some users?

app/assets/javascripts/quick_create.js
```diff
@@ -0,0 +1,93 @@
+function addToAutocomplete(json) {
+  var source = JSON.parse($("#" + json.autocomplete_id).attr("data-autocomplete-source"));
+  source.unshift(json.name)
```
nma: could these extra params be filtered on the ruby side?

app/assets/javascripts/quick_create.js
```diff
@@ -0,0 +1,93 @@
...
+  $("#" + json.autocomplete_id).attr("data-autocomplete-source", JSON.stringify(source));
+  var sourceIds = JSON.parse($("#" + json.autocomplete_id).attr("data-autocomplete-source-ids"));
+  sourceIds.unshift(json.id);
```
nma: ... vice versa, if we don't need them, would it be better to centralize it in the backend?

app/assets/javascripts/viewers.js
```diff
@@ -0,0 +1,9 @@
+var onReadyViewers = function() {
+  if (newOrEdit(["moments", "strategies"])) {
+  	$("#viewers_all").click(function() {
+  		$("#viewers_list :checkbox").prop("checked", $(this).prop("checked"));
```
nma: What happens when viewers > 20? will we have a long list of checkboxes?  - just something I noticed when I was playing with the branch

config/locales/ptbr.yml
```diff
@@ -59,6 +59,7 @@ ptbr:
      location: "Localização"
      description: 'Descrição'
      error_explanation: 'Please fill out the marked fields!'
+     press_enter: 'Pressione ENTER para adicionar'
    time_ago: "%{date} atrás"
    write_to_us: "join.ifme@gmail.com"
```
julianguyen: Ahh good question and catch, the reason why we have this is that the English sentence did not match structurally with the Spanish and Portuguese translations. Hmm maybe write_to_us can be renamed to something else then.

app/assets/javascripts/autocomplete_checkboxes.js
julianguyen: Yes, that is correct. So they can begin to type their new quick create content and then that will open the modal with the name field pre-populated.

app/assets/javascripts/autocomplete_checkboxes.js
```diff
@@ -0,0 +1,90 @@
...
+    return false;
+  }
+}
+
+function openQuickCreate(input, inputType) {
+  showBackdrop();
+  $("#" + inputType + "_quick_create").toggleClass("display_none");
```
julianguyen: That's a good point, hmm sounds like we can just refactor that all over the codebase XD

app/assets/javascripts/quick_create.js
julianguyen: It could, but then we would have make an Ajax call, and it seems unnecessary to me if it's already on the DOM

app/assets/javascripts/viewers.js
julianguyen: My plan is once I merge this PR, I'll implement a similar auto-complete feature for Viewers so that instead of displaying a potentially long list of allies, users can just type in people.

julianguyen:
Thanks so much for reviewing @nma :D

nma:
No problem, my github notifications are getting filtered, so pardon the delay on the follow up !

app/assets/javascripts/quick_create.js
nma: yeah that comment could be ignored, I don't have a habit to lean on .erb templates, but as it is now, we shift the heftier bits onto .erbs / rails so makes sense.

app/assets/javascripts/autocomplete_checkboxes.js
nma: okay, so outlining my steps.
1. I see the default input field text message  `Press ENTER to create a category` 
2. I hit enter a few times, but there is no action. (A little confused)
3. I enter a character (but it also removes the default help message), then it makes sense.

It feels like this wants to be a tooltip. I've used `tooltipster` to get that, so we might be able to open a separate improvement down the road.
https://github.com/Amunds/tooltipster-rails
http://iamceege.github.io/tooltipster/

For now the code looks good, so we can choose to defer to a new issue.

julianguyen: I can defer the tooltip to a follow up PR :D Thanks for the feedback!

PR #542, opened 2017-08-16
Fix allies search bug

julianguyen:


spec/controllers/search_controller_spec.rb
```diff
@@ -24,15 +30,15 @@
...
           get :index, search: { email: 'bar@email.com' }
 
-          expect(assigns(:matching_users)).to eq([user_three])
+          expect(assigns(:matching_users)).to include(user_three)
         end
       end
 
       context 'when have no passed email' do
         it 'sets the correct instance variables' do
           get :index, search: { email: '' }
```
HashNotAdam: Seems like this and the earlier test for redirection are the same. Could just use these parameters in the above test to ensure it redirects.

app/controllers/search_controller.rb
```diff
@@ -4,8 +4,14 @@ class SearchController < ApplicationController
   def index
     mail = params[:search][:email]
 
-    @matching_users = User.where.not(id: current_user.id).all
-    @matching_users = @matching_users.where(email: mail.strip) if mail.present?
+    if mail.present? && mail != current_user.email
```
HashNotAdam: So, if the user searches for themselves you want it to allies rather than show no results?

spec/controllers/search_controller_spec.rb
julianguyen: Good point! I'll consolidate them! Thanks :)

app/controllers/search_controller.rb
```diff
@@ -4,8 +4,14 @@ class SearchController < ApplicationController
   def index
     mail = params[:search][:email]
```
HashNotAdam: This actually causes a fatal error if no search parameters are set.

What about something like:
```ruby
def index
  permitted = params.require(:search).permit(:email)
  email = permitted[:email].strip
  raise ActionController::ParameterMissing unless permitted.present?

  @matching_users = User.where(email: email).where.not(id: current_user.id)
rescue ActionController::ParameterMissing
  respond_to do |format|
    format.html { redirect_to allies_path }
  end
end
```

HashNotAdam: Code example now out of date given 👇

julianguyen: Thanks great suggestion - I'll do this!

app/controllers/search_controller.rb
```diff
@@ -2,10 +2,14 @@
 
 class SearchController < ApplicationController
   def index
-    mail = params[:search][:email]
-
-    @matching_users = User.where.not(id: current_user.id).all
-    @matching_users = @matching_users.where(email: mail.strip) if mail.present?
+    permitted = params.require(:search).permit(:email)
+    raise ActionController::ParameterMissing if permitted.blank?
```
HashNotAdam: This raise was only valuable for testing if email is blank. If there are no permitted values, the previous statement should raise. If you are happy to do a search where email = '', then you can probably remove this line

Issue #486, opened 2017-07-15
Add to the mental health resources page!


julianguyen:
We have a [page](http://if-me.org/resources) on our site with mental health resources categorized under communities, education, hotlines, and services with relevant tags.

Post your favourite resources here. Feel free to make a pull request with yours, be sure include internationalization for the tags. Please use our #i18n channel on Slack to ask for translations!

shelleyvadams:
Still getting my bearings here, but I have years of bookmarks -- https://www.bibsonomy.org/user/shelley.adams/mentalhealth -- gleaning welcome.

julianguyen:
Wow, this is amazing, thank so much for sharing @shelleyvadams! <3 <3 <3
 Let's definitely add this!

Caro-Lin:
Dear all,

I'm totally new to the open source community and just tried to make my first pull-request by adding some entries to the resources page. I also created a list of books, videos, talks, all mental health related that might be helpful but doesn't really fit the criteria for the existing categories Education, Communities and Hotlines. Feel free to use it [gist-list](https://gist.github.com/Caro-Lin/6755faa435bfe488924d0e10650423da).

andy5995:
Is this a resource you'd like added?

[Canvas of the Minds](https://acanvasoftheminds.com/)

andy5995:
I'm adding more resources and will wait a few days to submit the PR. If anyone wants to suggest a resource, I'll add it.

andy5995:
I'd like to add some more music therapy resources @deejayres . I saw some mentioned in [your article](https://medium.com/ifme/13-songs-to-reduce-anxiety-stress-58eb9b362867); any other suggestions?

desirottman:
@andy5995 in my last blog post I mentioned https://www.looplabs.com/beta and https://www.ambient-mixer.com/ as places online that people with no musical experience could create tracks.. i'm digging around for more!

desirottman:
http://musictherapyactivities.wikia.com/wiki/Music_Therapy_Activities_Wiki

andy5995:
@desirottman Since you're interested in getting experience with git, I decided to submit my existing changes, so you can do a separate PR with some of your music therapy resources (and other resources you might think of) :) Let me know if you have any questions

sexybiggetje:
For the upcoming Dutch locale ('nl') one could add the following:

[...]

PR #544, opened 2017-08-16
[#486] tags added in en.yml

Caro-Lin:
tags forum and learning added


julianguyen:
@Caro-Lin You'll need to add the same values for ptbr and es! :D

PR #557, opened 2017-08-22
Fix CircleCI config

julianguyen:


julianguyen:
Daannng thank you @HashNotAdam!!!!

PR #559, opened 2017-08-23
Revert to CircleCI 1

HashNotAdam:
While CircleCI 2.0 appears to be a lot faster, and more flexible, than version 1, it does not currently have streamlined support for Heroku deployments.


HashNotAdam:
Moving to a new PR not coming from my master branch

Issue #543, opened 2017-08-16
Add messaging that says no search results found for Allies page


julianguyen:
Currently, the page looks like this when no results are found:
![image](https://user-images.githubusercontent.com/3010728/29346942-b6dba0be-81fd-11e7-8fc2-80ae842a6085.png)

Let's add some messaging to the body that says "User not found", be sure to translate this into Spanish and Portuguese too :)

geoninja:
@julianguyen @katrynmc This would be "usuário não encontrado" in Portuguese.

PR #568, opened 2017-08-25
[#543] Add messaging that says no search results found for Allies page

katrynmc:
image 2017-08-25 at 4 20 24 pm

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -5,6 +5,7 @@ class SearchController < ApplicationController
     permitted = params.require(:search).permit(:email)
     raise ActionController::ParameterMissing if permitted.blank?
     @matching_users = search_by_email(permitted[:email].strip)
+    @email_query = permitted[:email]
   rescue ActionController::ParameterMissing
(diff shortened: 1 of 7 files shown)


app/views/search/index.html.erb
```diff
@@ -27,4 +28,8 @@
           <% end %>
         </div>
     <% end %>
+  <% end %>
+  <% if @matching_users.empty? %>
```
julianguyen: You could just put this as an else since you already checked in line 13!

app/controllers/search_controller.rb
```diff
@@ -5,6 +5,7 @@ def index
     permitted = params.require(:search).permit(:email)
     raise ActionController::ParameterMissing if permitted.blank?
     @matching_users = search_by_email(permitted[:email].strip)
+    @email_query = permitted[:email]
```
julianguyen: We should also test this in `spec/controllers/search_controller_spec.rb`!

katrynmc: k!

Issue #564, opened 2017-08-25
Create helper method to handle redirects in controllers


julianguyen:
There is **a lot** of controller code that looks like this
```
respond_to do |format|
  format.html { redirect_to moods_path }
  format.json { head :no_content }
end
```

To minimize on code repetition, it would be great to call a single helper method and perhaps pass in the desired path as an argument.

camillevilla:
@julianguyen I found a method `redirect_to_path` in `app/controllers/application_controller.rb:333`. I did a lil' refactor in the categories controller, ran the tests, and it seems to work! Would it be cool if I start working on a PR?

julianguyen:
@camillevilla Heck yes! Assign yourself to the issue once you get added as a collaborator :D

Issue #534, opened 2017-08-04
Missing categories_controller_spec.rb and moods_controller_spec.rb


julianguyen:
We're missing specs! Ahh!

timorthi:
I'd like to work on this! :)

julianguyen:
@timorthi Yeah feel free to take this issue! Adding you as a collaborator to this repo so I can assign it to you. Also can you email join.ifme@gmail.com so we can add you to our Slack? Thanks!

timorthi:
Oops, I guess I should only be referencing the issue in the PR to avoid the commit spam. My bad!

julianguyen:
Woo thanks for working on this @timorthi :D

PR #573, opened 2017-08-25
DRY out simple controller redirects

camillevilla:
## Summary
This refactor addresses #564 and replaces 

```
respond_to do |format|
  format.html { redirect_to moods_path }
end
```

with the `redirect_to_path` method in the following controllers:
- categories
- groups
- medications
- meetings
- moments
- moods
- search
- strategies

## Notes
Hi @julianguyen! I tried to make sure I "broke" tests by (1) commenting out redirects, (2) replacing existing code with `redirect_to_path` (3) making sure tests passed again. 

But there were a couple of places where controller test coverage could be improved. In particular :
- `medications_controller_spec.rb`
- `meetings_controller_spec.rb`
- `moments_controller_spec`

These could be good tickets to add to Improving Test Coverage, something like #534.

### DIFF
--- a/app/controllers/medications_controller.rb
+++ b/app/controllers/medications_controller.rb
@@ -19,10 +19,7 @@ class MedicationsController < ApplicationController
       @page_edit = edit_medication_path(@medication)
       @page_tooltip = t('medications.edit_medication')
     else
-      respond_to do |format|
-        format.html { redirect_to medications_path }
-        format.json { head :no_content }
-      end
+      redirect_to_path(medications_path)
     end
   end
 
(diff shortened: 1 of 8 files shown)


julianguyen:
Could you create an issue for improving the test coverage into those specs you identified?

app/controllers/medications_controller.rb
```diff
@@ -73,18 +67,12 @@ def update
   # DELETE /medications/1.json
   def destroy
     @medication.destroy
-    respond_to do |format|
-      format.html { redirect_to medications_path }
-      format.json { head :no_content }
-    end
+    redirect_to_path(medications)
```
HashNotAdam: This one should be `medications_path`

Can you write a test that proves this doesn't work and then fix it?

HashNotAdam:
@julianguyen Oh.. boo... you merged before I could write above

app/controllers/medications_controller.rb
julianguyen: Oh shoot yeah. Can you address this in another PR @camillevilla?

camillevilla: whoops! will do @HashNotAdam @julianguyen

PR #576, opened 2017-08-26
Fixing minor invalid ES6 syntax usage

baohouse:


baohouse:
Ugh, how do you squash all of those commits. I must not be syncing the fork correctly.

PR #578, opened 2017-08-26
Fix bug in medications#destroy and add controller tests

camillevilla:
## Summary
This PR fixes a bug in `medications#destroy` introduced by #573 and adds a few controller tests to `medications_controller_spec.rb`.

hi @HashNotAdam @julianguyen! Here's the requested changes. I'll open some tickets for the other suggested specs tomorrow.

### DIFF
--- a/spec/controllers/medications_controller_spec.rb
+++ b/spec/controllers/medications_controller_spec.rb
@@ -35,5 +35,28 @@ describe MedicationsController do
         expect(subject).to eq('<div class="small_margin_top"><i class="fa fa-bell small_margin_right"></i>Refill reminder email, Daily reminder email</div>')
       end
     end
+
+    describe 'DELETE #destroy' do
+      let(:user) { create(:user) }
+      let!(:medication) { create(:medication, user: user) }
+
+      context 'when the user is logged in' do
+        include_context :logged_in_user
+        it 'deletes the medication' do
+          expect { delete :destroy, params: { id: medication.id } }
+            .to change(Medication, :count).by(-1)
+        end
+
+        it 'redirects to the medications index page' do
+          delete :destroy, params: { id: medication.id }
+          expect(response).to redirect_to medications_path
+        end
+      end
+
+      context 'when the user is not logged in' do
+        before { delete :destroy, params: { id: medication.id } }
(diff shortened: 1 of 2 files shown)


spec/controllers/medications_controller_spec.rb
```diff
@@ -35,5 +35,23 @@
         expect(subject).to eq('<div class="small_margin_top"><i class="fa fa-bell small_margin_right"></i>Refill reminder email, Daily reminder email</div>')
       end
     end
+
+    describe 'DELETE #destroy' do
+      let(:user) { FactoryGirl.create(:user1) }
```
HashNotAdam: Since only 1 user definition is used across the whole spec, I think it would be nice if every test used this definition placed directly under `describe MedicationsController`.

It really doesn't matter but it's also worth noting that there is a factory simply called "user" and you don't need to use the FactoryGirl class.
```
let(:user) { create(:user) }
```

spec/controllers/medications_controller_spec.rb
```diff
@@ -35,5 +35,23 @@
...
+      let!(:medication) { FactoryGirl.create(:medication, userid: user.id) }
```
HashNotAdam: This is great but you can save some characters with:
```
let!(:medication) { create(:medication, user: user) }
```

spec/controllers/medications_controller_spec.rb
```diff
@@ -35,5 +35,23 @@
...
+
+      context 'when the user is logged in' do
```
HashNotAdam: This is sensational but would you mind also testing what happens when the user isn't logged in?

spec/controllers/medications_controller_spec.rb
camillevilla: @HashNotAdam thanks for the tip! Glad to save some characters.

spec/controllers/medications_controller_spec.rb
camillevilla: Alrighty! I added a test similar to the this [category#destroy test](https://github.com/julianguyen/ifme/blob/master/spec/controllers/categories_controller_spec.rb#L185).

PR #582, opened 2017-08-28
Use Polymorphic Relationships for Comments

slauppy:
# Summary 

This addresses https://github.com/julianguyen/ifme/issues/547 by

- Renaming columns in the comments table:
  - comment_type -> commentable_type
  - commented_on -> commentable_id
- Adding a polymorphic association from comments to:
  - moments
  - strategies
  - meetings

# Notes

I would love help testing this change, since it touches so much of the system!


HashNotAdam:
This approach is exactly what I had in mind when I logged the issue; thank you.
I'm going to be unable to get too deep into the project for a few days, however, so I'm afraid I can't help test right now. If this is still open later in the week, however, I'd love to help.

slauppy:
Thanks @HashNotAdam ! 

So far, the only possible bug I have found is that when commenting on your own moment and setting the visibility to something other than "Share with everyone," the comment is posted but it tries to send out a notification and triggers this:

```
NoMethodError (undefined method `exists?' for #<User:0x007fdad5780228>):

app/models/comment.rb:72:in `send_notification!'
app/models/comment.rb:51:in `notify_of_creation!'
app/controllers/application_controller.rb:298:in `comment_for'
app/controllers/moments_controller.rb:21:in `comment'
app/controllers/application_controller.rb:31:in `block in with_timezone'
app/controllers/application_controller.rb:31:in `with_timezone'
```

I cannot replicate this behavior on master.

HashNotAdam:
@slauppy that's my error

The line reads `return unless User.find_by(id: user_id).exists?` but, if the user doesn't exist, that becomes `return unless nil.exists?`. Could be fixed by replacing that with `return if User.find_by(id: user_id).nil?`

julianguyen:
Any updates @slauppy?

slauppy:
@julianguyen yes! @HashNotAdam's fix worked (plus another minor fix). 😄

Users should actually get notifications for private moment comments once this is merged in, so I hope that's not a problem. I'm not sure how long this has been broken on master.

HashNotAdam:
I'll have a lot more time this week and look forward to sitting down for another review of this; it's a really helpful feature.

slauppy:
My pleasure!

PR #587, opened 2017-08-31
Add to google cal

### LINKED ISSUE #517: add_to_google_cal field needs to be added to medication_model
Currently, add_to_google_cal is simply an attr_accessor on the medication model. Because of this whether a calendar reminder was added is not saved to the model, and it also makes doing comparisons on this field difficult.

abrophy:
resolves #517

### DIFF
--- a/app/helpers/calendar_helper.rb
+++ b/app/helpers/calendar_helper.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module CalendarHelper
+  def new_cal_refill_reminder_needed?(medication)
+    if medication.add_to_google_cal &&
+       medication.refill &&
+       (medication.add_to_google_cal_changed? || medication.refill_changed?)
+      true
+    else
+      false
+    end
+  end
+
+  def new_cal_refill_reminder_unneeded?(medication)
+    if medication.add_to_google_cal &&
+       medication.refill &&
+       (medication.add_to_google_cal_changed? || medication.refill_changed?)
+      false
+    else
+      true
+    end
+  end
+end
--- a/app/helpers/calendar_helper.rb
+++ b/app/helpers/calendar_helper.rb
@@ -10,14 +10,4 @@ module CalendarHelper
       false
     end
   end
-
-  def new_cal_refill_reminder_unneeded?(medication)
-    if medication.add_to_google_cal &&
-       medication.refill &&
-       (medication.add_to_google_cal_changed? || medication.refill_changed?)
-      false
-    else
-      true
-    end
-  end
 end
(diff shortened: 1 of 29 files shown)


db/migrate/20170830075513_add_add_to_google_cal_to_medication.rb
```diff
@@ -0,0 +1,5 @@
+class AddAddToGoogleCalToMedication < ActiveRecord::Migration[5.0]
+  def change
+    add_column :medications, :add_to_google_cal, :boolean
```
alexiamcdonald: Perhaps add default to boolean?

abrophy: sure, right on, better to be explicit

alexiamcdonald:
This PR is having some changes done. Working with @abrophy to make it better. Thought we should let everyone else know what's happening. 😄

app/models/moment.rb
```diff
@@ -17,6 +17,7 @@
 #  comment    :boolean
 #  strategy   :text
 #  slug       :string
+#  sentiment  :float
```
julianguyen: I think this was accidentally committed

spec/models/moment_spec.rb
```diff
@@ -15,6 +15,7 @@
 #  comment    :boolean
 #  strategy   :text
 #  slug       :string
+#  sentiment  :float
```
julianguyen: I think this was accidentally committed too.

app/helpers/calendar_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+       medication.refill &&
+       (medication.add_to_google_cal_changed? || medication.refill_changed?)
+      true
+    else
+      false
+    end
+  end
+
+  def new_cal_refill_reminder_unneeded?(medication)
+    if medication.add_to_google_cal &&
```
julianguyen: Could replace the contents of this method with just `!new_cal_refill_reminder_needed?(medication)`

app/helpers/medication_refill_helper.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 module MedicationRefillHelper
+  include CalendarHelper
   # Save refill date to Google calendar
   def save_refill_to_google_calendar(medication)
     return true unless current_user.google_oauth2_enabled? &&
-                       medication.add_to_google_cal == '1'
+                       new_cal_refill_reminder_needed?(medication)
```
julianguyen: 🎉  yassss

PR #600, opened 2017-09-06
Updates Calendar Uploader to use CalendarV3 ruby client

### LINKED ISSUE #586: Google Calendar Reminders not working
While working on on #517 I ran into issues where the ruby google client isn't being instantiated correctly.

API changes in the newer version of the ruby client are to blame. Because of upgrade to newer version of rails we're locked into having to update the code to the newer way of doing things.

abrophy:
fixes #586


alexiamcdonald:
LGTM

PR #603, opened 2017-09-07
Always redirect //if-me.org to //www.if-me.org

HashNotAdam:
Let me know if the decision is changed and, instead, we want to go without www.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -2,5 +2,11 @@
 
 # This file is used by Rack-based servers to start the application.
 
+require 'rack-rewrite'
+
+use Rack::Rewrite do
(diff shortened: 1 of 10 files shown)


config.ru
```diff
@@ -2,5 +2,11 @@
 
 # This file is used by Rack-based servers to start the application.
 
+require 'rack-rewrite'
+
+use Rack::Rewrite do
+  r301(/.*/, 'https://www.if-me.org$&', host: 'if-me-new.org')
```
baohouse: if-me-new?

HashNotAdam: LOL
Sorry... obviously I didn't clean up properly. That whole discussion was taking _way_ too long given I'm at work

PR #605, opened 2017-09-07
README.md: added link to if-me.org with

andy5995:
prominent note to readers that installation
and configuration isn't required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 # if me
 
-if me is a community for mental health experiences that encourages people to
+[_if me_](https://www.if-me.org/) is a community for mental health experiences that encourages people to
 share their personal stories with trusted allies. Trusted allies are the people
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,7 @@
 
 # if me
 
-if me is a community for mental health experiences that encourages people to
+[_if me_](https://www.if-me.org/) is a community for mental health experiences that encourages people to
```
HashNotAdam: Please keep lines to 80 characters. In markdown, some exceptions are allowed for things like URLs or where a line break could cause difficulties with the flow of reading, however, most of the time you should introduce a new line.

andy5995: Done!

andy5995:
Hmmm... now that I think of it, I should forget about using italics. Agreed?

andy5995:
Thanks for reviewing this @julianguyen and @HashNotAdam

PR #610, opened 2017-09-08
Set expires headers on assets

HashNotAdam:
Since Heroku connects requests directly to the Ruby web server, managing the
response headers used on assets (CSS, JS, images, etc.) needs to be handled by
Rails. By setting an expiry in the future, the web browser is directed to cache
the assets and reload from memory instead of hitting the server again for a new
copy of the same file

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -90,6 +90,10 @@ Rails.application.configure do
   end
 
   config.public_file_server.enabled = true
+  config.public_file_server.headers = {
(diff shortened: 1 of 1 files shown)


baohouse:
Initially worried about cache busting, but realized we hash the CSS/JS filenames, so it looks good!

HashNotAdam:
@baohouse welcome to the Rails Asset Pipeline 😆

PR #616, opened 2017-09-10
Adding CSS Modules

baohouse:


baohouse:
Still working on this; it's breaking flow checks.

client/app/bundles/shared/startup/registration.jsx
```diff
@@ -0,0 +1,9 @@
+// @flow
+import ReactOnRails from 'react-on-rails';
+
+import Logo from '../components/Logo';
+
+// This is how react_on_rails can see the Components in the browser.
+ReactOnRails.register({
+  Logo,
```
julianguyen: Can get rid of trailing comma here

client/webpack.config.js
```diff
@@ -43,8 +47,24 @@ const config = {
       NODE_ENV: 'development', // use 'development' unless process.env.NODE_ENV is defined
       DEBUG: false,
     }),
+    new webpack.optimize.UglifyJsPlugin({
```
julianguyen: 🎉 🎉 🎉

client/app/bundles/shared/startup/registration.jsx
baohouse: I think Code Climate fails without the trailing comma, unless we put this on a single line, though I figure we will have additional components in the future. I do want to have a separate task of simplifying the folder hierarchy with component registration, but will need some time to think about how this can be achieved.

baohouse:
@julianguyen My commit dismissed your approval. :-(

PR #621, opened 2017-09-10
personalize subject field of invitation emails

andy5995:
Thanks to @ThePedroPaulino and @esBeee  for helping with this

### DIFF
--- a/config/locales/devise_invitable.ptbr.yml
+++ b/config/locales/devise_invitable.ptbr.yml
@@ -17,7 +17,7 @@ ptbr:
         submit_button: "Defina sua senha"
     mailer:
       invitation_instructions:
-        subject: "Você foi convidado para se juntar ao if me!"
+        subject: "{%name} convidou você para participar do if-me!"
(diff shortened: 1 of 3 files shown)


config/locales/devise_invitable.ptbr.yml
```diff
@@ -17,7 +17,7 @@ ptbr:
         submit_button: "Defina sua senha"
     mailer:
       invitation_instructions:
-        subject: "Você foi convidado para se juntar ao if me!"
+        subject: "{%name} convidou você para participar do if-me!"
```
julianguyen: The copy here should be `if me` instead of `if-me`. I know this is a source of constant confusion, especially given our domain name lol.

andy5995:
Fixed! Thanks for reviewing, Julia. :)

config/locales/devise_invitable.ptbr.yml
baohouse: Sounds like a need for branding guidelines

PR #623, opened 2017-09-12
Secret share expiry

SophieRoseMcDonald:


julianguyen:
Looks like you need to rebase from master to get the tests to pass!

app/models/moment.rb
```diff
@@ -46,15 +55,18 @@ def array_data
     self.strategy = strategy.collect(&:to_i) if strategy.is_a?(Array)
   end
 
-  def category_name
-    category.try(:name)
-  end
+  delegate :category_name, to: :category
```
julianguyen: Ooh I like this!

HashNotAdam: Does it work though 🤔 
Admittedly, I rarely use Rails 5 but I thought you need to supply `prefix: true` if the method name would be different on the object.
For instance:
```
delegate :name, to: :category, prefix: true
```

Am I wrong?

app/views/shared/_viewers_indicator.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="viewers_indicator">
   <% viewers = local_assigns[:data].viewers %>
-  <%= number_of_viewers(local_assigns[:data].userid, current_user.id, viewers) %>
+  <%= number_of_viewers(local_assigns[:data].userid, current_user.try(:id), viewers) %>
```
julianguyen: Nice!

config/locales/es.yml
```diff
@@ -245,6 +245,7 @@ es:
   edit_moment: "Editar Momento"
   tagged: "Etiquetar Momento"
   new: "Nuevo Momento"
+  secret_share: 'Compártelo en Secreto'
```
julianguyen: Looks like we're missing secret_share.create and secret_share.cancel. Ahh this is why your Circle build is failing.

config/locales/ptbr.yml
```diff
@@ -243,6 +243,7 @@ ptbr:
   edit_moment: 'Editar Momento'
   tagged: 'Momentos com Tags'
   new: 'Novo Momento'
+  secret_share: 'Compartilhar Segredo'
```
julianguyen: Likewise as above.

db/migrate/20170724205314_add_secret_shares_to_moments.rb
```diff
@@ -0,0 +1,7 @@
+class AddSecretSharesToMoments < ActiveRecord::Migration
+  def change
+    add_column :moments, :secret_share_identifier, :uuid
+    add_column :moments, :secret_share_expires_at, :datetime
+    add_index :moments, :secret_share_identifier, unique: true
```
julianguyen: Nice!

app/models/moment.rb
HashNotAdam: I'm actually confused by the original code.
Category, mood, and strategy are all arrays of integers; what is calling these methods and how are they in a position to return a name?

app/views/moments/show.html.erb
```diff
@@ -1,5 +1,14 @@
 <% title @moment.name %>
 <%= render :partial => '/moments/info', locals: { data: @moment } %>
+<% if Rails.configuration.secret_share_enabled && @moment.owned_by?(current_user) %>
```
julianguyen: We should specify here that the secret share URL expires in a day.

app/views/shared/_viewers_indicator.html.erb
HashNotAdam: Probably better as `current_user&.id` despite the ugly factor

app/controllers/secret_shares_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class SecretSharesController < ApplicationController
+  before_action :ensure_secret_share_enabled!
+  skip_before_action :if_not_signed_in, only: [:show]
+
+  def ensure_secret_share_enabled!
```
HashNotAdam: I think this should be a private method—it helps be clear about which methods are called by routes

julianguyen: Ooh great suggestion @HashNotAdam! We typically put private methods at the bottom of the controller!

app/models/moment.rb
```diff
@@ -38,6 +38,15 @@ class Moment < ApplicationRecord
   validates :userid, :name, :why, presence: true
   validates :why, length: { minimum: 1, maximum: 2000 }
   validates :fix, length: { maximum: 2000 }
+  validates :secret_share_expires_at,
+            presence: true, if: :secret_share_identifier?
+
+  scope :find_secret_share!, lambda { |identifier|
+    find_by!(
+      'secret_share_expires_at > NOW()',
```
julianguyen: Is NOW() equivalent to Time.now?

spec/models/moment_spec.rb
```diff
@@ -18,24 +18,44 @@
...
-			new_category = create(:category, userid: new_user.id)
-			new_mood = create(:mood, userid: new_user.id)
-			new_moment = create(:moment, userid: new_user.id, category: Array.new(1, new_category.id), mood: Array.new(1, new_mood.id))
-			expect(new_moment.viewers).to be_empty
-		  end
+
+	describe '.find_secret_share!' do
+
+		context 'when a valid secret share moment exists' do
+			let(:moment) { create(:moment, :with_user, :with_secret_share) }
```
HashNotAdam: TABS!
😱 😱 😱

spec/models/moment_spec.rb
```diff
@@ -18,24 +18,44 @@
 #
 
 describe Moment do
```
HashNotAdam: TABS!
😱 😱 😱 

Looks like someone had dropped tabs into this file so Atom has assumed that was the desired outcome and formatted all your code this way.

app/models/moment.rb
HashNotAdam: Welcome to SQL

HashNotAdam: Technically, this could be done with Arel but I thought it was a little early to scare them with that.

Would be something like:
```
find_by!(
  arel_table[: secret_share_expires_at].gt(Time.zone.now),
  secret_share_identifier: identifier
)
```

app/models/moment.rb
julianguyen: Oh so this was being used with autocomplete in the Moments and Strategy editors.

`app/views/shared/_autocomplete_checkboxes.html.erb`

```
<%= f.text_field "#{data_type}_name", data: { autocomplete_source: all_data.map(&:name), autocomplete_source_ids: all_data.map(&:id) }, class: 'full_field smaller_margin_top smaller_margin_bottom', placeholder: t('common.form.press_enter') %>
```

app/models/moment.rb
julianguyen: Ah gotcha, thanks for explaining! I haven't touched pure SQL in a while :P
What's the advantage of using Arel @HashNotAdam (other than not having to write SQL)?

config/environments/test.rb
```diff
@@ -41,4 +41,7 @@
 
   # Raises error for missing translations
   config.action_view.raise_on_missing_translations = true
+
+  # Enables secret share feature flag
+  config.secret_share_enabled = true
```
julianguyen: I think it would be worth mentioning in our documentation that we have this `secret_share_enabled` feature flag!

app/models/moment.rb
HashNotAdam: Here... my (ever humble) opinion is nothing.

It can be used programmatically, though, so if there was a situation where you wanted to pass around objects like you do with standard ActiveRecord elements that might take variables, you can have different methods tagging on pieces of your query without worrying about the order you are concatenating your string or replacing variables in a string.

It also means you don't need to worry about DB specific syntax. `NOW()` is extremely common but it is _really_ common for syntax between DBs to be different and, like with AR, using Arel means you can change from, say, PostgreSQL to MySQL without rewriting code.

app/models/moment.rb
HashNotAdam: Does this mean that somehow the array of integers is being converted into an array of objects 😨

spec/models/moment_spec.rb
```diff
@@ -18,24 +18,44 @@
...
-		it "create public moment" do
-			new_user = create(:user1)
-			new_user2 = create(:user2)
-			new_allies = create(:allyships_accepted, user_id: new_user.id, ally_id: new_user2.id)
-				new_category = create(:category, userid: new_user.id)
-			new_mood = create(:mood, userid: new_user.id)
-			new_moment = create(:moment, userid: new_user.id, category: Array.new(1, new_category.id), mood: Array.new(1, new_mood.id), viewers: [new_user2.id])
-			expect(new_moment.viewers.count).to eq(1)
+
+
```
HashNotAdam: No need to be greedy, one line break is enough.

julianguyen: Omg good catch. If you're using Atom or any other IDE, you can remove trailing whitespace through your settings. It does it automagically for you! 🌠

app/models/moment.rb
julianguyen: Yeah... :/ This is my doing lol. I used jquery autocomplete rather than writing my own implementation.

config/locales/es.yml
SophieRoseMcDonald: Yes, we added secret_share keys that need to be filled in for each translation. Shall we call on the translators to address this and fill in the missing keys?

app/views/moments/show.html.erb
SophieRoseMcDonald: This feature has a default of one day, it will be changed in the next feature, so you can choose time/date of expiry. I can add a comment to the code for now if you like.

julianguyen: @SophieRoseMcDonald Ah that's fine then, no need for comment! Thanks!

config/locales/es.yml
SophieRoseMcDonald: I put some translated missing keys as place holders.

config/environments/test.rb
TimMoore: Hopefully the flag will be short-lived... it can be deleted once the feature is production ready.

config/environments/development.rb
```diff
@@ -74,4 +74,6 @@
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.force_ssl = false
+
+  config.secret_share_enabled = true
```
TimMoore: Maybe it should be disabled by default on `master` (so it doesn't disrupt or confuse other contributors) and only manually changed to `true` locally by people who are working on the feature.

You could comment out this line and add a comment saying to uncomment it if you are working on Secret Share Moments.

app/models/moment.rb
```diff
@@ -38,6 +38,15 @@ class Moment < ApplicationRecord
   validates :userid, :name, :why, presence: true
   validates :why, length: { minimum: 1, maximum: 2000 }
   validates :fix, length: { maximum: 2000 }
+  validates :secret_share_expires_at,
+            presence: true, if: :secret_share_identifier?
+
+  scope :find_secret_share!, lambda { |identifier|
```
TimMoore: Personally, I would prefer to use a class method here instead of a `scope`.

```ruby
  def self.find_secret_share!(identifier)
    find_by!(
      'secret_share_expires_at > NOW()',
      secret_share_identifier: identifier
    )
  end
```

I know this is something that is a little controversial in the Rails community, but here are the reasons I think a class method is better:

- `scope` is not familiar to all Ruby/Rails developers. It's ActiveRecord "magic" that people have to learn, along with the `lambda` syntax, which is also unfamiliar to many newer Ruby developers.
- Class methods are used in a lot of places in Ruby code, so they are more familiar to most Ruby developers, and even new Ruby developers who have not learned them yet will encounter them before too long.
- The [docs for `scope`](http://api.rubyonrails.org/v5.1/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-scope) say "this is simply 'syntactic sugar' for defining an actual class method". In this case, however, the `scope` syntax is not really any shorter or more readable than the equivalent class method, so it doesn't provide much benefit.
- The docs also say, "The method is intended to return an ActiveRecord::Relation object, which is composable with other scopes." In other words, part of the purpose of defining a `scope` is that you can chain more criteria on to the end of it, but that is not possible here, because this performs a `find_by!` that returns the actual `Moment`, not a relation that can be chained.

Here are some good blog posts that describe the intended use of `scope`, and when it is better to use a `scope` than a class method and vice-versa:

- http://blog.plataformatec.com.br/2013/02/active-record-scopes-vs-class-methods/
- https://www.justinweiss.com/articles/should-you-use-scopes-or-class-methods/

config/environments/development.rb
julianguyen: ^ I like this idea.

app/models/moment.rb
SophieRoseMcDonald: Julia are we making this a class method or keeping it as a scope?

HashNotAdam: Gotta love when your coaches throw you around with conflicting information 😆

HashNotAdam: @SophieRoseMcDonald while I could fight @TimMoore all day about the value of using scopes, his last point about it returning an `ActiveRecord::Relation` is completely correct and I accept it was my mistake to suggest it in this instance.

TimMoore: Sorry if I confused things! I'm not totally opposed to scopes... I think those articles make some very good arguments for using it in specific cases. Anyway, as long as we agree about it here, it's all good and fighting won't be necessary 😄

julianguyen: Such an intense debate lol. I did some bedtime reading on best practices for scopes vs. method (thanks @TimMoore!). Do we care about whether we return a null object vs. a blank relation? Scopes will return a blank relation in the event no record exists. With methods, we have to address returning nil. I think returning nil would be fine in this situation.

Tim raises really good points about readability though. We do want to be beginner friendly. What about this "compromise" that's mentioned in the last article Tim shared?
```
Even then, it’s still good to put your selecting, sorting, joining, and filtering code inside scopes. Then, use your scopes inside your class method. You’ll end up with a clearer class method, and scopes you can use throughout your app.
```
Anyways time for bed for me lol.

TimMoore: @julianguyen in this case, it will raise an error if the record doesn't exist, which is what we want here

PR #624, opened 2017-09-12
adding resources #486

andy5995:
I've listed each resource in the two commit messages, along with tags I've added.

I wasn't sure where to put the Mental Health humor link; for now, it's listed under _Communites_

### DIFF
--- a/doc/pages/communities.json
+++ b/doc/pages/communities.json
@@ -82,5 +82,17 @@
     "link":"http://emotionsanonymous.org/",
     "tags":["free", "support_groups"],
     "languages":["en"]
+  },
+  {
+    "name":"In the Wake of Suicide...trying to understand",
+    "link":"https://lensgirl53.wordpress.com/",
+    "tags":["blog", "trigger_warnings", "suicide_prevention"],
+    "languages":["en"]
+  },
+  {
+    "name":"Mental Health Humor & Cartoons - Chato B. Stewart",
+    "link":"https://blogs.psychcentral.com/humor/",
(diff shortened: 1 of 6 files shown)


doc/pages/services.json
```diff
@@ -98,7 +98,19 @@
   {
     "name":"US Department of Veteran's Affairs | Mental Health",
     "link":"https://www.mentalhealth.va.gov/gethelp.asp",
-    "tags":["free", "suicide prevention", "homeless", "directory", "veterans"],
+    "tags":["free", "suicide prevention", "homeless", "directory", "veterans", "government"],
     "languages":["en"]
+  },
+  {
+    "name":"Resource Links - Disabled Veterans National Foundation",
```
andy5995: Oops I forgot to add (US) to the end here. Should I correct that? And also the link above (but it starts with "US").

julianguyen: Maybe rephrase to "Disabled Veterans National Foundation Resources (US)"?

doc/pages/communities.json
```diff
@@ -82,5 +82,17 @@
     "link":"http://emotionsanonymous.org/",
     "tags":["free", "support_groups"],
     "languages":["en"]
+  },
+  {
+    "name":"In the Wake of Suicide...trying to understand",
```
julianguyen: I would put these two resources under Education, since Communities is more for resources that are support groups or meetups :) These are great!

andy5995:
Thanks @julianguyen 😺

PR #626, opened 2017-09-13
Fix foreman bootstrap script

jnamster:
Encountered error that it could not find webpack when starting the server. 
Change made in Procfile.dev so it installed dependencies before the build.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -141,9 +141,6 @@ ActiveRecord::Schema.define(version: 20170830075513) do
     t.boolean  "comment"
     t.text     "strategy"
     t.string   "slug"
-    t.uuid     "secret_share_identifier"
-    t.datetime "secret_share_expires_at"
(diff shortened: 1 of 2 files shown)


HashNotAdam:
This branch relies on the Secret Share branches so we need to approve those first

HashNotAdam:
@julianguyen there is a huge amount of complexity added by that React branch, have you thought any more about simplifying the system? Obviously, I suggested guard but the specific answer doesn't matter so much as making life easy for people.

baohouse:
Perhaps you can create a clean branch with just the `yarn install` amendment?

db/schema.rb
```diff
@@ -141,6 +141,9 @@
     t.boolean  "comment"
     t.text     "strategy"
     t.string   "slug"
+    t.uuid     "secret_share_identifier"
```
julianguyen: Looks like this file was accidentally committed?

Issue #627, opened 2017-09-13
Issue setting up development environment in local after pulling latest master


charlesskariah:
Rails server is running and when i tried to load the url in browser, I am getting the following error,

```
Error during failsafe response: 
        WebpackerLite can't find the manifest file: /home/charlesskariah/ifme/public/webpack/development/manifest.json
        Possible causes:
          1. You have not invoked webpack.
          2. You have misconfigured WebpackerLite's config/webpacker_lite.yml file.
          3. Your Webpack configuration is not creating a manifest.
```
This was working fine with my master before.This is happening after pulling the latest master.

andy5995:
@charlesskariah this issue was just reported in #setup on Slack. Did you receive your invite yet?

alexiamcdonald:
@charlesskariah I've run into this problem. I will investigate on my end. But if you figure out something in the mean time please let me know.

charlesskariah:
@andy5995 I just found that invitaion email.My bad
@alexiamcdonald Will let you know if i came across a solution

andy5995:
Ah... well, someone else has run into this problem too. It's being discussed in #setup.

baohouse:
#626 is addressing this. For the time being, from the project folder,

```
cd client && yarn install
```

alexiamcdonald:
Make sure you use the homebrew yarn and you have updated homebrew

alexiamcdonald:
Running `foreman start -f Procfile.dev` instead of `rails s` fixes this issue. I've updated the readme. I'm closing this issue now as I believe it's solved.

yasirazgar:
I have tried all the above steps
1. on running` yarn install and rails s`, i am getting
```
Error during failsafe response:
        WebpackerLite can't find the manifest file: /Users/yasirazgar/git/ifme/public/webpack/development/manifest.json
        Possible causes:
          1. You have not invoked webpack.
          2. You have misconfigured WebpackerLite's config/webpacker_lite.yml file.
          3. Your Webpack configuration is not creating a manifest.
```

2. on running `foreman start -f Procfile.dev`, I get,
```
12:52:16 web.1    | started with pid 86329
12:52:16 client.1 | started with pid 86330
12:52:16 client.1 | sh: bin/client.sh: /bin/sh^M: bad interpreter: No such file or directory
12:52:16 client.1 | exited with code 126
12:52:16 system   | sending SIGTERM to all processes
12:52:16 web.1    | terminated by SIGTERM
```

[...]

Issue #613, opened 2017-09-08
react_on_rails documentation needed


abrophy:
Documentation has to be fleshed for react_on_rails on how to set up locally and build components, README should also be updated accordingly

PR #631, opened 2017-09-14
 [#627] Updates readme for starting server

alexiamcdonald:
This fixes issue [#627](https://github.com/julianguyen/ifme/issues/627) where starting `rails s` would produce a console error.

Updated instructions to start server with `foreman start -f Procfile.dev`


baohouse:
A small fix that addresses #613

PR #641, opened 2017-09-16
Adding Cloudinary remote-image fetching and scaling

baohouse:
This technique has Cloudinary download (i.e. fetch) the image from if-me.org (without requiring us to upload the image to Cloudinary). We can get have this working solution until we get around to implementing Cloudinary uploading upon deployment.

The Cloudinary CDN has an amazing feature to detect Retina screens and to send back higher-quality images for such screens. On Retina screens, the total images footprint went from 4.4 MB down to 468 kB.


app/services/profile_picture.rb
```diff
@@ -1,30 +1,57 @@
...
+    def cloudinary(path)
+      path&.include?('.cloudinary.com')
+    end
+
+    def get_cloudinary_image_id(path)
+      path.split('/').last.split('.').first
+    end
+
+    def full_url(path)
+      "https://#{primary_domain}#{path}"
```
julianguyen: Where is primary_domain coming from? I don't see it passed as an argument. Also can't you get the url from `request.url`?

app/assets/stylesheets/application.scss
```diff
@@ -555,51 +555,44 @@ img {
 .mini_profile_picture {
   width: 100px;
   height: 100px;
-  border-radius: 150px;
```
julianguyen: Ooh much cleaner! 😍

julianguyen:
This is fantastic! :D Thanks for working on this! Still going through this PR and have a feq questions, but overall LGTM.

app/services/profile_picture.rb
baohouse: production.rb had it but, I thought when I initially tested it it was accessible; I guess not. I'm updating the development.rb file to look similar to production.rb (as @HashNotAdam has done), and changing the accessing of the config value to `Rails.application.config.action_controller.asset_host` (which seems to work).

baohouse:
Another key concept with this PR, I noticed that the server was validating that the image was available, or else send back a default avatar. That seemed inefficient. What I did instead was place the burden of verification onto the client browser (since the server being able to access the file might differ from the browser being able to access it).

PR #654, opened 2017-09-20
Adding Latin subset of Lato font

baohouse:
Homepage by default primarily uses Latin scripts for Lato, so we'll optimize loading Latin-based Lato; any non-Latin characters will appropriately load the full set.

I had to use FontForge and manually inspect the Unicode ranges that the Latin-optimized and full set versions of the font contained, so that's where those numbers in the ranges come from.

Initial load of fonts on the homepage went from 1020 kB down to 289 kB (75 kB actually belongs to FontAwesome).


app/assets/stylesheets/base/_fonts.scss
```diff
@@ -1,221 +1,477 @@
 // This file contains the base fonts for the application.
 
-$font-cdn-base: 'https://cdn.jsdelivr.net/font-lato/2.0/Lato';
+$full-set-cdn: 'https://cdn.jsdelivr.net/font-lato/2.0/Lato';
+$non-latin-range:
```
julianguyen: Oh snap, you manually figured this out? Wow 🥇

baohouse: In my own weird way, going through each character is like Tibetan monks making mandala sand art.

julianguyen: Do you anticipate ever having to update this list?

baohouse: By that time we will want to have automated deployment scripts to split the font by language. This was just a low-hanging fruit in that Lato officially released Latin vs full set versions of its font.

PR #655, opened 2017-09-23
Styling tweaks and fixing viewers bug

julianguyen:
Will merge this as soon as it goes green to fix the viewers bug. Will revisit at a later time to write tests for this.


app/controllers/application_controller.rb
```diff
@@ -370,6 +370,6 @@ def respond_with_json(reponse)
   end
 
   def hide_page?(subject)
-    subject.viewer?(current_user) && current_user.mutual_allies?(subject.user)
```
julianguyen: Not sure why this logic was changed to this... This is the opposite of what we want lol.

baohouse:
OMG, LOL, the SVG is really just a PNG inside an SVG container. D'oh! It's way bigger than the other image.

julianguyen:
@baohouse lol goddamn I was tryna get a PR out before going out

PR #657, opened 2017-09-24
More bug fixes woo

julianguyen:
In retrospect, I should have divided of these fixes into separate PRs :P

### Add profile URL to @page_author
- Was not being set (done during Rails 5 upgrade by accident)
- Link to Moment/Strategy author's profile when the post does not belong to the current user and both users are allies

### Fixing SVG issues
- Current welcome.svg contains a png wrapped inside (not performant), so it was replaced with a purely vector version.
- Added SVG to the list of gzip-compressible assets.

image

image

### Fixing comments bugs and adding tests
- Added tests for comment.rb and generate_comment in application_controller.rb
- Sometime soon I'd like to move all the client side code for comments and notifications to React :)

### Simplify notification_data keys, cleanup syntax
- Consolidated notification_data keys i.e. momentid/strategyid/meetingid, moment/strategy/meeting to use typeid and typename respectively
- Made Codeclimate happy :)
- Did some refactoring in application_controller.rb as well, including using `.find_each` instead of `.each` (it's more [efficient](https://apidock.com/rails/ActiveRecord/Batches/find_each)!)


julianguyen:
Also, what are people's thoughts on testing out private methods?

HashNotAdam:
@julianguyen while I've certainly done it plenty of times, in an ideal world, I prefer to structure tests around behaviour, in which case, the private methods would be tested as a consequence of testing the expected behaviour.
While it wouldn't take you long to find examples of where I've taken the exact opposite approach in a blind goal of increasing the coverage percentage, I have come to feel that percentage seen in isolation is a vanity metric that can mask bigger problems and lead you to write tests that simply replicate what the method does.

app/controllers/meetings_controller.rb
```diff
@@ -31,45 +31,9 @@ def show
   end
 
   def comment
```
HashNotAdam: 👍

app/helpers/notification_mailer_helper.rb
```diff
@@ -5,25 +5,25 @@ def comment_on_moment_subject(data)
...
       val('comment_on_strategy_subject'),
       user: data['user'].to_s,
-      strategy: data['strategy'].to_s
+      strategy: data['typename'].to_s
     )
   end
 
   def comment_on_meeting_subject(data)
-    groupid = Meeting.where(id: data['meetingid']).first.groupid
+    groupid = Meeting.where(id: data['typeid']).first.groupid
```
HashNotAdam: `Meeting.find(data['typeid']).groupid`

app/helpers/notification_mailer_helper.rb
```diff
@@ -93,7 +96,7 @@ def new_meeting_subject(data)
   end
 
   def meeting_body(data)
-    meeting = Meeting.where(id: data['meetingid']).first
+    meeting = Meeting.where(id: data['typeid']).first
```
HashNotAdam: `Meeting.find(data['typeid'])`

app/mailers/application_mailer.rb
```diff
@@ -38,11 +38,11 @@ def comment_notify(data, recipient)
     @message += comment_text(@data, key)
 
     if comment_on_moment(@data) || comment_on_moment_private(@data)
```
HashNotAdam: There is a lot of duplication in here.
What if there was a method that returned the class name?

```
url = send("#{commented_model(@data)}_url", @data['typeid'])
link = link_to(I18n.t('click_here'), url)
```

```
def commented_model(data)
  data['type'].match(/comment_on_([^_]+)/)[1]
end
```

app/models/comment.rb
```diff
@@ -48,54 +48,71 @@ def create_from!(params)
...
-    notifications = Notification.where(userid: user_id).order(:created_at)
+  def send_notification(data, notifications, user_id)
     Pusher["private-#{user_id}"]
       .trigger('new_notification', notifications: notifications)
-
     NotificationMailer.notification_email(user_id, data).deliver_now
   end
 
+  def send_notification!(creator, association, user_id)
+    return if User.find(user_id).nil?
```
HashNotAdam: `find` throws an exception if no record exists but you could use `find_by(id: user_id)`

app/models/comment.rb
```diff
@@ -48,54 +48,71 @@ def create_from!(params)
...
+
+  def type
+    "comment_on_#{commentable_type}#{visibility == 'private' ? '_private' : ''}"
+  end
+
+  def unique_id(type)
+    "#{type}_#{id}"
+  end
+
+  def notifications(data, userid)
```
HashNotAdam: This method has an important side effect—it should either be `notifications!` or the creation should be handled elsewhere.

app/models/comment.rb
```diff
@@ -48,54 +48,71 @@ def create_from!(params)
...
+    Notification.create!(userid: userid, uniqueid: unique_id(type), data: data)
+    model_data = Notification.where(userid: userid)
+    model_data.order('created_at ASC') if commentable_type == 'meeting'
+    model_data.order(:created_at)
```
HashNotAdam: These 2 lines do the same thing so the 3 lines could be combined into 1

app/models/comment.rb
```diff
@@ -48,54 +48,71 @@ def create_from!(params)
...
+  end
+
+  # Notify MeetingMembers except for commenter that there is a new comment
+  def handle_meeting(association, creator)
+    MeetingMember.where(meetingid: commentable_id).find_each do |member|
+      next if member.userid == creator.id
```
HashNotAdam: Why not filter this at the DB?
```
MeetingMember.where(meetingid: commentable_id)
             .where.not(userid: creator.id)
             .find_each do |member|
```

spec/controllers/application_controller_spec.rb
```diff
@@ -384,4 +384,202 @@
       expect(controller.moments_stats).to eq('<div class="center" id="stats">You have written a <strong>total</strong> of <strong>3</strong> moments. This <strong>month</strong> you wrote <strong>2</strong> moments.</div>')
     end
   end
+
+  describe 'generate_comment' do
```
HashNotAdam: It would be worth running Rubocop over this file and just ignoring the "block too long" warnings

spec/controllers/application_controller_spec.rb
```diff
@@ -384,4 +384,202 @@
...
+    let(:user1) { create(:user1) }
+    let(:user2) { create(:user2) }
+    let(:user3) { create(:user3) }
+    let(:default_profile_picture) { "<img src='' class='mini_profile_picture'\n        onerror=\"this.onerror=null;this.src='/assets/default_ifme_avatar.png'\" />" }
+    let(:comment) { 'Hello from the outside'}
+
+    def delete_comment(comment_id)
+      "<div class=\"table_cell delete_comment\"><a id=\"delete_comment_#{comment_id}\" class=\"delete_comment_button\" href=\"\"><i class=\"fa fa-times\"></i></a></div>"
```
HashNotAdam:
```
%(<div class="table_cell delete_comment"><a id="delete_comment_#{comment_id}" class="delete_comment_button" href=""><i class="fa fa-times"></i></a></div>)
```

spec/controllers/application_controller_spec.rb
```diff
@@ -384,4 +384,202 @@
...
+    end
+
+    def comment_info(user)
+      "<a href=\"/profile?uid=#{controller.get_uid(user.id)}\">#{user.name}</a> - less than a minute ago"
```
HashNotAdam:
```
%(<a href="/profile?uid=#{controller.get_uid(user.id)}">#{user.name}</a> - less than a minute ago)
```

spec/models/comment_spec.rb
```diff
@@ -14,23 +14,203 @@
...
-	  new_moment = create(:moment, userid: new_user.id, category: Array.new(1, new_category.id), mood: Array.new(1, new_mood.id))
-	  new_comment = build(:comment, commentable_id: new_moment.id, visibility: 'all')
-
-	  expect(new_comment).to have(1).error_on(:comment_by)
-	end
+  let(:user1) { create(:user1) }
+  let(:user2) { create(:user2) }
+  let(:user3) { create(:user3) }
+  let(:short_comment) { 'Hello from the outside' }
+  let(:long_comment) { 'Pretty as a picture, Sweeter than a swisher. Mad cause I\'m cuter than the girl that\'s witchya. I can talk about it cause I know that I\'m pretty' }
```
HashNotAdam:
```
"Pretty as a picture, Sweeter than a swisher. Mad cause I'm cuter than the girl that's witchya. I can talk about it cause I know that I'm pretty"
```

app/controllers/application_controller.rb
```diff
@@ -103,92 +103,46 @@ def get_uid(userid)
...
     data = []
-
-    userid =
-      if profile.blank?
-        current_user.id
-      else
-        profile
-      end
-
+    userid = profile.blank? ? current_user.id : profile
```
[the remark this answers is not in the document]
julianguyen: Oh yeah looks like a user id, I'll update the variable name to make that more clear.

app/controllers/application_controller.rb
```diff
@@ -103,92 +103,46 @@ def get_uid(userid)
...
+          end
         end
       end
-    elsif data_type == 'strategy'
-      Moment.where(userid: userid).all.each do |moment|
-        if moment.strategy.present? && !moment.strategy.empty? && (profile.blank? || (profile.present? && (current_user.id == profile || moment.viewers.include?(current_user.id))))
-          data += moment.strategy
+    elsif data_type == 'mood' || data_type == 'strategy'
+      moments.find_each do |m|
+        data_obj = data_type == 'mood' ? m.mood : m.strategy
```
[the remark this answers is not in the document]
julianguyen: omg good catch

app/controllers/application_controller.rb
```diff
@@ -216,12 +170,12 @@ def generate_comment(data, data_type)
                                            current_user)
     end
 
-    if (data_type == 'moment' && (Moment.where(id: data.commentable_id, userid: current_user.id).exists? || data.comment_by == current_user.id)) ||
-       (data_type == 'strategy' && (Strategy.where(id: data.commentable_id, userid: current_user.id).exists? || data.comment_by == current_user.id)) ||
-       (data_type == 'meeting' && (MeetingMember.where(meetingid: data.commentable_id, userid: current_user.id, leader: true).exists? || data.comment_by == current_user.id))
-
+    if comment_deletable?(data, data_type)
```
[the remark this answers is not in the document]
julianguyen: Hmm this didn't work for me, but I'll revisit this in another refactor PR. Thanks!

app/controllers/application_controller.rb
```diff
@@ -314,6 +268,52 @@ def moments_stats
 
   private
 
+  def data_included?(data_type, data_id, data)
```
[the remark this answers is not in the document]
julianguyen: Ooooh nice! Thanks!

app/models/comment.rb
julianguyen: Good point XD

spec/models/comment_spec.rb
julianguyen: Ahh but I like single quotes outside LOL

baohouse:
Somehow my PR into your branch got wiped. 🤔 I'll make another PR.

julianguyen:
@baohouse Ack that's my fault, I force pushed.

PR #672, opened 2017-09-28
Add instructions to debug with foreman

SophieRoseMcDonald:


README.md
```diff
@@ -484,6 +484,18 @@ rubocop
 
 You can read about Rubocop [here](http://rubocop.readthedocs.io/en/latest/).
 
+## Debugging with Foreman
+
+Edit `config/puma.rb` and add `worker_timeout 3600`
```
julianguyen: What do you think of editing `config/puma.rb` and adding `worker_timeout 3600`?

README.md
```diff
@@ -484,6 +484,18 @@ rubocop
...
+In one terminal tab, run
+ ```
+ bundle exec foreman start client -f Procfile.dev
```
julianguyen: Maybe say something like:

Run `bin/start_app` which runs `bundle exec foreman start client -f Procfile.dev`

README.md
HashNotAdam: What about just adding this to the project?
```
worker_timeout 3600 unless Rails.env.production?
```
I think getting people to change their config will just lead to someone accidentally committing it to the project and then the timeout protection wouldn't exist in production.

julianguyen: Oh true that's a better idea! Thanks!

julianguyen:
Just an update, we moved sections of our README into smaller docs. So you'll have to add your documentation for this in TESTING.md :D

julianguyen:
Going to takeover this PR here: https://github.com/julianguyen/ifme/pull/672 so that we can get this out ASAP. Thanks so much @SophieRoseMcDonald for your work on this <3

PR #677, opened 2017-09-30
Fix bug in shared form view (closes #625)

### LINKED ISSUE #625: Submitting Moment with blank description results in a server error.
Form submission is allowed when the description field is blank, but thought field is filled in.  It may be that form_processing.js is checking the wrong field.

jeffreyroy:


app/views/shared/_autocomplete_checkboxes.html.erb
```diff
@@ -19,8 +19,10 @@
     <% end %>
 
     <div id="<%= "#{data_type_plural}_list" %>">
-      <% all_data.each do |item| %>
-        <div id="<%= "#{data_type}_name_#{item.id}" %>" class="<%= this_data.include?(item.id) ? 'display_block' : 'display_none' %>"><%= f.check_box(data_type, { multiple: true, checked: this_data.include?(item.id) }, item.id, nil) %><span><%= item.name %></span><br></div>
+      <% if !all_data.nil? && all_data.length > 0 %>
```
julianguyen: You can just check for `!all_data.empty?` instead!

camillevilla: hi @jeffreyroy (nice to run into you here 👋)! 

(1) Could you add a link to the issue number in your PR description as well? It'll make it easier to cross-reference the issue and PR later on. 
(2) `all_data.present?` is a nice alternative as well. Then you won't have to use `!`. 
I did a little more research into `.empty?` vs. `.blank?` on the Rails API docs. It looks like `.empty?` throws an error or `nil values, whereas `.blank?` will fail silently.

jeffreyroy: Hey Camille! :smile: Long time no see.  I hope you're doing well.  And thanks to you and Julia for the feedback.

Re the code, I was just using the test the original author used.  ```present?``` is definitely more concise, and I'll make that change.

julianguyen: Oooh great suggestion @camillevilla :D

Issue #539, opened 2017-08-14
Destroy associations via ActiveRecord dependents


HashNotAdam:
There are a lot of related records which are removed in controllers in the `remove` methods. These should be handled by creating an ActiveRecord association in the model and then setting the dependent.

A good example is `AlliesController.remove` which destroys notifications, moments, and strategies before finally deleting the Allyship.

eiguike:
I'll look into this issue.

PR #683, opened 2017-10-02
Moving responsibility to remove all notifications when a allyship is deleted from AlliesController to AllyshipModel

eiguike:
This pull request is related to issue #539 and it is just partial refact.

I moved all actions, for removing Notifications when an Allyship is deleted, done before in AlliesControoler to AllyshipModel. I think this way will shrink some lines in AlliesController leaving it simpler.

I tried some other approach like creating dependents into models, but in this case, for Allyship and Notifications, there are no correlations to do that - unless the action is deleting a user.


spec/models/allyship_spec.rb
```diff
@@ -49,4 +49,60 @@
 	  new_allies = build(:allyships_accepted, user_id: new_user1.id, ally_id: nil)
 	  expect(new_allies).to have(1).error_on(:ally_id)
 	end
+
+  context 'when destroying' do
```
julianguyen: Looks like we should also test that the ally is being removed from viewers lists too.

app/models/allyship.rb
```diff
@@ -57,4 +58,21 @@ def different_users
...
     errors.add(:ally_id, 'ally_id is nil') if ally_id.nil?
   end
+
+  private
+
+  def remove_activities_between_users
+    user_id = self.user_id
+    ally_id = self.ally_id
+
+    Notification.for_ally(user_id, ally_id).or(
```
julianguyen: We could move this into its own helper

app/models/allyship.rb
```diff
@@ -57,4 +58,21 @@ def different_users
...
+      Notification.for_ally(ally_id, user_id)
+    ).destroy_all
+
+    # Remove ally from all viewers lists
+    [Moment, Strategy].each do |viewed_class|
```
julianguyen: We could move this into its own helper.

app/models/allyship.rb
eiguike: Do you think moving this part to concen would be more appropriate?

julianguyen: We could move it into a concern, but will it be reused again? If not I think it's fine just sticking that logic for destroying Notifications into a private method in the model (same goes for removing allies from viewers lists).

julianguyen: I was thinking you should update it to this:

```
private

def remove_ally_notifications
   user_id = self.user_id
   ally_id = self.ally_id
   Notification.for_ally(user_id, ally_id).or(
      Notification.for_ally(ally_id, user_id)
    ).destroy_all
end

def remove_ally_viewers
   user_id = self.user_id
   ally_id = self.ally_id
   [Moment, Strategy].each do |viewed_class|
      viewed_class.destroy_viewer(user_id, ally_id)
      viewed_class.destroy_viewer(ally_id, user_id)
    end
end

def remove_activities_between_users
    remove_ally_notifications
    remove_ally_viewers
  end
```

spec/models/allyship_spec.rb
```diff
@@ -49,4 +49,83 @@
 	  new_allies = build(:allyships_accepted, user_id: new_user1.id, ally_id: nil)
 	  expect(new_allies).to have(1).error_on(:ally_id)
 	end
+  context 'when destroying' do
```
julianguyen: Looks like we're testing `remove_ally_notifications` which is good, but we should add tests for `remove_ally_viewers` :D

app/models/allyship.rb
eiguike: Thank you, I'll update it right now!

spec/models/allyship_spec.rb
eiguike: I forgot to add this test, thank you for catching this!

julianguyen:
Don't forget to add yourself to our Contributors page :D

PR #693, opened 2017-10-07
issue 668 navigation to "help" may need adjustment

zawmoelwin:
Take option 2 as solution. Rename the "Contributors" page to "Contribute"
change route contributors to contribute but keep the contributors_path, change the locale en "navigaiton.contributors"  to "Contribute" in footer section


.env
```diff
@@ -0,0 +1 @@
+
```
julianguyen: Please remove this file from the PR :)

config/locales/en.yml
```diff
@@ -23,7 +23,7 @@ en:
  navigation:
    about: 'About'
    blog: 'Blog'
-   contributors: 'Contributors'
+   contributors: 'Contribute'
```
julianguyen: If we're going to take this "Contribute" route, let's also rename the key to "contribute"

config/routes.rb
```diff
@@ -81,8 +81,8 @@
 
   get 'pages/home'
   match 'about', to: 'pages#about', via: :get
-  match 'contributors', to: 'pages#contributors', via: :get
-  match 'partners', to: 'pages#partners', via: :get
+  match 'contribute', to: 'pages#contributors', via: :get , as: 'contributors'
```
julianguyen: We might as well rename the page#contributors to Contribute

db/schema.rb
```diff
@@ -234,8 +234,8 @@
     t.datetime "invitation_sent_at"
     t.datetime "invitation_accepted_at"
     t.integer  "invitation_limit"
-    t.integer  "invited_by_id"
     t.string   "invited_by_type"
+    t.integer  "invited_by_id"
```
julianguyen: Please remove this file from the PR.

julianguyen: I can't comment on the package-lock.json file for some reason, but please also remove that file too.

config/locales/en.yml
julianguyen: We'll also need to translate this in all of our locales as well.

db/schema.rb
zawmoelwin: noted.

config/locales/en.yml
zawmoelwin: I will edit the key in all locales but we need to open another issue for translation.

zawmoelwin:
@julianguyen 
I can't remove the package.json as it is the config file for yarn install. Thank you for the warm welcome. In this update,  
-Remove .env
-Remove schema.rb
-Change the key navigation:contributors to navigation:contribute
-Change the page.contributors key to page.contribute
-Change the routes from contributors to contribute.
-Change rspec  test files from get: contributors to get: contribute.

camillevilla:
@julianguyen: just a heads up, @zawmoelwin and I are going to do some pair programming tomorrow (10/7) to rebase and clean up commit messages

Issue #357, opened 2016-11-08
Add more tests to meetings_controller_spec.rb


julianguyen:
Add more tests for this controller!

cpanato:
I have some questions about this controller, how/who can help me?

julianguyen:
@cpanato Can you email us at join.ifme@gmail.com so we can add you to our Slack? I would answer most of the questions :P

rhiroyuki:
Hi I would like to help solve this issue... is the email above still valid? What is the process to submit a pull request in this project?

andy5995:
Hi! Yes, that email is still valid.

PR #699, opened 2017-10-08
Add partially meeting controller spec tests

rhiroyuki:
This PR partially solves #357.

Covered methods:
```
MeetingsController:
GET #show
GET #delete_comment
GET #new
POST #comment
GET #edit
```

### DIFF
--- a/app/controllers/meetings_controller.rb
+++ b/app/controllers/meetings_controller.rb
@@ -57,7 +57,7 @@ class MeetingsController < ApplicationController
       Notification.where(uniqueid: public_uniqueid).destroy_all
     end
 
-    render nothing: true
+    head :ok
   end
 
   # GET /meetings/new
--- a/spec/controllers/meetings_controller_spec.rb
+++ b/spec/controllers/meetings_controller_spec.rb
@@ -179,8 +179,10 @@ RSpec.describe MeetingsController, :type => :controller do
   end
 
   describe 'GET #new' do
-    let(:user) { create(:user) }
-    let(:group_member) { create(:group_member, user: user, leader: true) }
+    let!(:user) { create(:user) }
+    let!(:group_member) do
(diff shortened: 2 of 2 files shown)


app/controllers/meetings_controller.rb
```diff
@@ -57,7 +57,7 @@ def delete_comment
       Notification.where(uniqueid: public_uniqueid).destroy_all
     end
 
-    render nothing: true
+    head :ok
```
julianguyen: Could you explain why you changed this and if this behaviour is equivalent to what was previously there?

spec/controllers/meetings_controller_spec.rb
```diff
@@ -12,5 +12,240 @@
...
+          before do
+            meeting_member.destroy
+            get :show, params: { id: meeting.id }
+          end
+
+          it { expect(response).to have_http_status(:redirect) }
+          it { expect(response).to redirect_to(group_path(meeting.groupid)) }
+        end
+
+        context 'inexistent meeting id' do
```
julianguyen: replace inexistent with non-existent :P

app/controllers/meetings_controller.rb
rhiroyuki: When I wrote the tests for delete_comment I got this warning.
```
 .DEPRECATION WARNING: `:nothing` option is deprecated and will be removed in Rails 5.1. Use `head` method to respond with empty response body. (called from delete_comment at ...ifme/app/controllers/meetings_controller.rb:60)
```

Searching for solutions I found this answer explaining to use 'head :ok'. 
https://stackoverflow.com/questions/34688726/the-nothing-option-is-deprecated-and-will-be-removed-in-rails-5-1

julianguyen: Oooh awesome catch/find!

Issue #665, opened 2017-09-26
Translation bug for Portuguese


SophieRoseMcDonald:
Translation bug for Portuguese 
```ActionView::Template::Error (translation missing: ptbr.date.formats.long):
1: <%= ::TimeAgo.created_or_edited(local_assigns[:data]) %>
2: <% if local_assigns[:data].category.count > 0 %>
3:   <br>
4:   <strong>

app/services/time_ago.rb:11:in `formatted_ago'
app/services/time_ago.rb:18:in `created_or_edited'
app/views/moments/_info.html.erb:1:in _app_views_moments__info_html_erb__673145105344836068_70307964841260'```

localhostpedro:
Checking this one....

julianguyen:
@localhostpedro Any updates on this?

localhostpedro:
@julianguyen yes I am : Changing my computers and reorganising my projects.  Be ready for tomorrow, top.

zawmoelwin:
Can anyone elaborate more on how to replicate the issue?

After trying to replicate the issue(cannot replicate the issue), but  found another bug , language locale translation missing. 
"translation missing: ptbr.datetime.distance_in_words.less_than_x_minutes"

How to recreate: Sign in. Change the app language to Portuguese. Create new moment by clicking on + sign novo memento. Fill the form and create the moment.  The created moment shows the message.

PR #715, opened 2017-10-14
This pull request try to fix translation bug,  please see the discussion #665

zawmoelwin:
add datetime keys for yml files to address the bug show after moment create :locale translation missing.
"translation missing: ptbr.datetime.distance_in_words.less_than_x_minutes"
This might be partial solution, as cannot replicate the issue #665 .

used the translation from the following open source and might need another eye 
https://github.com/svenfuchs/rails-i18n/tree/master/rails/locale

Please see the discussion #665 
@camillevilla  @julianguyen


baohouse:
Tests might be failing because the datetime tree probably needs to be nested under "devise"

PR #716, opened 2017-10-14
Fix Validations for Medication Create

thomasoboyle:
### Fixes a part of: https://github.com/julianguyen/ifme/issues/682
More fixes here! https://github.com/julianguyen/ifme/pull/721

### Change Log

- The else case the Medication Create action triggered when fields are not filled out (code snippets below) seems to reload the page, wiping out `medication.errors.messages`. Moving the medication validations into partial to avoids missing them because of page reload.  Validations now work properly for both create and edit. 

```
def create
    @medication = Medication.new(medication_params)
    return unless save_refill_to_google_calendar(@medication)

    if @medication.save
    else
      render_unprocessable_medication
    end
  end
```
```
...
private

  def render_unprocessable_medication
    respond_to do |format|
      format.html { render :new }
      format.json do
        render(json: @medication.errors,
               status: :unprocessable_entity)
      end
    end
  end
```

### DIFF
--- a/app/views/medications/_form.html.erb
+++ b/app/views/medications/_form.html.erb
@@ -1,10 +1,5 @@
 <div id="medication_content">
   <%= form_for(@medication) do |f| %>
-  <% if @medication.errors.any? %>
-    <div class="error_explanation">
-      <%= t('common.form.error_explanation') %>
-    </div>
-  <% end %>
 
   <div class="table">
 
--- a/app/views/medications/edit.html.erb
+++ b/app/views/medications/edit.html.erb
@@ -1,4 +1,4 @@
 <% title "#{t('common.actions.edit_instance', instance: @medication.name)}" %>
 
-<%= render 'validation_errors.html.erb' %>
+<%= render 'validation_errors' %>
 <%= render 'form' %>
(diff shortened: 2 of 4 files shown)


julianguyen:
Will get around to reviewing this, this week! :D Thanks so much <3

app/views/medications/edit.html.erb
```diff
@@ -1,2 +1,4 @@
 <% title "#{t('common.actions.edit_instance', instance: @medication.name)}" %>
+
+<%= render 'validation_errors.html.erb' %>
```
julianguyen: Can this just be `<%= render 'validation_errors' %>`?

app/views/medications/_form.html.erb
```diff
@@ -1,10 +1,5 @@
 <div id="medication_content">
   <%= form_for(@medication) do |f| %>
-  <% if @medication.errors.any? %>
```
julianguyen: Also to avoid duplication, should we just include `<%= render 'validation_errors' %>` here instead of in `app/views/medications/edit.html.erb` and `app/views/medications/new.html.erb`?

app/views/medications/edit.html.erb
thomasoboyle: It most certainly! Good catch!

app/views/medications/_form.html.erb
thomasoboyle: Hi Julia! I thanks for the code review and comments!

I tried this and unfortunately it does not work in this case. `@medication` in line ones does not have any errors because of the else block of the create action in the medications controller.

PR #718, opened 2017-10-15
This pull request try to fix translation bug, please see the discussion #665

zawmoelwin:
add datetime keys for yml files to address the bug show after moment create :locale translation missing.

"translation missing: ptbr.datetime.distance_in_words.less_than_x_minutes"
 This pull request fixes the  #665 

used the translation from the following open source and might need another eye
https://github.com/svenfuchs/rails-i18n/tree/master/rails/locale

This pull request might fail a bunch of tests due to difference in record create time error.
Please advise on how it should be handled. 
@julianguyen  @baohouse  @camillevilla


baohouse:
I looked at the test logs... it's failing because of the difference in expected vs actual values; we're expecting "less than a minute ago" but it's rendering as "1 minute ago", which means there's a race condition. This is tricky; this means the test is time dependent. We'll need someone with experience in writing automation tests to guide us with best practice regarding mocking up times. I've little experience in this area, but when I've done similar tests, we architected the platform so that we could control when "now" is.

zawmoelwin:
I am working on a solution to this but not sure if it is a good idea. We can use the created_at or edited_at data in creating the test-against object. or We can create a helper function for test that ignore the created_at or edited_at in compare.

timofonic:
@baohouse @zawmoelwin @camillevilla @julianguyen @haleyelder @andy5995 Any news about this?

PR #720, opened 2017-10-15
Google calendar for Meetings

cesargroening:
Pull request for add google calendar functionality to "meetings" for google authenticated users.

*Now, the meeting shows an option to "add event to google calendar" if have not added before.
 *And show "Delete from google calendar" if the event already exists and it's have not cancelled


andy5995:
Hi @cesargroening 

I noticed codeclimate is reporting errors. Is this a wip?

timofonic:
@cesargroening @julianguyen @andy5995 @haleyelder Any news about this? It seems a very interesting feature. I would like to be able to support more platforms outside Googlesphere, but this is a very good start! This software has potential to evolve into more scopes :D

PR #721, opened 2017-10-16
Validation fixes for Moments and Strategies

thomasoboyle:
### Change Log
- Move validations from Moments & Strategies forms into their own partials. As stated in https://github.com/julianguyen/ifme/pull/716,
moving the validations into partial triggers validation messages prior to page reload. Validations now work properly for both create and edit for Moments & Strategies.
- Add myself as a contributor ;)


app/views/moments/_form.html.erb
```diff
@@ -1,11 +1,6 @@
 <%= render partial: 'shared/editor_modal' %>
 
 <%= form_for(@moment) do |f| %>
-<% if @moment.errors.any? %>
```
julianguyen: Similar comment made in the other PR, can't we put `<%= render 'validation_errors' %>` here?

app/views/strategies/_form.html.erb
```diff
@@ -1,11 +1,6 @@
 <%= render partial: 'shared/editor_modal' %>
 
 <%= form_for(@strategy) do |f| %>
-<% if @strategy.errors.any? %>
```
julianguyen: Put `<%= render 'validation_errors' %>` here instead?

julianguyen:
You should also rebase from master! :)

thomasoboyle:
Yay! Thank you!

Issue #723, opened 2017-10-16
Tests failing due to assertions following page reloads


baohouse:
"Persisting browser locale after sign in When user sign in, sign out, then change local" in `persisting_browser_locale_after_sign_in_spec.rb` is failing. I suspect it is because the page loads asynchronously, but we don't have a wait_until the page finishes loading before we assert the page title is there.

Why it worked before might be because the page update was happening fast enough, but perhaps now it isn't. All a theory of course.

baohouse:
Reading this resource: https://aranair.github.io/posts/2016/07/27/capybara-and-waiting/

baohouse:
Looks like my initial analysis was wrong, sorry. Here's the test result:

> expected to find text "Profundiza en tus Momentos - eventos y situaciones que afectan tu salud mental" in "if me Página principal Momentos Estrategias Medicamento Grupos Aliados ¡Adelante! ¡Hola Some Person96! Aquí te decimos cómo empezar a compartir tus historias. Cuéntanos sobre ti. Personaliza tu Perfil - agrega una biografía y foto de perfil. Agrega Estados de Ánimo para describir mejor tus emociones. Agrega Aliados - amigos, familia, y seres queridos con los que quieras compartir. Escribe tus historias. Crea Categorías personalizadas para organizar tus momentos. Escribe sobre los Momentos - los eventos y situaciones que afectan tu salud mental. Cuídate. ¡Elabora tus Estrategias de cuidado personal! . ¿Tomas un Medicamento? ¡Realiza un seguimiento y crea recordatorios. ¿Necesitas apoyo extra? Únete a o crea un Grupo para hablar sobre ello. ¿Tienes alguna duda o pregunta? ¡Revisa nuestra página de Preguntas Frecuentes o escríbenos. if me Acerca de Blog Contribuidores Preguntas Frecuentes Socios Prensa Política de Privacidad conectarse Correo electrónico Facebook GitHub Instagram Medium Open Collective Patreon RSS Twitter recursos Comunidades Educación Líneas de Ayuda Servicios idioma English Español Português Svenska Retroalimentar Nosotros ♥ FOSS Licencia bajo AGPL-3.0"

What's really happening is that the test is unable to log back in the second time around.

baohouse:
[...]

PR #736, opened 2017-10-22
Making tests more resilient to page transitions

baohouse:
Addresses #723


baohouse:
Thanks @alvarocasadoc!

Issue #424, opened 2017-03-04
Invite multiple people by email


julianguyen:
Currently, users can only invite one person a time by email. We want to add the ability to add multiple emails that are comma separated (including comma-space separated).

Current behaviour:
![image](https://cloud.githubusercontent.com/assets/3010728/23581658/b53c321c-00cc-11e7-9803-67a659649da4.png)

nehalita:
@julianguyen I'll try this one!

liirene1:
Note: This issue has a prerequisite before completing this task: Devise knowledge required. You have to override the Devise authentication.

liirene1:
To see this component on localhost: 
1. Click **Sign In** on header
2. use the test username and passwords in the _seeds.rb_ file 
3. Click **Allies** on header 
4. Click Invite **Your Loved Ones** 

http://localhost:3000/users/invitation/new

<img width="585" alt="screen shot 2017-03-22 at 10 38 08 pm" src="https://cloud.githubusercontent.com/assets/16128022/24233560/497af062-0f50-11e7-85ef-e6207c59c392.png">

logeshmallow:
I can add the multiple email invite provision. Can I take that and send you the pull request?

julianguyen:
Hi @logeshmallow! Feel free to work on it and make a PR!

julianguyen:
Thank you

julianguyen:
@logeshmallow How is this going?

logeshmallow:
@julianguyen I have worked on it and I will try to complete it by this weekend and send you PR.

jnamster:
![](https://github.trello.services/images/mini-trello-icon.png) [(3) Invite multiple people by email](https://trello.com/c/kKFi5Xqh/23-3-invite-multiple-people-by-email)

haleymnatz:
@julianguyen I'd like to give this a shot if that's alright!

PR #742, opened 2017-10-26
Issue 424

haleymnatz:
- Configure Invitations controller to allow multiple email addresses to be entered (sends unique token to each user)
- Reconfigure invalid email error on invitations form to be displayed as flash message
- Add input field placeholder with instruction
- Small CSS change to extend field length
- Add contributor blurb
- Tests for invitations controller

Addresses #424.


andy5995:
Hi @haleymnatz 

I see codeclimate reports 2 errors. Is this a wip?

haleymnatz:
@andy5995 yeah, just 2 rubocop errors to take care of. I'll fix those and merge the latest master into my branch.

config/locales/devise_invitable.en.yml
```diff
@@ -4,6 +4,7 @@ en:
       invited: "You have a pending invitation, accept it to finish creating your account."
     invitations:
       send_instructions: "An invitation email has been sent to %{email}."
+      failed_send: "An error occurred. %{email} could not be invited."
```
andy5995: From a grammar POV, this sounds a little choppy to me. I'm used to see things like

>Error: ${email} could not be invited

or 

>${email} could not be invited (an unknown error occurred)

But it's not up to me @haleymnatz , I just wanted to make a note of it. What does @julianguyen  or @baohouse think?

andy5995:
You've tested this and it works? If so, I'm confident I can approve it and you can merge it. Looks pretty straightforward. If you've tested it and it works as intended, I'll say LGTM!

config/locales/devise_invitable.en.yml
haleymnatz: @andy5995 I definitely can change it if that's preferred. I'll just have to get the translations for the changed version as well.

haleymnatz:
@andy5995 it passed all of the tests locally, and it sent emails via the dev emailing method (opens up an 'email' tab in my browser). If you're unsure though, it certainly can wait!

andy5995:
Hmmm...  I may as well give Julia a chance to look it over. :) Like you said, no rush. :)

julianguyen:
Tested your branch locally and I'm getting this error:

image

The letter opener gem mocks an email that looks like this:
image

Could you also look into the bug where the subject in the email is %{name}?

app/assets/stylesheets/application/allies.scss
```diff
@@ -12,3 +12,7 @@
 #allies_search {
 	text-align: center;
 }
+
+#user_email {
+	width: 100%;
```
julianguyen: Ooh this is great!

app/controllers/users/invitations_controller.rb
```diff
@@ -3,20 +3,27 @@
 module Users
```
julianguyen: This spec doesn't exist currently but should create invitations_controller_spec.rb and tests these methods extensively.

Along with testing the working cases, you should also test invalid input as well!

config/locales/devise_invitable.en.yml
julianguyen: You could just rephrase this to `An error occurred, "%{email}" could not be invited.` I would put quotes between "%{email}" in case there is a blank string.

haleymnatz:
@julianguyen in response to the comment with the screenshot - what functionality would you like to see when the user enters a blank string after a comma? Do you just want me to add quotes around %{email}, or do you want me to have the code ignore the empty space completely and not return an error message?

Also, regarding the %{email} name in the email's subject: that was actually happening on the master branch before I started working on this feature. It was just brought up in another issue ( #752 ). I'd be happy to work on that too, but maybe in a separate PR to keep this one from covering too much?

julianguyen:
@haleymnatz oooh yeah let's fix the subject line bug in another PR then! Good call

julianguyen:
Sorry I didn't reply to the rest of your comment @haleymnatz!

RE:

```
 in response to the comment with the screenshot - what functionality would you like to see when the user enters a blank string after a comma? Do you just want me to add quotes around %{email}, or do you want me to have the code ignore the empty space completely and not return an error message?
```

For both cases, I would always return quotes e.g. `"%{email}"`. Let me know if you have any questions about adding tests for the controller as well! :)

haleymnatz:
@julianguyen I have made the changes your requested and added tests. If you feel that there are any cases/scenarios that I didn't write tests for but should, let me know. Otherwise, I think this is good to go?

julianguyen:
will re-review sometime this weekend! Great work :)

app/controllers/users/invitations_controller.rb
```diff
@@ -3,20 +3,34 @@
...
-        respond_with_navigational(resource) { render :new }
-      end
+    def invitation_flash_messages(invites, fails)
+      return unless is_flashing_format?
+      invites_flash(invites) unless invites.empty?
+      fails_flash(fails) unless fails.empty?
+    end
+
+    def invites_flash(invites)
+      set_flash_message :notice, :send_instructions, email: invites.join(', ')
```
julianguyen: Can we join using just `','` (without a space after)?

doc/pages/blurbs.json
```diff
@@ -214,5 +214,14 @@
     "link_name": "maribelduran",
     "link": "http://github.com/maribelduran",
     "social": "github"
+  },
+  {
+    "name": "Haley Mnatzaganian",
```
julianguyen: <3

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+            invite_two_friends
+            friend1 = User.find_by(email: invitee1)
+            friend2 = User.find_by(email: invitee2)
+            expect(friend1.invitation_token).not_to eq(friend2.invitation_token)
+          end
+        end
+      end
+
+      context 'when invalid params are given' do
+        let(:invalid_invite) { post :create, params: { user: { email: "#{invalid_email}" } } }
```
julianguyen: I would also test when there's a combination of valid and invalid params e.g. "invalid_email, valid_email@example.com"

app/controllers/users/invitations_controller.rb
julianguyen: Thanks for addressing this!

spec/controllers/users/invitations_controller_spec.rb
julianguyen: Tested locally and it looks like it the valid email will get an invite and the invalid one won't which is good! :)

Issue #712, opened 2017-10-13
Add character limit counter to CKEditor modal in mobile web


julianguyen:
Let's add the character limit counter to the bottom of the CKEditor went it's opened in a modal in mobile web.

Ignore the green reload icon in the screenshots, that's a Grammarly Chrome Extension lol.

![image](https://user-images.githubusercontent.com/3010728/31562006-b425bd2a-b00e-11e7-9b5b-e2f34a69de09.png)

![image](https://user-images.githubusercontent.com/3010728/31562019-bf3509dc-b00e-11e7-8f5f-a99a090d0357.png)

tsara27:
Hey @julianguyen!
I can help you adding that. :D

PR #745, opened 2017-10-26
Add char counter on modal editor

tsara27:
This fix enhancement for #712. I have added character count on modal editor. This might add character counter to every text editor that appears when editing in mobile view.


tsara27:
Thanks @julianguyen! 😄

Issue #741, opened 2017-10-25
Can't leave moods/categories/strategies/viewers blank when updating a Moment/Strategy


alvarocasadoc:
I don't know if this is already reported or if I'm doing anything wrong, but I've noticed that when I try to edit a moment and remove a viewer from the list the change is not updated in the database. Below is a little demo of the bug.
![](https://user-images.githubusercontent.com/7248474/31982969-773a902e-b95b-11e7-8053-f6a48eecda91.gif)

julianguyen:
Uhoh! Thanks for identifying this, this is definitely a bug!

alvarocasadoc:
I'm going to try to fix this one then

alvarocasadoc:
I've noticed this also happens to moods, categories and strategies. This bug happens only when a user is trying to delete all the viewers from a Moment or a Strategy, this means the param. array is nil when the edit form is submitted. 

My approach is to assign an empty array to the moods, categories, viewers or strategies whenever the corresponding param. array is nil due to the lack of any item selected in the list.

julianguyen:
Thanks for offering to fix this @alvarocasadoc, much appreciated!

PR #756, opened 2017-10-29
Fixing issue #741 

buitranquyet:
Issue: https://github.com/julianguyen/ifme/issues/741
Overlap with what have been done in PR: https://github.com/julianguyen/ifme/pull/753
But this is another approach and also come with some simple specs


buitranquyet:
Closing this PR as it solves the same issue that #753 does, but #753 have been merged.

PR #758, opened 2017-10-30
add/change info on group page [wip][skip ci][closes #606]

### LINKED ISSUE #606: Change text displayed for users with 0 groups
[](https://postimg.org/image/b1j793n5x/)

me:
> 
> As a new user, I found this confusing: "You aren't a part of any groups yet! Click to create one or join any available groups below." The confusion was that there were no groups listed, but directly underneath the "Connect" menu is shown with all the social networks. So I thought, am I being directed to join _those_ groups? I realized "no". So there is some feedback from a new user perspective
> 
> Perhaps in this case, there should be a message under that first line that states something like "0 groups found" ? 

julianguyen:

> Hmmm yeah that could be a good improvement to indicate none exist between you and your allies.

hashnotadam:

> Or, if there are not groups, don't show "or join any available group below"?

tmoore:

> Maybe even a little instruction text about what groups are and how to create or find them.

andy5995:
* display more detailed info about groups by using text from FAQ
* wip

I changed part of the EN text (pending approval) to:
You aren't a part of any groups yet, but you can click %{icon} to create one!

Formerly, it was:
You aren't a part of any groups yet! Click to create one or join any available groups below.

Let me know if that's ok, @julianguyen, and I'll get the translations and test the patch.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -145,7 +145,7 @@ en:
    available_groups: 'Available Groups'
    joined_groups: 'Joined Groups'
    leader: '(Leader)'
-   instructions: "You aren't a part of any groups yet! Click %{icon} to create one or join any available groups below.<br><br>Support groups are great for having open discussions about mental health! You can create groups, invite members, and organize meetup events online or offline. You can also join groups your allies have created."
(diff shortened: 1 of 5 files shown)


andy5995:
PR to fix this in #759

PR #759, opened 2017-10-30
[wip][skip ci] add info to group page (closes #606)

### LINKED ISSUE #606: Change text displayed for users with 0 groups

andy5995:
If you like this change, I'll do the other LANG files, truncating off
everything after the two `<br />` tags, and get the translations for
the sentence that precedes it.

### DIFF
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -49,8 +49,10 @@
     }) %>
   <div class="spacer"></div>
 
-  <h1 class="faq_heading"><%= t('pages.faq.group_question') %></h1>
-  <br />
+  <h1 class="faq_heading">
+    <%= t('pages.faq.group_question') %>
+  </h1>
+
   <%= raw t('pages.faq.group_answer') %>
 <% end %>
 
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -47,12 +47,11 @@
   <%= raw t('groups.index.instructions', {
       icon: '<i class="fa fa-plus-circle"></i>'
     }) %>
+  <br /><br />
+  <h1 class="faq_heading"><%= t('pages.faq.group_question') %></h1>
+  <br />
+  <%= raw t('pages.faq.group_answer') %>
 <% end %>
-
-<br /><br />
-<h1 class="faq_heading"><%= t('pages.faq.group_question') %></h1>
-<br />
-<%= raw t('pages.faq.group_answer') %>
 
 <% if @available_groups.any? %>
   <div class="spacer"></div>
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -47,7 +47,8 @@
   <%= raw t('groups.index.instructions', {
       icon: '<i class="fa fa-plus-circle"></i>'
     }) %>
-  <br /><br />
+  <div class="spacer"></div>
+
   <h1 class="faq_heading"><%= t('pages.faq.group_question') %></h1>
   <br />
   <%= raw t('pages.faq.group_answer') %>
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -49,6 +49,11 @@
     }) %>
 <% end %>
 
+<br /><br />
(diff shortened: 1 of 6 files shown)


app/views/groups/index.html.erb
```diff
@@ -49,6 +49,11 @@
     }) %>
 <% end %>
 
+<br /><br />
+<h1 class="faq_heading"><%= t('pages.faq.group_question') %></h1>
+<br />
```
andy5995: I added this extra `<br />` because for some weird reason I didn't get an extra line break after the `</h1>` tag

app/views/groups/index.html.erb
```diff
@@ -49,6 +49,11 @@
     }) %>
 <% end %>
 
+<br /><br />
```
andy5995: Oops.. I need to move my new block of code inside the `else` clause above, right?  After this line? `  }) %>` ?

```
<% else %>
  <%= raw t('groups.index.instructions', {
      icon: '<i class="fa fa-plus-circle"></i>'
    }) %>
<% end %>
```

app/views/groups/index.html.erb
```diff
@@ -47,6 +47,10 @@
   <%= raw t('groups.index.instructions', {
       icon: '<i class="fa fa-plus-circle"></i>'
     }) %>
+  <br /><br />
+  <h1 class="faq_heading"><%= t('pages.faq.group_question') %></h1>
+  <br />
```
julianguyen: You shouldn't need to put a `<br>` here since there's a `<h1>` above it.

app/views/groups/index.html.erb
```diff
@@ -47,6 +47,10 @@
   <%= raw t('groups.index.instructions', {
       icon: '<i class="fa fa-plus-circle"></i>'
     }) %>
+  <br /><br />
```
julianguyen: Using `<br>` should be sufficient. Though I would either wrap the lines before that in a `<p>` or replace `<br /><br />` with `<div class="spacer"></div>`

config/locales/en.yml
```diff
@@ -145,7 +145,7 @@ en:
    available_groups: 'Available Groups'
    joined_groups: 'Joined Groups'
    leader: '(Leader)'
-   instructions: "You aren't a part of any groups yet! Click %{icon} to create one or join any available groups below.<br><br>Support groups are great for having open discussions about mental health! You can create groups, invite members, and organize meetup events online or offline. You can also join groups your allies have created."
+   instructions: "You aren't a part of any groups yet, but you can click %{icon} to create one!"
```
julianguyen: Wouldn't this translation need to change for other languages too?

app/views/groups/index.html.erb
andy5995: yes, one wouldn't think so. :) But when I tested it, there was no line break. Strange, isn't it. I'll try it your way again, and test it when the patch is done. Any problems and I'll follow up with you on that.

config/locales/en.yml
andy5995: yes, I wanted your approval before getting translation for 4 other languages ;) I noted that in the comments when I submitted the PR. ;)

julianguyen: Thanks Andy!

andy5995:
You're welcome, Julia!

So.. I have to confess, I didn't test it one last time. I see that removing the `<br />` under the `<h1>` tag took away a line break.

Wierd, huh? You mentioned in a line comment that a `<br />` shouldn't be needed. I agree. Wierd..

screenshot_2017-11-08_03-14-18

PR #761, opened 2017-10-31
Add Calender

yuganshchokra:
Pull request for add google calendar functionality to "meetings" for google authenticated users.

*Now, the meeting shows an option to "add event to google calendar" if have not added before.
*And show "Delete from google calendar" if the event already exists and it's have not cancelled


kelunik:
Please close and mark as invalid. User reported.

Issue #752, opened 2017-10-28
Ally Invite Email Bug?


3minus1:
So when I try to invite an ally(tested both on production and locally), the subject of the email doesn't show up the inviter's name. And also, sender's name shows up as `smtp.sendgrid.net`

![screen shot 2017-10-28 at 11 31 46 am](https://user-images.githubusercontent.com/13695333/32131765-fcabeefe-bbd3-11e7-8c19-ff99ca179f2c.png)

julianguyen:
Hey @3minus1, this is correct. Thanks for calling this out! @haleymnatz fixed the subject bug in a [related PR ](https://github.com/julianguyen/ifme/pull/742) . The sender name is correct, we use SendGrid :)

andy5995:
Is that a bug I caused? (#621) The sender's name is not showing up because quotes are not used?

As for changing the "from" field from "smtp.sendgrid.net", there was a comment by @haleyelder about that in https://github.com/julianguyen/ifme/issues/127#issuecomment-327630565

>It doesn't look like if-me.org has SPF or DKIM records so, while the from address should absolutely be changed, the secret sauce will be creating a new DNS record.

haleymnatz:
@andy5995 I don't think that's what is causing the issue, because there also aren't quotes around %{email} in the email body (as can be seen in the screenshot above - "Metal Muse"). But I could be wrong?

haleymnatz:
I figured out the source of the issue, and have a possible solution. I assigned myself to the issue and will take it on once my other PR is pulled in (since they are both related to invitations).

PR #762, opened 2017-11-02
Issue 752

haleymnatz:
- Add `subject_for` method in CustomDeviseHelper that is only called when `invitations_instructions` key is used. Method allows name of inviting user to be passed as an argument (the name previously wasn't being passed in anywhere, leading to the bug). If no name is present, selects generic invited key in translation files.
- Tests for `subject_for` method
- Create Invitations Helper & `inviter_name` method to make code more DRY.

Addresses #752 .

### DIFF
--- a/app/helpers/invitations_helper.rb
+++ b/app/helpers/invitations_helper.rb
@@ -2,7 +2,7 @@
 
 module InvitationsHelper
   def inviter_name(resource)
-    User.where(id: resource.invited_by_id).first.presence.try(:name) ||
+    User.where(id: resource.invited_by_id).first.try(:name) ||
       'Someone'
   end
 end
--- a/app/helpers/invitations_helper.rb
+++ b/app/helpers/invitations_helper.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module InvitationsHelper
+  def inviter_name(resource)
+    User.where(id: resource.invited_by_id).first.presence.try(:name) ||
+      'Someone'
+  end
+end
--- a/app/helpers/invitations_helper.rb
+++ b/app/helpers/invitations_helper.rb
@@ -2,7 +2,6 @@
 
 module InvitationsHelper
   def inviter_name(resource)
-    User.where(id: resource.invited_by_id).first.try(:name) ||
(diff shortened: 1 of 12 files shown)


app/helpers/invitations_helper.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module InvitationsHelper
+  def inviter_name(resource)
+    User.where(id: resource.invited_by_id).first.presence.try(:name) ||
```
julianguyen: Could do `User.find(resource.invited_by_id).try(:name)` instead!

app/helpers/invitations_helper.rb
```diff
@@ -0,0 +1,8 @@
...
+      'Someone'
```
julianguyen: This should be a key in our translation files!

spec/mailers/custom_devise_mailer_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+describe CustomDeviseMailer do
+  let(:inviter) { create(:user2) }
+  let(:recipient) { '123@abc.com' }
+  let(:key) { 'invitation_instructions' }
+  let(:email) { Devise.mailer.deliveries.last }
+
+  describe 'subject_for' do
+    context 'if the inviting user has a name' do
+      it 'includes the inviter\s name in the subject' do
```
julianguyen: Should this be `'includes the inviter\'s name in the subject'`?

app/helpers/invitations_helper.rb
haleymnatz: Could you clarify what you're thinking exactly? I think I'm a little confused.

julianguyen: Oh sorry! What I meant to say was someone should be a value in en.yml and that needs to be translated to our other languages too.

haleymnatz:
@julianguyen I implemented the changes you suggested. This should be ready to be merged!

timofonic:
@haleymnatz Nice!

@julianguyen @andy5995 Any hopes to see this merged? :D

PR #764, opened 2017-11-11
pages/education.json: add resource

andy5995:


timofonic:
@andy5995 Thanks! I have some ideas related to ADHD, but I'm also bored of common ADHD sites that are more into business than caring.

@julianguyen @baohouse Any news?

andy5995:
@timofonic Thanks for the support!

Hi @baohouse :)

PR #771, opened 2017-12-05
767 Alphabetize Methods

dbarner1:
Issue 767

Alphabetizes the methods in all models.


dbarner1:
First commit here - woot!

PR #773, opened 2017-12-07
Increase locale selector Capybara wait time

TimMoore:
This should help prevent flaky tests.

### DIFF
--- a/spec/support/page_transition_support.rb
+++ b/spec/support/page_transition_support.rb
@@ -1,12 +1,16 @@
+# frozen_string_literal: true
+
 module PageTransitionSupport
-  def change_language(languageCode)
+  def change_language(language_code)
     within 'select[name=locale]' do
-      find("option[value=#{languageCode}]").click
-      expect(find("option[value=#{languageCode}][selected]", wait: 5)).to be_present
+      find("option[value=#{language_code}]").click
+      expect(
+        find("option[value=#{language_code}][selected]", wait: 5)
+      ).to be_present
     end
   end
 end
 
 RSpec.configure do |config|
-  config.include PageTransitionSupport, :type => :feature
+  config.include PageTransitionSupport, type: :feature
 end
--- a/spec/support/page_transition_support.rb
+++ b/spec/support/page_transition_support.rb
@@ -2,11 +2,11 @@ module PageTransitionSupport
   def change_language(languageCode)
     within 'select[name=locale]' do
       find("option[value=#{languageCode}]").click
-      expect(find("option[value=#{languageCode}][selected]")).to be_present
+      expect(find("option[value=#{languageCode}][selected]", wait: 5)).to be_present
     end
   end
 end
-    
+
 RSpec.configure do |config|
   config.include PageTransitionSupport, :type => :feature
 end


spec/support/page_transition_support.rb
```diff
@@ -1,12 +1,16 @@
...
+
 module PageTransitionSupport
-  def change_language(languageCode)
+  def change_language(language_code)
     within 'select[name=locale]' do
-      find("option[value=#{languageCode}]").click
-      expect(find("option[value=#{languageCode}][selected]")).to be_present
+      find("option[value=#{language_code}]").click
+      expect(
+        find("option[value=#{language_code}][selected]", wait: 5)
```
TimMoore: The key part is `wait: 5`. The default wait time is 2 seconds, which might not be enough time for the page to reload.

All of the other changes are Rubocop fixes.

TimMoore: If people still have problems, we can increase this further. There's not much downside to having a long wait time, because if it sees the element earlier, it won't wait the entire duration.

julianguyen: Hmmm should we be relying on waits and sleeps to solve these flakey test problem?

TimMoore: Sleeps, no. A sleep would pause execution for 5 seconds minimum. A wait in Capybara is more like a time out: it will wait _up to_ 5 seconds for the element to appear.

This is the best way that I know of to solve this kind of problem with Capybara.

TimMoore:
@julianguyen we could increase the default Capybara wait time with `Capybara.default_max_wait_time = 5`.

The reason I chose not to do that is that this could potentially make other tests slower.

Working with Capybara in other projects, I have found that the wait time is not a "one size fits all" thing. It is good to use different wait times depending on what you are doing in the test. Anything that causes a full page reload can be a bit slower. This locale selector actually triggers two requests in sequence: first an AJAX request that updates the user's locale, then a page reload. That is just enough to sometimes push it over the 2 second default limit. Other operations might be expected to be quicker, especially anything on the client side that involves asynchronous JavaScript _without_ making a request to the server.

If you set a large wait time by default, there are a couple of possible drawbacks:

- When tests do fail, they will take longer to fail. Not the biggest problem, because hopefully this is rare, but can be annoying.
- If you have any negative assertions (in other words, tests that something _isn't_ on the page) Capybara will wait for the entire wait time before passing. This is a bigger problem, because it would slow down other tests.

For example, in the `user_creates_a_draft_moment.rb` feature spec, it has this code:

```ruby
      login_as ally
      visit back
      within '#page_title_content' do
        expect(page).not_to have_content 'My new moment'
      end
```

The `expect(page).not_to have_content 'My new moment'` line will wait for the entire `Capybara.default_max_wait_time` duration, because it wouldn't be correct to assert that the page doesn't have certain text if the page hasn't completely loaded yet. If we set the default to be larger, all of the tests that use this kind of pattern will become slower.

On the other hand, if you think it would be a worthwhile trade-off to set a large default max wait time just so people don't have to think about it, we could try that. Optionally, we could also change all of the negative assertions like the example above to have lower explicit wait times.

See https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends

julianguyen:
Thanks so much for taking the time to explain all this @TimMoore! I really appreciate it :)

PR #792, opened 2018-01-26
778 input component

HeyStevenXu:
A React component for the two inputs we have in the design. Let me know if I need to make any changes!

### DIFF
--- a/client/app/bundles/shared/components/Input.jsx
+++ b/client/app/bundles/shared/components/Input.jsx
@@ -17,15 +17,19 @@ type Props = {
   maxLength?: number
 };
 
-export default class Input extends React.Component<Props, {}> {
-  constructor(props) {
+type State = {
+  value: string | number
+}
+
+export default class Input extends React.Component<Props, State> {
+  constructor(props : Props) {
     super(props);
     this.state = { value: this.props.value || "" };
   }
 
-  onChange = e => {
+  onChange = (e : SyntheticEvent<HTMLInputElement>) => {
     e.preventDefault();
-    this.setState({ value: e.target.value });
+    this.setState({ value: e.currentTarget.value });
   };
 
   render() {
--- a/client/app/bundles/shared/components/Input.jsx
+++ b/client/app/bundles/shared/components/Input.jsx
@@ -1,0 +1,48 @@
+//@flow
+import React from "react";
+import css from "./Input.scss";
+
+type Props = {
+    dark?: boolean;
+    type? : string;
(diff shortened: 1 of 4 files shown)


client/app/bundles/shared/components/Input.scss
```diff
@@ -0,0 +1,54 @@
...
+  padding: 12px 0;
+  letter-spacing: 0.0944444em;
+  text-indent: 4px;
+  text-transform: uppercase;
+  opacity: 0.5;
+}
+
+.labelLight {
+  composes: label;
+  color: #ffffff;
```
julianguyen: We should use $background-color here. Feel free to update its value in `app/assets/stylesheets/base/_colors.scss` to `#FFFFFF` :)

client/app/bundles/shared/components/Input.scss
```diff
@@ -0,0 +1,54 @@
+@import "base.scss";
+
+.label {
+  font-family: $font-family;
+  padding: 12px 0;
+  letter-spacing: 0.0944444em;
```
julianguyen: We should use px values here. Thoughts @nshki?

julianguyen:
Think you'll have to rebase from master again to get the tests running properly!

client/app/bundles/shared/components/Input.jsx
```diff
@@ -0,0 +1,56 @@
...
+    this.setState({ value: e.currentTarget.value });
+  };
+
+  render() {
+    const {dark, label, name, placeholder, type, readonly, disabled, required, minLength, maxLength } = this.props;
+    return (
+        <div>
+          <div className={dark ? css.labelDark : css.labelLight}>{label}</div>
+          <input
+            className={dark ? css.inputDark : css.inputLight}
```
julianguyen: Hmm I think we also need to have an "id" attribute!

HeyStevenXu:
Thanks Julia! Made those changes. Question though - I see on codeclimate that the render function can't contain more than 25 lines? I feel like this would be difficult to achieve for more complicated components. I can reduce the line count for this component, but are we planning for this to be a hard rule?

If either of you have more comments, just let me know :)

client/app/bundles/shared/components/Input.scss
nshki: I think `em`s are okay here, mainly because that would ensure `letter-spacing` scales with font size changes.

julianguyen:
You can ignore that Codeclimate warning about line numbers, in your PR it's fine!

Issue #788, opened 2018-01-09
[Component] Footer


nshki:
## Overview

We are looking for an awesome human to build this out as a React component, preferably within Storybook. You can view and get details (colors, dimensions, export SVGs, etc.) by visiting [the Figma designs](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618) and creating a free account.

This is part of the app redesign (#691). This incorporates another component #784.

## Screenshot from [the Figma document](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618):

![screenshot 2018-01-09 at 6 32 30 pm](https://user-images.githubusercontent.com/1121087/34748628-7eb52098-f56b-11e7-89cd-dab9407a9844.png)

Seunadex:
Hi @nshki I'll like to take this up if no one is working on it yet.

nshki:
@Seunadex That would be great! This component incorporates one other component (#784), but feel free to start this one.

Seunadex:
@nshki Concerning the dropdown on the lower left of the footer ,do I need to add that or will it come from the reusable dropdowns in another issue?

nshki:
Ideally the dropdown should be built in a separate PR first, and then incorporated into this one!

Seunadex:
@nshki Also i'm issue getting to use the image in the text on the lower right of the footer

Seunadex:
@nshki I'm trying to raise a PR now, but i'm getting the `can't automatically merge` error

nshki:
@Seunadex Don't worry about the background images, those are simply a part of the brochure page backgrounds. The background color of the footer should just be a slightly transparent shade with no images.

The error might be because you don't have the latest changes pulled down. Make sure you pull down the latest changes from the original repo.

Seunadex:
@nshki , @julianguyen  I have a PR up kindly review and let me know if you have any feedback

PR #794, opened 2018-02-02
React component for the footer

Seunadex:
#### What does this PR do?
It add the react component for the footer as started in this [issue](https://github.com/ifmeorg/ifme/issues/788)

screen shot 2018-02-17 at 6 15 52 am
screen shot 2018-02-17 at 6 16 26 am
screen shot 2018-02-17 at 6 16 41 am


app/assets/stylesheets/application/footer.scss
```diff
@@ -5,8 +5,14 @@
 /* Footer */
 
 #footer {
-	background-color: $secondary-background-color;
+	// background: $secondary-background-color;
```
julianguyen: Why is this commented out and not just removed?

app/assets/stylesheets/application/footer.scss
```diff
@@ -5,8 +5,14 @@
...
 	padding: 40px 0px;
+	/* Gradient */
+	left: 0px;
+	top: 0px;
+
+	background: linear-gradient(142.51deg, #700F3C 62.56%, #D0E799 204.67%);
```
julianguyen: I see that you re-defined `$secondary-background-color` as this. Why not use that?

client/app/bundles/shared/components/Footer.jsx
```diff
@@ -0,0 +1,54 @@
...
+import css from './Footer.scss';
+
+const Footer = () => (
+  <div className={css.footer}>
+    <div id={css.footer_content}>
+      <div className={css.table}>
+        <div className={css.row}>
+          <div className={[css.table_cell, css.resources].join(' ')}>
+            <ul>
+              <h6 className={css.footer_header}>Resources</h6>
```
julianguyen: We should be using i18n keys here since we translate our app into different languages.

@nshki @baohouse Do you think it's better to use i18n keys directly in here or pass the text through props? For a basic component like a button, I think passing through props makes sense. But a footer is a more "complicated" component 🤔

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,75 @@
...
+      
+      ul > li {
+        list-style: none;
+        line-height: 24px
+      }
+      
+      ul li a {
+        text-decoration: none;
+        color: $background-color;
+        letter-spacing: 0.05em;
```
julianguyen: Let's use px here!

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,75 @@
...
+      text-align: right;
+      color: $background-color;
+      margin-right: 30px;
+      font-family: Lato, Inter UI, Hiragino Sans;
+      font-style: Regular;
+      font-size: 20px;
+
+      p {
+        font-size: 18px;
+        line-height: 0px;
```
julianguyen: Does this need to be set to 0px? Isn't this the default?

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,75 @@
...
+    .resources {
+      text-align: right;
+      margin-right: 30px;
+    }
+
+    .love_foss {
+      text-align: right;
+      color: $background-color;
+      margin-right: 30px;
+      font-family: Lato, Inter UI, Hiragino Sans;
```
julianguyen: We should just set this to Lato.

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,75 @@
+@import 'base';
+
+.footer {
+  width: 100%;
+  height: 450px;
+  background: rgba(0, 0, 0, 0.5);
+  font-family: Lato;
+  font-style: regular;
```
julianguyen: Does this need to be here?

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,75 @@
+@import 'base';
+
+.footer {
```
julianguyen: @nshki out of curiosity, is there a responsive mobile/tablet version of the footer?

nshki: There should be, but I ended up not mocking one up, which is my bad! However, I think the columns should simply stack and text align center on smaller screens.

client/app/bundles/shared/components/Footer.jsx
nshki: I like passing the text through props as a consistent pattern across all components! It would keep the components a little cleaner and lower the barrier of entry for new contributors. That being said though, I don't have a lot of experience working with multilingual apps, so I'd love to know if there are any additional insights here I can learn from.

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,103 @@
...
+      
+      ul > li {
+        list-style: none;
+        line-height: 24px
+      }
+      
+      ul li a {
+        text-decoration: none;
+        color: $background-color;
+        letter-spacing: 1px;
```
nshki: Let's take out the letter-spacing here! I just checked the Figma CSS, and the 0.05em that was recommended was likely just for the column label.

client/app/bundles/shared/components/Footer.scss
```diff
@@ -0,0 +1,103 @@
...
+    .table_cell {
+      width: 26%;
+
+      ul {
+        margin-top: 50px;
+
+        .footer_header {
+          color: $header-text-color;
+          text-transform: uppercase;
+          margin-bottom: 10px;
```
nshki: Let's change to 5px here! Let's also make the opacity 0.5.

app/assets/stylesheets/base/_colors.scss
```diff
@@ -4,7 +4,7 @@
 $primary-text-color: #000;
 $secondary-text-color: #6D0839;
 $highlight-text-color: #D0E799;
-$header-text-color: #fff;
+$header-text-color: #555;
```
nshki: Let's change this to `#fff`! We'll adjust the opacity of the actual header in the CSS.

client/app/bundles/shared/components/Footer.jsx
```diff
@@ -0,0 +1,54 @@
...
+              <li><a href="/contribute">Contribute</a></li>
+              <li><a href="/faq">FAQ</a></li>
+              <li><a href="/partners">Partners</a></li>
+              <li><a href="/press">Press</a></li>
+              <li><a href="/privacy">Privacy Policy</a></li>
+            </ul>
+          </div>
+          <div className={[css.table_cell, css.love_foss].join(' ')}>
+            <h4>We Love FOSS</h4>
+            <p>Licensed under AGPL-3.0</p>
```
nshki: Let's make sure the `<h4>` and `<p>` here are sitting about 5px away from each other vertically. They're currently too far apart.

client/app/bundles/shared/components/Footer.jsx
baohouse: Passing in a language mappings object into the Footer component via props just means you have yet-another-DTO to maintain, especially if I were using Flow or TypeScript. Keep it simple and just use directly. Libraries like react-18next automagically make it easy to use the direct-usage pattern, but under the hood is actually passing in an language map through the React component's context or props and rewriting the content of the <Trans> components.

baohouse: Actually on second thought, looking at the example at /apps/views/moments/index.html.erb, how would we directly use the ruby translation helper method? The React component is separate. We'll need to dig more into this.

julianguyen: Hmmm it makes sense to use i18n directly in the component, but I see UI components as something reusable that you pass data into. We could split the footer component into smaller ones, have a generic component for each column. Then pass props into each one.

client/app/bundles/shared/components/Footer.scss
julianguyen: Would be helpful to have mockups for that if you have any time :P

Seunadex:
@julianguyen @nshki all comments have been addressed, kindly go through again, having some codeclimate issue though.

client/app/bundles/shared/components/Footer/Footer.scss
```diff
@@ -0,0 +1,133 @@
...
+    .table_cell {
+      width: 26%;
+
+      @media screen and (max-width: $small) {
+        width: 75%;
+      }
+
+      
+
+      ul {
```
nshki: Let's make sure the default left padding isn't there and adjust spacing accordingly! It's especially noticeable when the columns stack on top of one another and the text is slightly off-center.

client/app/bundles/shared/components/Footer/Footer.scss
```diff
@@ -0,0 +1,133 @@
+@import '../base';
+
+.footer {
+  width: 100%;
+  height: 450px;
```
nshki: Let's actually take out the fixed height so the container can resize based on its children. This causes the "We <3 FOSS" copy to overlap a bit with the bottom part of the container as well, so if you could fix that up, that would be great!

Seunadex:
@nshki , @julianguyen  All your comments has been addressed and the language dropdown has also been added to the footer, kindly check it out

Issue #781, opened 2018-01-09
[Component] Textarea


nshki:
## Overview

We are looking for an awesome human to build this out as a React component, preferably within Storybook. You can view and get details (colors, dimensions, export SVGs, etc.) by visiting [the Figma designs](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618) and creating a free account.

This is part of the app redesign (#691).

## Screenshot from [the Figma document](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618):

![screenshot 2018-01-09 at 6 21 54 pm](https://user-images.githubusercontent.com/1121087/34748329-fe3d5cec-f569-11e7-80d2-09678c5afb5c.png)

nshki:
All you @HeyStevenXu!

PR #803, opened 2018-02-07
781 textarea component

HeyStevenXu:
A Textarea React component. Aesthetically, I chose to make the animation more subtle than the input components, so it would be easier on the eyes for extended writing. Let me know what you think!


client/app/bundles/shared/components/Textarea.scss
```diff
@@ -0,0 +1,54 @@
...
+$focus-background-color-lighten: lighten($container-color, 3%);
+$light-font-weight: 300;
+
+.label {
+  color: $container-color;
+  font-family: $font-family;
+  font-size: $small-font;
+  padding: 12px 0;
+  letter-spacing: 0.0944444em;
+  text-indent: 4px;
```
nshki: Let's make this a padding, since when the label breaks onto multiple lines on smaller screens it looks a little funky!

HeyStevenXu: Thanks for noticing that! I also ended up adjusting the padding and line height in the input portion, just a heads up! Hopefully made it easier to read.

client/app/bundles/shared/components/Textarea.scss
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.textarea {
+  color: rgba(255, 255, 255, 0.8);
+  box-sizing: border-box;
+  background-color: $container-color;
+  font-family: $font-family;
+  font-size: $font18;
+  line-height: $font30;
+  text-indent: 12px;
```
nshki: Let's make sure we avoid using text-indent here as well, since it makes the text look funky on multiple lines in the textarea! Sorry I didn't catch this the first time!

HeyStevenXu: Not at all! I thought it was intentional, so I left it there. Changes made :)

client/app/bundles/shared/components/Textarea.jsx
```diff
@@ -0,0 +1,91 @@
...
+    this.state = { value: "", active: false };
+  }
+
+  onChange = (e: SyntheticEvent<HTMLInputElement>) => {
+    e.preventDefault();
+    this.setState({ value: e.currentTarget.value });
+  };
+
+  onFocus = (e: SyntheticMouseEvent<HTMLInputElement>) => {
+    this.setState({ active: true });
```
julianguyen: I don't think we need to pass in the event if we're not gonna use it.

client/app/bundles/shared/components/Textarea.jsx
```diff
@@ -0,0 +1,91 @@
...
+  };
+
+  onBlur = (e: SyntheticMouseEvent<HTMLInputElement>) => {
+    this.setState({ active: false });
```
julianguyen: Same thing here about the event argument.

client/app/bundles/shared/components/Textarea.scss
```diff
@@ -0,0 +1,52 @@
...
+$focus-background-color-darken: rgba(0, 0, 0, 0.4);
+$focus-background-color-lighten: lighten($container-color, 3%);
+$light-font-weight: 300;
+
+.label {
+  color: $container-color;
+  font-family: $font-family;
+  font-size: $small-font;
+  padding: 12px 4px;
+  letter-spacing: 0.0944444em;
```
julianguyen: Should this be in px?

HeyStevenXu: I think you left a similar comment before, and this was nshki's response: https://github.com/ifmeorg/ifme/pull/792/files/bddb09bfe3dba6058c8b07fe5fc3a3c441ec3f4b#diff-8c6a9053fe95ea482db2fa6012fd737e

I ended up assuming that ems were okay. But if not, let me know!

client/app/bundles/shared/components/Textarea.jsx
HeyStevenXu: Thanks for point that out! Removed!

nshki:
Just to reference this with the issue, this PR is for #781.

client/app/bundles/shared/components/Textarea.scss
julianguyen: Oh true! Okie dokes, em is fine then :P Should we round up the value though @nshki?

nshki: Can't hurt to round to 0.095!

HeyStevenXu: Haha sure thing! Adjusted the line in Input.scss as well for good measure :)

PR #828, opened 2018-02-22
Add Filipino Translation

vinzruzell:


vinzruzell:
May I ask what is the problem on this?

vinzruzell:
Pls help me with this.

cartothemax:
Hey @vinzruzell!

So the problem here is that the locale you added to the configuration is not showing up in the `user.rb` model https://github.com/ifmeorg/ifme/blob/master/app/models/user.rb#L75

This test https://github.com/ifmeorg/ifme/blob/master/spec/controllers/locales_controller_spec.rb actually loops through all of the locales in the configuration (the file that you added `fl` to) and does a comparison to make sure that all of the locales in the configuration are showing up in the `user.rb` model.

In this pull request https://github.com/ifmeorg/ifme/pull/825/files you have the fix made that will correct this failure.

Would it be possible to combine all of your current open pull requests for this translation into one? Thanks!

Issue #783, opened 2018-01-09
[Component] Resource


nshki:
## Overview

We are looking for an awesome human to build this out as a React component, preferably within Storybook. You can view and get details (colors, dimensions, export SVGs, etc.) by visiting [the Figma designs](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618) and creating a free account.

This is part of the app redesign (#691). This also will incorporate another component #782.

## Screenshot from [the Figma document](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618):

![screenshot 2018-01-09 at 6 24 47 pm](https://user-images.githubusercontent.com/1121087/34748419-6732c8f4-f56a-11e7-869a-90258ef1ee50.png)

Seunadex:
@nshki @julianguyen Kindly give more clarity to this, i seems not to fully get how it should be?

nshki:
Just as a public FYI, we had a discussion regarding the above question over Slack. The Figma mockup of the resources page was referenced for context.

baohouse:
@nshki Is it the case that the tags and the author are optional, and that it's possible to have:
1. both visible
2. tags visible only
3. author visible only
4. neither visible

Seunadex:
From what I'm aware of, it is either the tags or the author which is the only difference between the two components, @nshki can do more justice to that.

baohouse:
@nshki For this component, is author and tags mutually exclusive, or can both exist simultaneously?

nshki:
@baohouse In the designs I only used the tags for external resources and authors for internal ones, but it makes sense to be able to have both tags and authors co-exist. Sorry about the late reply, by the way!

Seunadex:
This is completed and can be closed now

PR #853, opened 2018-03-14
Resource component

Seunadex:
#### What does this PR do?
It adds the resources component as specified in this [issue](https://github.com/ifmeorg/ifme/issues/783)


client/app/bundles/shared/components/Resource/Resources.jsx
```diff
@@ -0,0 +1,31 @@
+// @flow
+import React from 'react';
+import css from './Resources.scss';
+import Tag from '../Tag';
+
+type Props = {
+  resourcesObj?: object,
```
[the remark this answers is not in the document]
Seunadex: Yes, this object contains all the props that will be passed to individual resource.

client/app/bundles/shared/components/Resource/Resources.jsx
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react';
+import css from './Resources.scss';
+import Tag from '../Tag';
+
+type Props = {
+  resourcesObj?: Object,
+  tagged?: boolean,
+};
+
+export default class Resources extends React.Component<Props> {
```
nshki: Oooh, sorry I didn't catch this before, but can we rename this to `Resource`? Just singular, not plural.

client/app/bundles/shared/components/Resource/Resources.jsx
```diff
@@ -0,0 +1,31 @@
...
+  constructor(props: Props) {
+    super(props);
+  }
+
+  render() {
+    const { resourcesObj, tagged } = this.props;
+    const resources = tagged ? resourcesObj.tags.map((tag, index) =>
+      <Tag normal label={tag} key={index} />,
```
nshki: I know this is not the same component used, but for `<Tag normal />`s, the background color is now off for some reason. Could you tweak it so that it is `rgba(255, 255, 255, 0.1)`?

Seunadex:
@nshki Apologies for not notifying before closing the PR, I've implemented both your old and new comments except the one with `& + & {}` , I've gotten the concept of how it works but somehow it not working on a div, thats the only feedback left for me to implement.

Seunadex:
@nshki @julianguyen All comments have been addressed , just trying to make sure the tests passes.

client/app/bundles/shared/components/Resource.jsx
```diff
@@ -0,0 +1,30 @@
...
+  tagged?: boolean,
+  link?: string,
+  name?: string,
+  author?: string,
+};
+
+export default class Resource extends React.Component<Props> {
+  render() {
+    const { tags, link, name, author, tagged } = this.props;
+    const resources = tagged ? tags.map(tag =>
```
baohouse: This looks like two different components. One could be AuthorResource and the other TagResource. I would advise against having two different components inside a single component name, especially if they have different props requirements.

Seunadex:
@baohouse I have addressed your comments but kindly shed more light on why you think having it separated is better.

baohouse:
@Seunadex I apologize, I was looking purely at the code, and didn't understand the feature. I'm asking @nshki to clarify the feature. It may very well be the case that we collapse it back to one Component, with tags and author being able to exist simultaneously, or neither at all.

Generally a good coding practice I recommend when writing React components is to have "dumb" components and "smart" components. Dumb components are relatively simple; they do not have much logic or state to worry about. These are easy to write unit tests for. Smart components, however, have to deal with logic and state; the tests for them are more involving. It is helpful to treat Smart components as containers for several dumb components. A good example is a ResponsiveLayout smart component which either displays, based on the browser with, the WideLayout (1600px) simple component, or the MobileLayout (width 480px) simple component.

Seunadex:
@baohouse have you gotten any clarity yet? kindly share, and its back to just one component.

baohouse:
@Seunadex So looks like we should support the ability to display both tags and author simultaneously.

Seunadex:
screen shot 2018-03-28 at 10 54 45 am

@baohouse Just for clarification, are you saying . we can have the resource like this ☝️ or the way I did it before where specifying `tagged` differentiates between external or internal like this 👇 ?


screen shot 2018-03-28 at 10 58 41 am

screen shot 2018-03-28 at 10 58 50 am

baohouse:
@Seunadex The first example is correct. A little spacing between the author and tags, and some Karma tests, and this PR should be ready to go. Thank you sir!

client/app/bundles/shared/components/Resource.scss
```diff
@@ -0,0 +1,38 @@
...
+    padding-bottom: 10px;
+  }
+  
+  div > a {
+    text-decoration: none;
+    color: $header-text-color;
+  }
+
+  .author {
+    color: $author-text-color;
```
baohouse: The author text color should be 50% opacity. Maybe use the **rgba** sass function.

baohouse:
@Seunadex 

1. Can you create another Storybook story for displaying a resource that shows both author and tags?
2. I made a comment regarding the author text color
3. We should have tests written for our components. If you need help with this, let me know.
4. I'm attaching an image of the component when the width shrinks. There are some minor styling issues.
  a. The title line-height should be tighter
  b. For the tags, when it wraps to the second line, it should line up against the left edge as the first line.

screen shot 2018-04-04 at 4 30 06 pm

baohouse:
@Seunadex 

1. Updated the storybook to display the component in several scenarios.
2. Added automation tests. I tested pretty much every combination of props. I would recommend we rely on fewer props; writing automation tests for every combination gets to be pretty tedious. For example, instead of checking the "external" flag, just pass in the "author" text and render if available. Anyway, I didn't want to change the props right now.

screen shot 2018-04-06 at 6 03 53 pm
screen shot 2018-04-06 at 6 03 58 pm
screen shot 2018-04-06 at 6 04 04 pm

baohouse:
@nshki Just need your blessing now.

Seunadex:
@baohouse nice work

Issue #846, opened 2018-03-10
[Component] Heading(s)


nshki:
## Overview

We are looking for an awesome human to build this out as a React component, preferably within Storybook. You can view and get details (colors, dimensions, export SVGs, etc.) by visiting [the Figma designs](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618) and creating a free account.

This is part of the app redesign (#691).

## Screenshot from [the Figma document](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618):

![screenshot 2018-03-10 at 12 05 24 pm](https://user-images.githubusercontent.com/1121087/37244780-d170b5a8-245b-11e8-88f8-94c196c3df05.png)

aSquare14:
I'd like to claim this issue :)

nshki:
All yours @aSquare14!

PR #855, opened 2018-03-15
Headings created

aSquare14:
issue #846 resolved.
Created Headings .

### DIFF
--- a/client/app/bundles/shared/components/Heading.jsx
+++ b/client/app/bundles/shared/components/Heading.jsx
@@ -1,0 +1,23 @@
+//@flow
+import React from 'react';
+import css from './Heading.scss';
+
+type Props = {
+  large?: boolean,
+  small?: boolean,
+  text?: boolean,
+  label?: string,
+};
+
+export default class Heading extends React.Component<Props> {
+  render() {
+    const { large, small, text, label } = this.props;
+    const labelClassNames = `${css.label} ${large ? css.large : ''} ${small ? css.small : ''} ${text ? css.text : ''}`;
+
+    return (
+      <span>
+        <div className={labelClassNames}>{label}</div>
+      </span>
+    );
+  }
+}
--- a/client/app/bundles/shared/components/Heading.jsx
+++ b/client/app/bundles/shared/components/Heading.jsx
@@ -1,4 +1,4 @@
-//@flow
+// @flow
 import React from 'react';
 import css from './Heading.scss';
 
--- a/client/app/bundles/shared/components/Heading.jsx
+++ b/client/app/bundles/shared/components/Heading.jsx
@@ -15,7 +15,9 @@ export default class Heading extends React.Component<Props> {
     const labelClassNames = `${css.label} ${large ? css.large : ''} ${small ? css.small : ''} ${text ? css.text : ''}`;
 
     return (
-        <div className={labelClassNames}>{label}</div>
+      <div className={labelClassNames}>
+        {label}
+      </div>
     );
   }
 }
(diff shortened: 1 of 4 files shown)


client/app/bundles/shared/components/Heading.jsx
```diff
@@ -0,0 +1,23 @@
+// @flow
+import React from 'react';
+import css from './Heading.scss';
+
+type Props = {
+  large?: boolean,
```
julianguyen: Should all these fields be mandatory rather than optional?

aSquare14: I think mandatory as then user can use whatever he requires based on props passed. @nshki What are your views on this ?

nshki: I think it makes sense to make the `label` prop mandatory, but the rest can stay optional.

client/app/bundles/shared/components/Heading.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default class Heading extends React.Component<Props> {
+  render() {
+    const { large, small, text, label } = this.props;
+    const labelClassNames = `${css.label} ${large ? css.large : ''} ${small ? css.small : ''} ${text ? css.text : ''}`;
+
+    return (
+      <span>
+        <div className={labelClassNames}>{label}</div>
+      </span>
```
nshki: Instead of having two layers of elements, let's just keep it at one. I think a heading element would make sense here!

client/app/bundles/shared/components/Heading.jsx
aSquare14: I don't know how to make them optional. What do I have to do ?

client/app/bundles/shared/components/Heading.jsx
aSquare14: Which headings should I use. I am thinking
h1 - for large
h2 - for normal
h3- for small
h4- for text

client/app/bundles/shared/components/Heading.jsx
julianguyen: You used `?` so that should be enough :)

client/app/bundles/shared/components/Heading.jsx
julianguyen: I personally think we should be using divs that way if we change the hierarchy of headings we don't need to refactor all of H level headings. I think a div without a parent span should be enough!

aSquare14: okay ! Cool !

client/app/bundles/shared/components/Heading.jsx
aSquare14: Okay :)

aSquare14:
@julianguyen  Why are the tests failing ? It is showing Rspec problem. How do I fix it ?

julianguyen:
Looks like flakey tests. Argh. I'll look for a way this weekend to reduce the flakiness. Sorry about that!

aSquare14:
Okay ! :)

aSquare14:
Thanks a lot @julianguyen ! Your kind words keep me and Prateksha motivated :heart:  I'll merge it after the tests ?

julianguyen:
Yep yep! It should go green 🍀

aSquare14:
@julianguyen  Could you trigger it again ? I do not have write permissions and hence cant trigger it !

julianguyen:
Looks like there are legitimate issues this time around. Scroll down the page to where the red box is highlighted for more details! Let me know if you need help. I'll be away from internet for an hour or so though.

aSquare14:
@julianguyen  Can you please approve again ? The tests finally passed :)

Issue #795, opened 2018-02-04
Incorporating facebook


aSquare14:
# Description

- We can write a general sessions controller function and integrate login with Facebook.
- It is better to have both options as some people prefer using facebook login and will be motivated to share about If-me on their facebook.

- Also,we can give the user an option of sending their facebook friends invites to the website in "Allies" section.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

aSquare14:
presently @steffanibrasil and I are working on it.

baohouse:
Still need to add if-me.org to the domain whitelist in the Facebook app. @stefannibrasil has control over it I believe.

baohouse:
We did a merge into production, but I had to disable and revert the PR #856 because of an authorization flaw in which logging in gave you access to other people's posts. We'll need to investgate the cause of that and perhaps use a staging environment for testing.

baohouse:
If there's anyone who can look into the authorization issue, as I don't have much experience in the whole backend architecture.

julianguyen:
Holding off for now since Facebook recently faced security vulnerabilities with their oauth service.

stefannibrasil:
hi, all! I want to get back to contribute to ifme. Do you think I can continue on this? Thank you :)

julianguyen:
@stefannibrasil Feel free to! That would be great, I would just watch out for any security issues.

stefannibrasil:
great, I'll get back to this, then, thank you, @julianguyen :)

julianguyen:
OAuth service issues have been resolved! This issue is up for grabs :)

PR #856, opened 2018-03-16
Add facebook authentication

stefannibrasil:
Add Facebook login as an alternative to Google account.

This is the first part of incorporating Facebook as @aSquare14 suggested on #795 

I was able to do some refactoring but would appreciate more tips on that.


julianguyen:
Hey @stefannibrasil can you try rebasing? It looks like there's a bunch of conflicts -- not sure if they are contributing to the issues you're encountering. Regardless, you should rebase :P

config/env/development.example.env
```diff
@@ -12,6 +12,10 @@ GOOGLE_CLIENT_ID=""
 GOOGLE_CLIENT_SECRET=""
 GOOGLE_APP_NAME=""
 
+# Configuration values for Oauth through Facebook
```
julianguyen: We will have to setup a way for folks with the existing codebase to update their env files. This can either be done manually by telling folks on Slack/documentation or we can write an automated script for this.

config/initializers/devise.rb
```diff
@@ -317,4 +317,17 @@
...
+      :facebook,
+      ENV['FACEBOOK_APP_ID'],
+      ENV['FACEBOOK_APP_SECRET'],
+      scope: 'public_profile,email',
+      info_fields: 'email,first_name,last_name',
+      client_options: {
+         site: 'https://graph.facebook.com/v2.11',
+         authorize_url: "https://www.facebook.com/v2.11/dialog/oauth"
+      },
+      :strategy_class => OmniAuth::Strategies::Facebook
```
julianguyen: use `strategy_class: OmniAuth::Strategies::Facebook` instead

config/initializers/devise.rb
```diff
@@ -317,4 +317,17 @@
         scope: 'userinfo.email,userinfo.profile,calendar'
       }
     )
+
+    config.omniauth(
+      :facebook,
+      ENV['FACEBOOK_APP_ID'],
+      ENV['FACEBOOK_APP_SECRET'],
+      scope: 'public_profile,email',
+      info_fields: 'email,first_name,last_name',
```
julianguyen: Just double-checking here, should this be a string or an array?

spec/models/user_spec.rb
```diff
@@ -44,45 +44,60 @@
 describe User do
```
julianguyen: I would add a test for when omniauth is not successful and throws an error

app/controllers/registrations_controller.rb
```diff
@@ -6,7 +6,7 @@ def after_update_path_for(_resource)
   end
 
   def update_resource(resource, params)
-    if current_user.provider == 'google_oauth2'
+    if current_user.provider == 'google_oauth2' || current_user.provider == 'facebook'
```
cartothemax: The conditional for this statement is getting a little too long. You could move `current_user.provider == 'google_oauth2' || current_user.provider == 'facebook'` into it's own private method within this controller.

Here is a great article that details the pattern of refactoring conditionals -> http://designisrefactoring.com/2015/03/09/refactoring-away-a-conditional/

config/env/development.example.env
stefannibrasil: Good point. I am gonna see how to make a scrip for this!

config/initializers/devise.rb
stefannibrasil: From the [Devise documentation](https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview), it should be a string.

app/controllers/registrations_controller.rb
stefannibrasil: Thanks, awesome article! :+1

spec/models/user_spec.rb
stefannibrasil: Yes, I was thinking about that too. Gonna work on that.

app/views/devise/registrations/edit.html.erb
```diff
@@ -11,16 +11,20 @@
...
-    <div class="label">
-      <%= f.label t('common.form.email') %>
-    </div>
-    <% if current_user.provider == 'google_oauth2' %>
-      <div class="small_font small_margin_bottom">
-        <%= raw t('devise.registrations.google_auth1') %>
-      </div>
-    <% end %>
-
+	<div class="field">
```
julianguyen: Let's fix the indentation here :) Should be two spaces a tab!

app/views/devise/shared/_links.erb
```diff
@@ -1,12 +1,19 @@
...
+          <%= button_to t('devise.shared.sign_in_google'), user_google_oauth2_omniauth_authorize_path %>
+        <% elsif provider.to_s == "facebook"  %>
+          <%= button_to t('devise.shared.sign_in_facebook'), user_facebook_omniauth_authorize_path %>
+        <% end %>
       <% elsif provider.to_s == "google_oauth2" %>
-        <%= button_to t('devise.shared.sign_up_google'), omniauth_authorize_path(resource_name, provider) %>
-      <% end -%>
-    <% end -%>
+          <%= button_to t('devise.shared.sign_up_google'), user_google_oauth2_omniauth_authorize_path %>
+      <% else provider.to_s == "facebook" %>
```
julianguyen: Should this be elsif too?

stefannibrasil: Good point, you're right. Thanks, @julianguyen. I'll also create a new Issue for the env vars script, just in case someone can work on it before me.

baohouse:
@stefannibrasil Last test to fix is that the Vietnamese language files need to be updated to include the new Facebook-related text.

PR #857, opened 2018-03-16
Add my name (George Karametas) to contributors

GeoDoo:


baohouse:
I would recommend cropping the image to remove the transparent padding.

PR #858, opened 2018-03-16
Splitting up Storybook stories into separate files

baohouse:


.codeclimate.yml
```diff
@@ -22,7 +22,7 @@ engines:
     - vendor/assets/javascripts
     - client/karma.conf.js
   fixme:
-    enabled: true
+    enabled: false
```
julianguyen: Why is it false now?

baohouse: Having TODO in the code is not necessarily something that should invoke a CodeClimate error whose resolution is "Must be Fixed Immediately".

baohouse:
Fixed some tests and merge conflicts @julianguyen!

PR #864, opened 2018-03-21
[WIP] Splitting up Logo file; Adding Ant Design

baohouse:
Previously the Logo would resize using media queries. But it's best to have the logo behave exactly the same regardless of the viewport width, and to compose several sizes into a responsive component.

Also this was a chance to figure out how to code up a component factory pattern with Flow. After a lot of trial and error, I think I figured it out. The flow parser seems to have problems when you have too many arrow functions (curried functions) strung in succession. Either that or there _is_ a way to do it, and I haven't figured it out yet.

### DIFF
--- a/doc/pages/contributors.json
+++ b/doc/pages/contributors.json
@@ -258,5 +258,9 @@
   {
     "name": "George Karametas",
     "link": "https://geodoo.work"
+  },
+  {
+    "name": "Annette Alt",
+    "link": "https://homestylecookingwithsylvya.wordpress.com/"
   }
 ]


baohouse:
Adding Ant Design to this as well.

client/.babelrc
```diff
@@ -1,15 +1,27 @@
...
   "env": {
-    "test": {
+    "development": {
       "plugins": [
-        "transform-es2015-modules-commonjs"
+        "flow-react-proptypes",
+        [
+          "import",
+          {
+            "libraryName": "antd",
```
julianguyen: Are we going to be using Ant for all of our components now? Just wondering what the motivation is here.

julianguyen: Ah I see in our #react channel. So are we just using Ant for its grid system? If so, is there a way to just import that instead of the whole framework?

baohouse: @julianguyen You will be pleased to know that using [babel-plugin-import](https://github.com/ant-design/babel-plugin-import) (which is maintained by the ant team well) will load only the components you imported into your code, not the whole framework.

I think to answer the question of whether all of our components will use Ant, I say if any of the volunteers want to override the style of an Ant component, they have the option to now. And it won't unnecessarily bloat the code, so we have the flexibility to incrementally bring in Ant components as needed.

baohouse:
@julianguyen I fixed some webpack build issues that existed prior to this PR. By upgrading UglifyJS, I was able to get the production JS bundle cut from 483 kb gzipped to 244 kb!!! Woot!

baohouse:
I'm going to need to downgrade AntD to one that is compatible with React 15, because we can't upgrade React to 16; it would break tests.

Issue #874, opened 2018-03-31
Implement lazy-loading Avatar component


baohouse:
- [x] Enhance Avatar React component (#849) with [react-lazyload](https://github.com/jasonslyvia/react-lazyload)
- [x] Implement Avatar React component into _if me_ application
- [x] Observe in production that uploaded files are correctly pulled from Cloudinary
- [x] Observe in production that local files are correctly pulled from Cloudinary

PR #876, opened 2018-04-01
Refactored ProfilePicture to use React component

baohouse:
Addresses #874 

Change overview:
1. Updated ProfilePicture class to use Avatar React component.
2. Enhanced Avatar component to use react-lazyload
3. Updated tests that assert on the ProfilePicture
4. Fixed some tests for language toggling involving page transitions

PR checklist:
- [x] yarn test
- [x] yarn lint
- [x] rspec
- [x] rubocop
- [x] Manual inspection of _if me_ homepage for visual parity
- [x] Manual check for lazy loading of images to make sure the images do not load until the user scrolls them into view
- [x] Manual check that image size on _if me_ homepage is correct in both desktop and mobile widths
- [x] Manual inspection on _if me_ profile page for visual parity


client/app/bundles/shared/components/_colors.scss
```diff
@@ -0,0 +1,11 @@
+/**
+ * This file contains the color palette for the Kujali interface.
+ */
+
+$theme-color-1: #6d0839;
```
julianguyen: Rename as `$theme-color-primary`? Really though it should be colour 🌈

client/app/bundles/shared/components/_colors.scss
```diff
@@ -0,0 +1,11 @@
...
+$theme-color-2: #d0e799;
```
julianguyen: Rename as `$theme-color-secondary`?

client/app/stories/Avatar.stories.jsx
```diff
@@ -1,21 +1,38 @@
...
-  .add('With name', withInfo({})(() =>
-    (<Avatar
-      src={photo}
-      alt="Julia Nguyen"
-      name="Julia Nguyen"
-    />),
+  .add('With name', withSource(
+    <Row style={{ backgroundColor: '#aaa' }}>
+      <Col span={8}>
+        <Avatar src={photoJulia} name="Julia &ldquo;Fleurchild&rdquo; Nguyen" displayname />
```
julianguyen: LOL 🌷👧

client/karma.conf.js
```diff
@@ -20,16 +20,23 @@ webpackConfig.externals = {
   'react-addons-test-utils': 'react-dom',
 };
 
+function getSpecs(specList) {
+  if (specList) {
+    return specList.split(',');
+  } else {
+    return [
+      { pattern: 'app/**/*.spec.js', watched: true },
```
julianguyen: Hmm not 100% sure why it's broken up like this, can you explain? Thanks!

app/assets/stylesheets/application.scss
```diff
@@ -593,22 +586,30 @@ img {
 }
 
 .small_profile_picture_div {
-  width: 100px;
-  height: 100px;
+  width: 1px;
```
julianguyen: Why is this 1px?

app/assets/stylesheets/application.scss
```diff
@@ -593,22 +586,30 @@ img {
...
+      width: 50px;
+      height: 50px;
+    }
   }
 }
 
 .profile_picture_div {
-  width: 150px;
-  height: 150px;
+  width: 1px;
```
julianguyen: Similar question as above ^

client/app/stories/Avatar.stories.jsx
baohouse: Haha, I'm just testing to see what the component looks like when the line length exceeds the width of the image. Honest!

client/app/bundles/shared/components/_colors.scss
baohouse: We can call it primary, that's fine too. It's possible in the future that a theme has more than 2 colors, so that's why I just used straight up numbers. In the future, you can then assign variables:
```
$text-highlight: $theme-color-secondary;

mark { color: $text-highlight }
```

"Colour"? Haha, I would say CSS use of "color" makes things more consistent, but, if you really want, I can make the change. 😆

client/karma.conf.js
baohouse: In the ```client/README.md``` I documented a way to run specific tests. I looked up on Stackoverflow on how to do this, and this was the best answer I could find so far. So if you specify a list in the environment properties, it will run those, otherwise falling back to all tests.

app/assets/stylesheets/application.scss
baohouse: This is a div element made to behave like a table cell. Since we want to control the width by way of the content's width (which in this case is the Avatar component), we can reduce the number of places to update width information from two places to one.

app/assets/stylesheets/application.scss
baohouse: Ditto as above 😄

client/app/bundles/shared/components/_colors.scss
julianguyen: I think if there's more than two colours in the future, we should just give them straight up names

client/webpack.config.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * TODO: Split up development configuration from production configuration
```
nma: it is a bit more involved than that sadly, we will need to move react_on_rails to v9 and bump us up to webpacker instead of webpacker_lite gem

client/webpack.config.js
```diff
@@ -35,11 +40,13 @@ const config = Object.assign(baseConfig, {
   context: resolve(__dirname),
 
   entry: {
-    'webpack-bundle': [
+    'moments-app_bundle': glob.sync('./app/bundles/momentsApp/startup/*'),
```
nma: will we be separating bundles for momentsApp / alliesApp and etc going forward? Or only moments app due to the chart.js requirements?

app/views/layouts/application.html.erb
```diff
@@ -9,9 +9,17 @@
...
 
-    <!-- Webpack Assets -->
-    <%= javascript_pack_tag 'webpack-bundle' %>
-    <%= stylesheet_pack_tag 'webpack-bundle' %>
+    <!-- Webpack Assets START -->
+    <%= javascript_pack_tag 'shared_bundle' %>
+    <%= stylesheet_pack_tag 'shared_bundle' %>
+
+    <!-- TODO: Need a way to load this bundle only on Moments pages -->
+    <%= javascript_pack_tag 'moments-app_bundle' %>
```
nma: my thoughts here is that once we move the entirety of the app to React, we can just inject one bundle. This feels like a  transition code atm.

client/webpack.config.js
baohouse: We should let webpack dynamically determine chunking. If we use UI routing, then this becomes far more feasible. Right now we're manually creating chunk entries, but that's not scalable.

PR #880, opened 2018-04-06
Miscellaneous cleaning / optimizations

baohouse:


client/.babelrc
```diff
@@ -18,6 +18,7 @@
     }
   },
   "plugins": [
+    "lodash",
```
julianguyen: Ooof do we really need lodash? I would rather not add on bulky libraries and depend on ES6.

client/app/bundles/momentsApp/components/Chart/ChartControl.jsx
```diff
@@ -57,7 +57,7 @@ export default class ChartControl extends React.Component<chartControlProp, char
   render() {
     const { types } = this.props;
     const buttons: any[] = [];
-    _.each(types, (value: string, index: number) => {
+    each(types, (value: string, index: number) => {
```
julianguyen: We could completely ditch lodash here and use a for...of loop or forEach.

baohouse: Haha, yah, we could. If we ever need lodash in the future, the babel-plugin-lodash treeshakes the lodash bundle, so only the methods you use gets imported, provided you import them as 
```
import { each } from 'lodash';
```
instead of 
```
import _ from 'lodash'
```

But so far I've only found one usage for lodash. We can remove it for now.

nma: That's actually good to know that we have that option.

baohouse:
@nma Julia was thinking of bundle size mainly. Lodash is pretty large, and I saw it too when I did a bundle breakdown report; it was the largest module. So I introduce babel-plugin-lodash to do reduce the size. In the end, we only use it in one place, and we only needed to use Array.map rather than _.each.

julianguyen:
@nma I think lodash is helpful when we didn't have access to useful functions in vanilla JS. ES6+ has added a lot of these functions and we should use them over the lodash equivalent. I think it's better to reduce the number of abstractions and stick as close as possible to JS.

client/.babelrc
julianguyen: I think we can get rid of this now? If not, we can address this in another PR!

client/app/bundles/momentsApp/components/MomentCard/MomentCardSettings.jsx
```diff
@@ -8,21 +8,23 @@ type MomentCardSettingsProp = {
...
       <div className={css.settings}>
-        <i role="presentation" onClick={this.props.action.viewer} className={`fa fa-lock ${css.action}`} />
         {
-          this.props.cardType !== 'Example' &&
-          <span>
-            <i role="presentation" onClick={this.props.action.delete} className={`fa fa-trash-o ${css.action}`} />
-            <i role="presentation" onClick={this.props.action.edit} className={`fa fa-pencil ${css.action}`} />
-          </span>
+          ['edit', 'delete', 'viewer'].map(op => action[op] &&
+            <i role="presentation" onClick={action[op]} className={`fa ${classMap[op]} ${css.action}`} />,
```
dhaval0894: This should have a "key" prop.

dhaval0894:
Thanks a lot, @baohouse for the code refactoring, cleaning. Just that small change, everything else looks great!!

PR #881, opened 2018-04-09
Vietnamese Translations

julianguyen:
Vietnamese translations! Huge thanks to Van-Anh Le for doing these wonderful translations 💜 I learned so much. Cảm ơn nhiều!

- [x] vi.yml
- [x] devise.vi.yml
- [x] devise_invitable.vi.yml
- [x] kaminari.vi.yml
- [x] Developer steps


baohouse:
We should use ISO-639-1 language code, so it would be "vi" not "vn". This might be the reason why the tests are failing.

```
// Instead of this
import vi from 'react-intl/locale-data/vn';
// This is probably the correct way
import vi from 'react-intl/locale-data/vi';
```

client/app/libs/i18n/I18nSetup.js
```diff
@@ -14,6 +14,7 @@ import it from 'react-intl/locale-data/it';
 import nb from 'react-intl/locale-data/nb';
 import sv from 'react-intl/locale-data/sv';
 import nl from 'react-intl/locale-data/nl';
+import vn from 'react-intl/locale-data/vn';
```
baohouse: Try ```import vn from 'react-intl/locale-data/vi';```

julianguyen: @baohouse Omg thanks, I didn't realized vi was the official code and not vn lol

PR #882, opened 2018-04-09
Upgrade our usage of the react_on_rails gem from v8 to v9.

nma:
Getting us onto the new react_on_rails doesn't seems like a hard problem. Managed to fix all the compat issues and have a working solution.

The jump from v9 - v10 is just a gem and npm module version change, and we will be able to use the HMR settings from their sample after that. 

Will be testing out HMR solution in a followup PR.


nma:
So CSS bundles are working for me now locally. I'm also going to need to merge from master / do a raw checkout  of my branch to verify and it should be good to go.

The fix was that `webpacker` gem is version-locked to a specific version of `webpack`. That was why our .css bundles were not coming out. 

**Pending Tasks:**
- [ ] Merge Master
- [ ] Raw checkout / rebuild and test

Issue #885, opened 2018-04-13
Use IETF language tag throughout code


baohouse:
I noticed **ptbr** being used, which should be **pt-BR** according to IETF language code.
https://en.wikipedia.org/wiki/IETF_language_tag

We should use well-known standards to refer to languages, especially in the future when we need to support locale fallbacks.

Part of this refactoring could also be to have the `config/locales` folder split up into their respective IETF subfolders, so you would have `config/locales/en-US` and `config/locales/pt-BR`.

baohouse:
<img width="262" alt="screen shot 2018-04-17 at 7 41 56 am" src="https://user-images.githubusercontent.com/4228012/38877208-06406704-4213-11e8-81a0-b9fbfc880bb4.png">

This seems to be the case when the I18N export to JavaScript occurs, though we have some inconsistency. Notice pt for Portuguese. There's the non-standard "ptbr" versus IETF "pt-BR".

PR #888, opened 2018-04-18
i18n Refactor

baohouse:
* Cleaning up Dropdown for general usage
* Enhancing Storybook infrastructure
* Fixing naming of Brazilian Portugese IETF language tag (#885)


baohouse:
screen shot 2018-04-18 at 8 55 21 am
screen shot 2018-04-18 at 8 55 36 am

Now with background-color selection and locale selection built into every story.

client/.storybook/addons.js
```diff
@@ -1,4 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies, import/no-unresolved, import/extensions */
 
-import '@storybook/addon-actions/register';
+// import '@storybook/addon-actions/register';
```
julianguyen: Why not remove this completely instead of commenting it out?

client/app/bundles/shared/components/Dropdown/Dropdown.scss
```diff
@@ -1,52 +1,49 @@
...
+  $border-radius: $rounded-corner,
+  $border: none,
+  $appearance: none
+) {
+  padding: $padding;
+  font-size: $font-size;
+  color: $color;
+  border: $border;
+  border-radius: $border-radius;
+  /** TODO: Use PostCSS to add vendor prefixes **/
```
julianguyen: Nice!

client/app/bundles/shared/components/Dropdown/DropdownFactory.jsx
```diff
@@ -0,0 +1,36 @@
...
+  onChange?: (event: Event) => any,
+  options: Option[],
+  value?: any,
+};
+
+const DropdownFactory = (variationClassName: string) => {
+  const DropdownComponent = ({ onChange, options, value: propValue }: Props) => (
+    <select className={`${css.dropdown} ${variationClassName}`} onChange={onChange} value={propValue}>
+      {options.map(({ key, label, value }: Option) => (
+        <option key={key || (value !== null && value !== undefined ? value : label)} value={value}>
```
julianguyen: Try replacing `(value !== null && value !== undefined ? value : label)` with `value || label`

client/app/bundles/shared/components/Dropdown/DropdownFactory.jsx
```diff
@@ -0,0 +1,36 @@
...
+          {label}
+        </option>
+      ))}
+    </select>
+  );
+  DropdownComponent.defaultProps = {
+    onChange: () => {},
+    options: [],
+    value: undefined,
```
julianguyen: I think you can just exclude line 30 and value will be implicitly undefined.

client/app/bundles/shared/components/Footer/Footer.scss
```diff
@@ -3,9 +3,9 @@
 
 .footer {
   width: 100%;
-  background: rgba(0, 0, 0, 0.5);
   font-family: $font-family;
   position: relative;
+  background-color: rgba(0, 0, 0, 0.5);
```
julianguyen: Should this colour be a variable? 🤔

client/app/libs/i18n/I18nSetup.js
```diff
@@ -3,24 +3,21 @@ import { addLocaleData } from 'react-intl';
...
-import br from 'react-intl/locale-data/br';
-
-const ptbr = br[0];
-ptbr.locale = 'ptbr';
 
 // Initizalize all locales for react-intl.
 // eslint-disable-next-line import/prefer-default-export
 export const loadLocales = () => {
-  addLocaleData([...en, ...es, ...it, ...nb, ...sv, ...nl, ...vi, ptbr]);
+  addLocaleData([].concat(en, es, it, nb, nl, pt, sv, vi));
```
julianguyen: Ooh I like this a lot better! :)

client/app/stories/stories.scss
```diff
@@ -0,0 +1,5 @@
+body {
+  background: none;
+  padding: 20px;
+  height: auto;
```
julianguyen: Shouldn't the values for background and height be those values by default?

client/app/bundles/shared/components/Dropdown/DropdownFactory.jsx
baohouse: Actually I should just use id generator when key is not provided. It's possible for two options to have same value.

client/app/stories/stories.scss
baohouse: Storybook has its own CSS styles, so we have to override them to achieve things like configurable background color. I think padding is just there to give space so that our components don't collide with the edge of the browser.

client/app/bundles/shared/components/Dropdown/DropdownFactory.jsx
baohouse: I'll get back to you on this... I think I had to include it for lints to pass. But I'll double check.

client/.storybook/addons.js
baohouse: I'm gonna reinstate this, after reading the docs, I now know what this is for. I'll wire up its usage in all the stories.

client/app/bundles/momentsApp/components/Chart/Chart.jsx
```diff
@@ -17,30 +17,19 @@ const colorSchemes = ['#6D0839', '#66118', '#7F503F', '#775577', '#CCAADD'];
...
-              ytitle={this.props.ytitle}
-              data={this.props.data}
-              colors={colorSchemes}
-            />}
+const Chart = ({ chartType, ...props }: chartShape) => (
+  chartType === 'Line'
+    ? <LineChart {...props} colors={colorSchemes} />
+    : <AreaChart {...props} colors={colorSchemes} />
+);
+Chart.displayname = 'Chart';
```
nma: Ahh interesting, good to know that there is a way to expose this to enzyme tests.

client/app/bundles/shared/components/Dropdown/Dropdown.scss
baohouse: The lazy bum's ethos: do more with less code.

client/.storybook/addons.js
baohouse: Cool, added it. It does underscore the problem that some of our components did not properly pass along event handlers, so I've marked them as a TODO and will file issues separately to address them, as well as update the component wiki page to make sure this is something developers think about.

baohouse:
@julianguyen Yah, it just needs the approval of a reviewer.

PR #889, opened 2018-04-24
fix(client): add an explicit requirement to rebuild node-sass due to …

nma:
…the node-gyp compilation not building sass bindings properly through yarn

After scouring the internet, it looks like this is a safe solution to let our pipeline trigger the node-sass rebuild. If we run into more of these issues with other node packages, we can reopen this issue and take a look at a scaling solution.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -14,7 +14,8 @@
     "lint": "yarn lint:flow && yarn lint:eslint",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "start": "yarn storybook"
+    "start": "yarn storybook",
+    "postinstall": "npm rebuild node-sass"
   },
   "cacheDirectories": [
     "node_modules",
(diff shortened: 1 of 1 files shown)


client/package.json
```diff
@@ -14,7 +14,8 @@
     "lint": "yarn lint:flow && yarn lint:eslint",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "start": "yarn storybook"
+    "start": "yarn storybook",
+    "heroku-postbuild": "npm rebuild node-sass"
```
julianguyen: Would you call `heroku-postbuild` in the CircleCI config?

nma: Lets use @baohouse  PR. I will close this as they both solve the same problem.

I have my own CIrcleCI to get us on CircleCI 2.0, but its breaking the PR process for my branches sadly. 

So I will close this in favor of @baohouse 's working branch.

nma:
Closing in favor of #890

PR #893, opened 2018-05-14
Revert "Add facebook authentication"

baohouse:
Reverts ifmeorg/ifme#856

Since we don't have a staging environment, tested this in production. While I was able to log in via FB, it gave me access to another person's account in if-me. The Moments were not my own.

We'll need to put the Facebook Authentication PR back into development to fix the authorization issue.


nma:
Looks good to me. We can revert this back to master.

Issue #899, opened 2018-05-28
404 in installation.md


alekpentchev:
Hello!

I have spotted an error in INSTALLATION.md file in PostgreSQL installation section.

1. What about is the error?
After clicking the link to instructions on creating new user (just under the headline) the user sees 404 error.

2. How to reproduce it?
Click on "instructions" link.

2. What is expected?
Link should navigate to an existing page.

Comment:
I have searched for common errors (where the link should navigate to) in the whole repository and could not find any related page.

camillevilla:
hi @alekpentchev! Thanks for reporting this error. That should link to this page:
https://github.com/ifmeorg/ifme/wiki/Common-Dev-Environment-Errors#postgresql-bad-connection

We used to store that information in a `COMMON_ERRORS.md` and moved it over to a wiki page.

Would you be interested in submitting a PR to change the link?

alekpentchev:
Hi! Yes, sure, I will create a PR in a few minutes.

Transvivor:
Has this issue been resolved? If It has which from my investigating it looks like it has. We should resolve it.

PR #903, opened 2018-06-05
Update bad link for PostgreSQL instructions

alekpentchev:
There was a bad link in PostrgreSQL section in INSTALLATION.md. I have changed the link according to instructions from issue #899


camillevilla:
Thanks for contributing to the docs, @alekpentchev!

Issue #780, opened 2018-01-09
[Component] Switch (Checkbox)


nshki:
## Overview

We are looking for an awesome human to build this out as a React component, preferably within Storybook. You can view and get details (colors, dimensions, export SVGs, etc.) by visiting [the Figma designs](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618) and creating a free account.

This is part of the app redesign (#691).

## Screenshot from [the Figma document](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618):

![screenshot 2018-01-09 at 6 20 24 pm](https://user-images.githubusercontent.com/1121087/34748273-c7e96b54-f569-11e7-89d7-761851a49eba.png)

julianguyen:
@ariperez how is this coming along?

ariperez:
Julia...apologize for the delay, I wanted to finish it up before I took off
on vacation last week, but that didn't happen. Working on it now and will
have an update by end of weekend.

aSquare14:
@ariperez  Are you working on this? If not, can I take it up ?

ariperez:
@aSquare14 , I'll go ahead and assign to you.  Took on more than I can handle at this point. :(

aSquare14:
@ariperez No worries. Thanks!  :)

DomitriusAnthony:
Would like to take this on if I can!

aSquare14:
Go ahead!  I'm presently working in something else :)

DomitriusAnthony:
Thanks @aSquare14 ! You're throwing me my first PR :)

prateksha:
@nishki Is anyone working on this? If not, I would like to work on it.

nshki:
@prateksha I'm not quite sure if anyone's actively working on the issue at the moment. If @DomitriusAnthony can confirm within a few days that would be ideal, but if not, you can go ahead!

julianguyen:
I'm going to finish up this feature

aSquare14:
Cool

PR #910, opened 2018-07-03
Switch component added

prateksha:
#780 Resolved!


client/app/bundles/shared/components/Switch.scss
```diff
@@ -0,0 +1,64 @@
...
+  display: inline-block;
+  width: 119px;
+  height: 49px;
+}
+
+.switch input {
+  position: absolute;
+  top: -99999px;
+  left: -99999px;
+}
```
nshki: Ahh, while this works, I think you can get away with just using `display: none;`!

client/app/bundles/shared/components/Switch.scss
```diff
@@ -0,0 +1,64 @@
...
+  line-height: 40px;
+  color: $mulberry-wood;
+  letter-spacing: 9.44%;
+
+  cursor: pointer;
+  top: 0; left: 0; right: 0; bottom: 0;
+  background-color: $mulberry-wood;
+  border: 0px solid rgba(255, 255, 255, 0.7);
+  border-radius: 5px;
+  -webkit-transition: .2s;
```
nshki: You can omit the vendor prefix!

client/app/bundles/shared/components/Switch.scss
```diff
@@ -0,0 +1,64 @@
...
+  position: absolute;
+  content: "OFF";
+  height: 43px;
+  width: 71.4px;
+  left: 2.3px;
+  top: 3.5px;
+  background-color: rgba(255, 255, 255, 0.7);
+  border: 0px solid rgba(255, 255, 255, 0.7);
+  border-radius: 5px;
+  -webkit-transition: .2s;
```
nshki: You can omit the vendor prefix here!

client/app/bundles/shared/components/Switch.jsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+type State = {
+  checked: boolean
+};
+
+export default class Switch extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = { checked: this.props.checked || false };
```
julianguyen: You could do `!!this.props.checked` instead

client/app/bundles/shared/components/Switch.jsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  handleChange = () => {
+    this.setState({ checked: !this.state.checked })
+  };
+
+  render () {
+    const switchClassNames = `${css.switch}`;
```
julianguyen: I don't think you need the back ticks here or in the following line. You should be able to pass in the values directly.

client/app/bundles/shared/components/Switch.scss
```diff
@@ -0,0 +1,64 @@
...
+  top: 0; left: 0; right: 0; bottom: 0;
+  background-color: $mulberry-wood;
+  border: 0px solid rgba(255, 255, 255, 0.7);
+  border-radius: 5px;
+  -webkit-transition: .2s;
+  transition: .2s;
+}
+
+input:checked + .slider {
+  background-color: #BACF88;
```
julianguyen: Let's put this value in a variable! Same goes for value on line 42.

client/app/bundles/shared/components/Switch.scss
```diff
@@ -0,0 +1,64 @@
...
+}
+
+input:focus + .slider {
+  box-shadow: 0 0 1px #BACF88;
+}
+
+.slider:after {
+  position: absolute;
+  content: "OFF";
```
julianguyen: Let's pass in an i18n key here

client/app/bundles/shared/components/Switch.scss
```diff
@@ -0,0 +1,64 @@
...
+  top: 3.5px;
+  background-color: rgba(255, 255, 255, 0.7);
+  border: 0px solid rgba(255, 255, 255, 0.7);
+  border-radius: 5px;
+  -webkit-transition: .2s;
+  transition: .2s;
+}
+
+input:checked + .slider:after {
+  content: 'ON';
```
julianguyen: Let's pass in an i18n key here

client/app/bundles/shared/components/Switch.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+type Props = {
+  id?: string
+};
+
+type State = {
+  checked: boolean
+};
+
+export default class Switch extends React.Component<Props, State> {
```
julianguyen: It would be great to create a unit test for this component and test toggling it.

client/app/bundles/shared/components/Switch.jsx
prateksha: In line 16, why is it 'this.props.checked' and not 'this.state.checked'?

client/app/bundles/shared/components/Switch.scss
prateksha: How do I pass an i18n key here? I'm unable to understand. I tried using attribute.

client/app/bundles/shared/components/Switch.jsx
aSquare14: I've tested this ! Let me know if it is the right way !

client/app/bundles/shared/components/Switch.jsx
julianguyen: Ooh, you're right it should be `!!this.state.checked`

client/app/bundles/shared/components/__tests__/Switch.spec.jsx
```diff
@@ -0,0 +1,17 @@
...
+describe('Switch', () => {
+  it('always renders the switch', () => {
+    let wrapper = shallow(<Switch />);
+    expect(wrapper.find('input').exists()).toEqual(true);
+  });
+  it('checks if switch toggles on click', () => {
+    let wrapper = shallow(<Switch />);
+    expect(wrapper.find('input').exists()).toEqual(true);
+    wrapper.find('input').simulate('click');
+    expect(wrapper.find('input').exists()).toEqual(true);
```
julianguyen: I would test the value of checked i.e. `wrapper.state('checked')`

julianguyen:
I'm going to finish up this feature. I'm sorry we didn't get a chance to wrap this up for RGSoC.

aSquare14:
Not a problem . All the best !

PR #915, opened 2018-07-10
Removed non-RESTful routes on groups_controller.

jaspervanbrian:
Made a dedicated controller for those routes called MembershipsController
Modified it to follow RESTful verbs
Edited helpers and routes also.


app/controllers/groups/memberships_controller.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+module Groups
```
julianguyen: Nice!

app/helpers/groups_helper.rb
```diff
@@ -36,14 +36,35 @@ def delete_group_link(group, attrs = {})
             }.merge(attrs)
   end
 
+  def kick_member_link(group, member)
+    form_for(
+      :anything,
```
julianguyen: What does this refer to? Tried to look up documentation, but couldn't find anything.

spec/controllers/groups/memberships_controller_spec.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.describe Groups::MembershipsController, type: :controller do
+
```
julianguyen: Can you add test coverage for this?

app/helpers/groups_helper.rb
jaspervanbrian: Should I change this to `form_tag`? Basically this is for the kick member link as I transformed it from `link_to` tag to `form_for` tag to make the `method: :delete` work. (It's basically a hackish solution to a problem we can't understand). So the purpose of `:anything` is just to fill the placeholder model because `form_for` expects a model beforehand. If you try to remove it, it will not work.

spec/controllers/groups/memberships_controller_spec.rb
jaspervanbrian: Need a pair for helping me in writing tests because as of now, I don't know how to write tests yet. :sweat_smile:

app/assets/javascripts/notifications.js
```diff
@@ -141,9 +141,12 @@ var onReadyNotifications = function() {
   });
 
   $('.tip_notifications_text').click(function(event) {
-    event.stopPropagation();
+    if(! $(event.target).hasClass("kick")){
```
julianguyen: Just for stylistic purposes let's do `if (!$(event.target).hasClass('kick')){`
Also just for future reference, let's add a brief comment above describing why this was done

julianguyen: Can you update this as I requested? ^

padi:
Nice work @jaspervanbrian!

PR #917, opened 2018-07-22
Client code cleanup: re-organize file structure, set up prettier

julianguyen:
Starting work for cleaning up code in the client
* Re-structured file structure
* Use named exports over default exports
* Set up prettier and run during `yarn lint`
* Delete separate client README, already captured in Wiki
* Disable defaultTypes for React components
* Run React tests in Circle build

There will be future PRs with more cleanup :)


client/package.json
```diff
@@ -91,7 +93,8 @@
     "karma-spec-reporter": "^0.0.32",
     "karma-webpack": "^2.0.4",
     "node-sass": "^4.5.3",
-    "optimize-css-assets-webpack-plugin": "^3.1.1",
+    "optimize-css-assets-webpack-plugin": "^5.0.0",
```
julianguyen: Doesn't technically need to be updated, I was trying to re-enable it but it still wasn't working. Will try again in a follow up PR.

Issue #811, opened 2018-02-14
Refactor applications controller


camillevilla:
Hi! Our `application_controller_spec.rb` has soo many tests . Yay! But it could take advantage of some shared examples, let statements, etc. This could make the code less repetitive) and easier to read if you're not familiar with the file.

The wonderful @whyderrick has offered to help out with this effort. If anyone is interested in pairing, doing some analysis and splitting up work, comment here or on the #dev channel on Slack.

aSquare14:
I want to work on this !

whyderrick:
@aSquare14 hey Atibhi! I'll be working on this tomorrow, but I'm in SF so it'll be sometime between 00:00 and 7:00 Bangalore time. It might be tough to pair, but I'd love for you to do the code review or to pick up on it if I don't finish tomorrow.

aSquare14:
@whyderrick That would be cool! 😊

aSquare14:
@whyderrick  What is the progress?

vildevev:
Hi, I'm willing to work on this issue @camillevilla

camillevilla:
@vildevev did a great job of starting this refactor. We now have some shared examples (`let` and `before`) blocks that folks can try using. 

Other ideas for refactoring:
- [ ] move additional variables into shared examples with `let` 
  - [ ] `mood`
  - [ ] `category`
  - [ ] `moment`
  - [ ] `strategy`
  - [ ] `new_allies`
- [ ] extract helpers to another file (e.g. make creating + sign_in user a single method)
- [ ] find a DRYer way to create categories that have similar values. Is there a shorter way to create `new_category1`, `new_category2` etc. when they all have the same `userid`?

**Notes**: Folks should feel free to claim one of these items and comment on this issue. I highly recommend picking one bullet point and submitting a small PR first. It's better to move a little work forward than to take on the whole thing.

StaphSynth:
Just letting people know, I'm currently working on refactoring these tests as well as the controller itself.

PR #919, opened 2018-07-23
811 refactor application controller spec

StaphSynth:
# Description 

Refactoring `application_controller_spec.rb`.

## More Details

There's plenty to do in this controller and its specs. I thought since this is my first contribution to this project, I'd stop now and put up a smaller PR that's easier to review and take this a bit at a time.

I've refactored the spec for `#most_focus`, but so far only the `category` section. I've added several `context` blocks to improve the readability of the spec, but I was figuring a lot of this out from code context, so please correct me if I'm mistaken about the purpose of some of these tests.

While doing this, I also did a tiny bit of refactoring of a couple of the controller methods. This was mostly done while figuring out what they were doing in an attempt to make them easier to read. Again, please correct me if I've over-stepped the mark or mis-read the intentions here.

I look forward to your feedback :)

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/811

# Test Coverage
✅

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -418,7 +418,7 @@ class ApplicationController < ActionController::Base
       query = Strategy.published
     end
 
-    query.where(userid: user.id).all.recent.map do |story|
+    query.where(user_id: user.id).all.recent.map do |story|
       story if story.viewers.include?(current_user.id)
     end.compact
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -110,6 +110,7 @@ class ApplicationController < ActionController::Base
   end
 
   # rubocop:disable MethodLength
+  # TODO: move this method out of the controller + refactor
   def most_focus(data_type, profile_id)
     data = []
     user_id = profile_id || current_user.id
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -288,21 +289,18 @@ def data_included?(data_type, data_id, data)
...
-      freq.delete(max)
-      data.delete(max)
+    # Turn data array into hash of value => freq_of_value
+    freq = data.each_with_object(Hash.new(0)) do |value, hash|
+      hash[value] += 1
     end
-    result
+
+    # Sort hash and take highest 3 values
+    freq.sort_by do |occurances, _value|
```
julianguyen: Change occurances => occurrences

spec/controllers/application_controller_spec.rb
```diff
@@ -12,89 +12,103 @@
...
-          new_category4 = create(:category, userid: user2.id)
-          new_moment = create(:moment, userid: user2.id, category: Array.new(1, new_category2.id), viewers: Array.new(1, user1.id), published_at: Time.zone.now)
-          new_strategy = create(:strategy, userid: user2.id, category: [new_category1.id, new_category2.id, new_category3.id, new_category4.id], published_at: Time.zone.now)
-          result = controller.most_focus('category', user2.id)
-          expect(result.length).to eq(1)
-          expect(result[new_category1.id]).to eq(nil)
-          expect(result[new_category2.id]).to eq(1)
-          expect(result[new_category3.id]).to eq(nil)
-          expect(result[new_category4.id]).to eq(nil)
+          it 'includes duplicate categories once' do
```
julianguyen: This test is unclear to me, could you clarify this in the test? It's not clear how the two expects imply that the duplicate categories appear once.

app/controllers/application_controller.rb
StaphSynth: Thanks. I'm a terrible speller :)

spec/controllers/application_controller_spec.rb
StaphSynth: The test is meant to demonstrate that if I create a `Moment` and a `Strategy` and give each one the same category, that the category will only appear once in the hash returned by `#most_focus`. I _think_ that was part of the intent in the original test.

However, now that you mention it, I think the first `expect` statement is quite unclear. I should assert that `subject.length == 1`. Following on from that, the actual value of `subject[category.id]` should be 2 since the one category has appeared twice.

julianguyen: Thanks for clarifying and let's do what you suggested!

app/controllers/application_controller.rb
julianguyen: No worries! I use a spell checker in my IDE, that might be helpful

julianguyen:
Looks like the Circle build is failing cause we have to rename all instance of "userid" to "user_id"! We recently ran a migration to change all of the foreign keys to this.

StaphSynth:
Yeah I was in the middle of fixing that stuff yesterday and a work party got in the way. Will finish it up later today when my hangover subsides :)

PR #923, opened 2018-07-29
Refactor: remove duplicate code

julianguyen:
# Description 

Went through [refactoring suggestions](https://codeclimate.com/github/ifmeorg/ifme/issues?category=duplication&engine_name%5B%5D=structure&engine_name%5B%5D=duplication) for duplicated code in CodeClimate and updated codebase

## More Details

* Refactor: remove duplicate code - meetings_controller.rb
* Refactor: remove duplicate code - notification_mailer_helper.rb
* Refactor: remove duplicate code - create methods in shared.rb
* Refactor: remove duplicate code - update methods in shared.rb
* Refactor: remove duplicate code - delete_notifications in shared.rb
* Replace usage of raw with sanitize
* Fix broken grid on groups page

# Test Coverage

✅

### DIFF
--- a/app/controllers/concerns/shared.rb
+++ b/app/controllers/concerns/shared.rb
@@ -10,13 +10,11 @@ module Shared
   def shared_create(model_object, model_name)
     respond_to do |format|
       if model_object.save
-        format.html { redirect_to redirect_path(model_object, model_name) }
+        format_html_success(format, model_object, model_name)
         format.json { render :show, status: :created, location: model_object }
       else
         format.html { render :new }
-        format.json do
-          render json: model_object.errors, status: :unprocessable_entity
-        end
+        format_json_failure(format, model_object)
       end
     end
   end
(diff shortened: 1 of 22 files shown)


app/controllers/concerns/shared.rb
```diff
@@ -0,0 +1,65 @@
...
+    public_uniqueid = "comment_on_#{model_name}_#{commentid}"
+    Notification.where(uniqueid: public_uniqueid).destroy_all
+    private_uniqueid = "comment_on_#{model_name}_private_#{commentid}"
+    Notification.where(uniqueid: private_uniqueid).destroy_all
+  end
+
+  private
+
+  def redirect_path(model_object, model_name)
+    if model_name == 'mood'
```
StaphSynth: Better to use a case statement here instead?

app/controllers/meetings_controller.rb
```diff
@@ -450,5 +359,49 @@ def hide_page(meeting)
     )
     !(meeting_obj.exists? && meeting_member.exists?)
   end
+
+  # rubocop:disable MethodLength
+  def notifications_for_members(members, type)
```
StaphSynth: I think it would be a good idea to move this out of the controller and into a separate `Notifications` service.

app/controllers/concerns/shared.rb
StaphSynth: Perhaps we can use a `case` statement here instead of a series of `if`s?

app/controllers/meetings_controller.rb
julianguyen: That's a better idea, I'll look into that. Thanks!

StaphSynth: I should as well. I've seen it done, but never written it myself.

PR #925, opened 2018-07-29
Separate notification.rb concern into services

julianguyen:
# Description 

After having a discussion in #refactoring on concerns vs. services, I realized that the notifications.rb concern I created in the [last merged commit](https://github.com/ifmeorg/ifme/commit/f9b9e3e3e897866ae1513edc02e78b51a1132e72) was more of a service because there was business logic.

Also created shared_destroy for strategies, moods, and categories

## More Details

Create two services:
* comment_notifications_service.rb
* meeting_notifications_service.rb

# Test Coverage

✅


app/services/comment_notifications.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class CommentNotifications
+  attr_reader :comment_id, :model_name
+
+  def initialize(args)
```
StaphSynth: If the `comment_id` and `model_name` args are always required, I think we should make them named params:
`def initialize(comment_id:, model_name:)`

app/services/comment_notifications.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class CommentNotifications
```
StaphSynth: I think this should be renamed `CommentNofificationService` to reflect its purpose.

app/controllers/moments_controller.rb
```diff
@@ -63,7 +62,8 @@ def delete_comment
     end
 
     if comment_exists && ((is_my_comment && is_a_viewer) || is_my_moment)
-      delete_comment_and_notifications(params[:commentid], 'moment')
+      CommentNotifications.remove(comment_id: params[:commentid],
```
StaphSynth: Heh, I didn't know you could do that in Ruby. TIL :)

app/services/meeting_notifications.rb
```diff
@@ -1,27 +1,29 @@
...
+  attr_reader :current_user
+  attr_reader :meeting
+  attr_reader :type
+  attr_reader :members
+  attr_reader :uniqueid
 
-  included do
-    helper_method :notifications_for_meeting_members,
-                  :delete_comment_and_notifications
+  def initialize(args)
```
StaphSynth: Again, I think named args would be good here.

app/services/meeting_notifications.rb
```diff
@@ -1,27 +1,29 @@
 # frozen_string_literal: true
 
-module Notifications
-  extend ActiveSupport::Concern
+class MeetingNotifications
```
StaphSynth: Again, I think we should call this `MeetingNotificationService`.

app/services/meeting_notifications.rb
```diff
@@ -1,27 +1,29 @@
...
+  def self.handle_members(args = {})
+    new(args).handle_members
   end
 
   # rubocop:disable MethodLength
-  def notifications_for_meeting_members(meeting, members, type)
-    uniqueid = "#{type}_#{current_user.id}"
+  def handle_members
     members.each do |member|
       next if member.user_id == current_user.id
```
StaphSynth: If the members are being passed by the caller to the service, you could _probably_ assume they won't contain the current user. In the event that they do, perhaps we should reject the current user in `initialize` instead of performing a `next` here. So something like:

```
def initialize(args)
  @members = args[:members].reject { |member| member.id == current_user.id }
  ...
end
```

spec/services/comment_notifications_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    context 'when model_name is moment' do
+      let!(:model_name) { 'moment' }
+      let!(:moment) { create(:moment, user_id: user.id, viewers: [ally.id]) }
+      let!(:comment) { Comment.create_from!(comment: 'Hello', commentable_type: model_name, commentable_id: moment.id, comment_by: ally.id, visibility: visibility) }
+
+      context 'when comment visibility is all' do
+        let!(:visibility) { 'all' }
+
+        it 'works correctly' do
```
StaphSynth: I'm a big fan of describing what the thing should be doing in the test because it makes it so much easier to figure out the behaviour of the code later (and for new contributors as well). I'd prefer to see something like `it 'notifies the {who or whatever its supposed to be notifying}' do`

julianguyen:
Thanks for taking a look @StaphSynth, it's much appreciated :) 
Hope you feel better and you should definitely focus on getting rest.

app/services/comment_notifications.rb
jaspervanbrian: Yeah I think it should be renamed. It is useful for code readers, especially when they see the code at first glance and they will be like "oh okay this is injected in a service".

app/controllers/moments_controller.rb
jaspervanbrian: Is this some kind of dependency injection?

app/services/comment_notifications.rb
```diff
@@ -0,0 +1,22 @@
...
+    new(args).remove
+  end
+
+  def remove
+    Comment.find(comment_id).destroy
+    public_uniqueid = "comment_on_#{model_name}_#{comment_id}"
+    Notification.where(uniqueid: public_uniqueid).destroy_all
+    private_uniqueid = "comment_on_#{model_name}_private_#{comment_id}"
+    Notification.where(uniqueid: private_uniqueid).destroy_all
+  end
```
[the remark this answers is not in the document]
julianguyen: No worries good question it's cause of the following lines:
```
attr_reader :comment_id, :model_name

  def initialize(args)
    @comment_id = args[:comment_id]
    @model_name = args[:model_name]
  end
```

Accessor methods give you access to instance variables. I'm using the reader method cause I don't need to write values.

julianguyen:
Thanks for the feedback @StaphSynth and @jaspervanbrian. Could y'all re-review when you have the time to? :)

app/services/comment_notifications.rb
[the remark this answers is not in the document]
jaspervanbrian: Oh okay looks good to me.

PR #962, opened 2018-09-16
Update 20170724205314_add_secret_shares_to_moments.rb

aSquare14:
# Description 

The Active Migration version should be specified. Otherwise, `rake:db migrate` fails.
I suspect this to be due to updating the new rails version.


julianguyen:
Oooh great catch! :D

Issue #976, opened 2018-09-18
Add headers for security


sexybiggetje:
# Description 

The domain if-me.org is missing some headers which could improve it's security.

PASS(if-me.org): HTTP redirects to HTTPS using a 301
FAIL(if-me.org): Strict-Transport-Security missing
FAIL(if-me.org): X-Frame-Options missing
FAIL(if-me.org): X-Content-Type-Options missing
FAIL(if-me.org): X-XSS-Protection missing
FAIL(if-me.org): Referrer-Policy missing
FAIL(if-me.org): Content-Security-Policy missing
FAIL(if-me.org): Feature-Policy missing
PASS(if-me.org): Site sends 'Server', but probably only a vendor ID: Cowboy
PASS(if-me.org): Site doesn't send 'X-Powered-By'
FAIL(if-me.org): Site sends 'Via', probably disclosing version info: 1.1 vegur
PASS(if-me.org): Site doesn't send 'X-AspNet-Version'
PASS(if-me.org): Site doesn't send 'X-AspNetMvc-Version'
PASS(if-me.org): No SCM repository at: http://if-me.org/.git/HEAD
PASS(if-me.org): No SCM repository at: http://if-me.org/.hg/store/00manifest.i
PASS(if-me.org): No SCM repository at: http://if-me.org/.svn/entries

# Reproduction Steps

Grab a header analyzer like twa (https://github.com/woodruffw/twa). Analyse the headers and see that it outputs something similar to above.

# Do you want to be the assignee to work on this?

🚫

Not much at a pc these days.

julianguyen:
Great call out! Thank you :)

PR #996, opened 2018-09-23
#976 Missing HTTP security headers added

mialahmoka:
Added HTTP security headers to application.rb

# Description 
Added the following headers to application.rb:
  'X-Frame-Options' => 'SAMEORIGIN',
  'X-XSS-Protection' => '1; mode=block',
  'X-Content-Type-Options' => 'nosniff',
  'X-Download-Options' => 'noopen',
  'X-Permitted-Cross-Domain-Policies' => 'none',
  'Referrer-Policy' => 'strict-origin-when-cross-origin',
  'Strict-Transport-Security' => 'max-age=31536000'
All of these headers are necessary to prevent cross site scripting, clickjacking, and a non default option (STS) that prevents man-in-the-middle attacks by stripping TLS. 
## More Details
While security headers are necessary, it does not make the website invulnerable. It's always a good idea to test for cross site scripting throughout the entire domain and patch it as soon as possible.

In addition, the missing Content-Security-Policy header requires the additional file content_security_policy.rb which is missing from config/initializers/.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/976

# Screenshots

image

# Test Coverage
In progress

### DIFF
--- a/application.rb
+++ b/application.rb
@@ -46,9 +46,9 @@ module Ifme
      'Referrer-Policy' => 'strict-origin-when-cross-origin',
      'Strict-Transport-Security' => 'max-age=31536000'
     }
-    
+
     # gzip the html/json responses
-    config.middleware.use Rack::Deflater, 
+    config.middleware.use Rack::Deflater,
       include: %w[text/html application/json image/svg+xml]
     # export translations for use in javascript
     config.middleware.use I18n::JS::Middleware
--- a/application.rb
+++ b/application.rb
@@ -37,20 +37,19 @@ module Ifme
       end
     }
 
-    
     config.action_dispatch.default_headers = {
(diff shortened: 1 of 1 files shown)


application.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
```
julianguyen: Out of curiosity why did you create a new application.rb file and not use the one in the config folder?

mialahmoka: Didn't actually mean to do that, git wouldn't commit any of my changes in bash so I tried the upload tool on the site. Sorry it's my first time using git

mialahmoka: I can't overwrite application.rb so I'll just close this and start a new pull request

PR #1000, opened 2018-09-23
#976 HTTP security headers patch

mialahmoka:
added security headers, split line 51 to make rubocop happy

# Description 
Added the following headers to application.rb:
  'X-Frame-Options' => 'SAMEORIGIN',
  'X-XSS-Protection' => '1; mode=block',
  'X-Content-Type-Options' => 'nosniff',
  'X-Download-Options' => 'noopen',
  'X-Permitted-Cross-Domain-Policies' => 'none',
  'Referrer-Policy' => 'strict-origin-when-cross-origin',
  'Strict-Transport-Security' => 'max-age=31536000'
All of these headers are necessary to prevent cross site scripting, clickjacking, and a non default option (STS) that prevents man-in-the-middle attacks by stripping TLS. 
## More Details
While security headers are necessary, it does not make the website invulnerable. It's always a good idea to test for cross site scripting throughout the entire domain and patch it as soon as possible.

In addition, the missing Content-Security-Policy header requires the additional file content_security_policy.rb which is missing from config/initializers/.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/976

# Screenshots

image

# Test Coverage

✅


mialahmoka:
The way I usually check HTTP headers is through burp proxy, but the tool linked in the original issue should work as well

julianguyen:
Thanks for sharing! We should write some documentation on how to check that in our Wiki using the tools y'all suggested.

sexybiggetje:
> In addition, the missing Content-Security-Policy header requires the additional file content_security_policy.rb which is missing from config/initializers/.

Is this something that needs to be done? I don't see such a file being added in this PR.
I'll checkout and review this PR tonight. Thanks so far :)

mialahmoka:
I'm new to ruby and don't know all that's involved in making that file from scratch, however I think it's best if we merge this now and create a new PR once that file is created for security's sake. Otherwise as it stands HTTP header injection is very easy.

PR #1001, opened 2018-09-24
Design touchups

nshki:
# Description 

This PR consists of various small tweaks to the site redesign, from font sizes to background colors to scroll behavior. I also added a blurb since I haven't done so yet over the past year! 😬


app/assets/stylesheets/core/footer.scss
```diff
@@ -55,7 +59,9 @@
   }
 
   &Extras {
-    padding: $size-0 114px 66px 114px;
+    max-width: 1200px;
+    margin: 0 auto;
+    padding: $size-0 34px 66px;
```
julianguyen: Out of curiosity, would we want this section to have the same padding for smaller screen breakpoints?

app/assets/stylesheets/static/static_content.scss
```diff
@@ -84,7 +118,42 @@
...
+
+    [type='text'],
+    [type='email'],
+    [type='password'],
+    [type='submit'] {
+      width: 100%;
+    }
+
+    [type='checkbox'] {
+      float: left;
```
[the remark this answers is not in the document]
julianguyen: Ohhh I see, that works then. That way you don't have to go around add divs with flexboxes around things. Thanks for explaining!

app/assets/stylesheets/core/footer.scss
nshki: Mm actually good point. Going to reduce mobile horizontal padding here to stay consistent with the static content container.

Issue #997, opened 2018-09-23
Move out comment related methods in application_controller.rb to a helper


julianguyen:
# Description 

application_controller.rb has a lot of miscellaneous methods that are used throughout the app including methods for comments.

Let's move this out to a comment helper!

# Do you want to be the assignee to work on this?

🚫

PR #1007, opened 2018-09-25
Add comments helper

nshki:
# Description 

Moves all comment-related methods in `application_controller.rb` into its own helper, and makes sure that all controllers that use those methods include the helper.

## Corresponding Issue

#997 

# Test Coverage

✅


spec/helpers/comments_helper_spec.rb
```diff
@@ -0,0 +1,194 @@
+# frozen_string_literal: true
+
+describe CommentsHelper, type: :controller do
+  let(:user1) { create(:user1) }
+  let(:user2) { create(:user2) }
+
+  controller(ApplicationController) do
```
julianguyen: Even though we don't run Rubocop on test files, I think you can do `controller(ApplicationController) {}` if you want to keep it as a one-liner

nshki: Gotcha! Not that I think we will put anything in that block, but I'm going to leave it two lines to keep consistent with how RSpec documents it.

PR #1008, opened 2018-09-25
Deal with render blocking JS scripts

julianguyen:
# Description 

- Removed unused scripts
- Add async to third-party scripts
- Add defer to webpack_bundle.js because async doesn't work and messes up async (according to React on Rails docs)
- Set up split chunks in Webpack
- FIx gridItemBoxGhost for contributors grid since changing gridItemBoxDark broke the designs in post actions

## More Details

Another PR will investigate inline CSS, will take a look at https://github.com/GoogleChromeLabs/critters/tree/master/src

# Test Coverage

🚫  N/A


app/views/layouts/application.html.erb
```diff
@@ -20,32 +20,26 @@
...
-    <style type="text/css">
-      .ui-autocomplete-loading { background: white url("/assets/ui-anim_basic_16x16.gif") right center no-repeat; }
-    </style>
-
-    <!-- Google Maps API -->
-    <script src="https://maps.googleapis.com/maps/api/js?key=<%= ENV['GOOGLE_API_KEY'] %>&libraries=places&callback=initAutocomplete"
-        async defer></script>
+    <% if join_path? || update_account_path? %>
+      <!-- Google Maps API -->
+      <script src="https://maps.googleapis.com/maps/api/js?key=<%= ENV['GOOGLE_API_KEY'] %>&libraries=places&callback=initAutocomplete" async></script>
```
baohouse: I forget, are the address fields rendered via ERB or React? If React, then you might have a race condition here since the webpack_bundle contains the React components, and the Google Maps lib needs to load before webpack_bundle.

julianguyen: It's by ERB so I think there shouldn't be a race condition, but yeah with React there would be. Good call out!

app/views/layouts/application.html.erb
```diff
@@ -20,32 +20,26 @@
...
+    <%= javascript_pack_tag 'webpack_bundle', defer: Rails.env.production? %>
     <%= stylesheet_pack_tag 'webpack_bundle' %>
     <!-- Webpack Assets END -->
 
     <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
-    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload', async: Rails.env.production? %>
+
+    <!-- Jstimezonedetect -->
+    <script src='https://cdnjs.cloudflare.com/ajax/libs/jstimezonedetect/1.0.4/jstz.min.js' async></script>
```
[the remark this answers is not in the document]
julianguyen: Hmmm yeah, that sounds right to me. I should move this to React including what's happening in the onReadyApplication() function. Thanks!

app/views/layouts/application.html.erb
```diff
@@ -20,32 +20,23 @@
...
     <!-- Webpack Assets END -->
 
     <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
-    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload', async: Rails.env.production? %>
 
     <% if user_signed_in? %>
       <!-- Pusher -->
-      <script src="https://js.pusher.com/3.0/pusher.min.js"></script>
+      <script src="https://js.pusher.com/3.0/pusher.min.js" async></script>
```
nshki: Any reason why `<script>` is being used here instead of `javascript_include_tag`?

app/views/layouts/application.html.erb
```diff
@@ -20,32 +20,23 @@
...
-    <style type="text/css">
-      .ui-autocomplete-loading { background: white url("/assets/ui-anim_basic_16x16.gif") right center no-repeat; }
-    </style>
-
-    <!-- Google Maps API -->
-    <script src="https://maps.googleapis.com/maps/api/js?key=<%= ENV['GOOGLE_API_KEY'] %>&libraries=places&callback=initAutocomplete"
-        async defer></script>
+    <% if join_path? || update_account_path? %>
+      <!-- Google Maps API -->
+      <script src="https://maps.googleapis.com/maps/api/js?key=<%= ENV['GOOGLE_API_KEY'] %>&libraries=places&callback=initAutocomplete" async></script>
```
nshki: Same thing here RE: using `javascript_include_tag`.

app/views/layouts/application.html.erb
julianguyen: No reason, I just copied at pasted the script link from these third-party APIs in their README. I can convert this over to javascript_include_tag to be consistent with Rails ERB syntax :)

app/views/layouts/application.html.erb
julianguyen: Thanks for explaining this! I'll keep this mind

PR #1010, opened 2018-09-26
Various optimizations: inline JS and CSS, a11y improvements

julianguyen:
# Description 

- Add data-turbolinks-track to deferred web_bundle.js & inline JS and CSS for Rails assets
- All text remains visible while webfont loads
- A11y improvements
- Remove unused CSS
- Renamed snakecase classes/ids to camelcase with the exception of styles related to comments (this is going to be rewritten in React anyways and the code is gross)

## More Details

I couldn't get inline_js working for the Webpack JS asset :( Will investigate another time
Future TODO: find a way to combine/minify application.css and webpack_bundles.css

# Test Coverage

✅


julianguyen:
Hmmm I might have to move the SCSS in /client back to the Rails side 🤔

client/app/components/Tag/Tag.scss
```diff
@@ -31,8 +31,8 @@
   }
 
   &.secondary {
-    color: $white;
```
julianguyen: Had to change this for colour contrast improvements

julianguyen:
Hmmm so I think I came up with a possibly hacky way of getting inline JS and CSS for the webpack bundles.

In application.html.erb do
```
if Rails.env.production?
  <%= javascript_pack_tag 'webpack_bundle', 'data-turbolinks-track': true, defer: true %>
  <%= stylesheet_pack_tag 'webpack_bundle', media: 'all' %>
else 
   <%= inline_css 'webpack_bundle.css' %>
    <%= inline_js 'webpack_bundle.js' %>
end
```

Then in webpack.config.js exclude the hash name from outputFilename so that it's just `[name]` 🤔

app/controllers/application_controller.rb
```diff
@@ -210,7 +210,7 @@ def moments_stats
 
     return '' if total_count <= 1
 
-    result = '<div class="center" id="stats">'
+    result = '<div class="center" class="stats">'
```
baohouse: class attribute twice?

spec/controllers/application_controller_spec.rb
```diff
@@ -148,18 +148,18 @@
     it 'has more than one moment created this month' do
       new_moment1 = create(:moment, user_id: user1.id)
       new_moment2 = create(:moment, user_id: user1.id)
-      expect(controller.moments_stats).to eq('<div class="center" id="stats">You have written a <strong>total</strong> of <strong>2</strong> moments.</div>')
+      expect(controller.moments_stats).to eq('<div class="center" class="stats">You have written a <strong>total</strong> of <strong>2</strong> moments.</div>')
```
baohouse: Class attributes should be merged (there's two of them in the same element)

PR #1017, opened 2018-09-26
Move moments stats method into helper

DamonClark:
Fixes [#1014](https://github.com/ifmeorg/ifme/issues/1014)

Moved the moments_stats method from the application_controller into its respective helper. 

It looks like it also pushed some changes from my local package.json and yarn.lock files. So I am not sure if this can be merged. Please let me know and I can try making the proper adjustments.


app/helpers/moments_helper.rb
```diff
@@ -99,6 +99,35 @@ def moment_form_inputs
   end
   # rubocop:enable MethodLength
 
+  # rubocop:disable MethodLength
+  def moments_stats
+    total_count = current_user.moments.all.count
```
julianguyen: This is a good opportunity to reduce the size of this method and put some of the logic into private methods in this helper file.

julianguyen: We also want to move the tests related to this code (in application_controller_spec.rb) into moments_helper_spec.rb (which unfortunately doesn't exist yet)

julianguyen:
Closing this since another PR was opened

PR #1018, opened 2018-09-30
Bump Rails to 5.2.1

greysteil:
# Description 

Upgrade Rails to 5.2.1, the latest version 🎉 

## More Details

5.2.1 includes deprecation warnings for methods that will be removed in 6.0.0, so it's a little less straightforward than the 5.0 to 5.1 upgrade was. That said it shouldn't be that bad.

~~A couple of feature specs are failing for me locally, but I think that may just be because my setup isn't right. I've opened this so they'll run in CI and I can double-check.~~ Looks like it was setup related - tests are green 💚


app/helpers/application_helper.rb
```diff
@@ -120,7 +120,10 @@ def get_icon_text(icon, text)
   def image_link_to(image_path, url, image_tag_options = {},
                     link_to_options = {})
     link_to url, link_to_options do
-      image_tag image_path, image_tag_options
+      default_alt = File.basename(image_path, ".*").
```
greysteil: Rails appears to have accidentally removed the default alt in v5.2.1 (I think they intended only to deprecate it). The logic here is a straight copy from what Rails used to do, and future-proofs us.

spec/controllers/pages_controller_spec.rb
```diff
@@ -4,7 +4,7 @@
   describe 'GET #home' do
     it 'respond to request' do
       get :home
-      expect(response).to be_success
+      expect(response).to be_successful
```
greysteil: The `success?` predicate is deprecated in favour of `successful?`

app/helpers/application_helper.rb
```diff
@@ -120,7 +120,10 @@ def get_icon_text(icon, text)
   def image_link_to(image_path, url, image_tag_options = {},
                     link_to_options = {})
     link_to url, link_to_options do
-      image_tag image_path, image_tag_options
+      default_alt = File.basename(image_path, '.*')
```
julianguyen: Could you explain what's happening in default_alt and why it's necessary?

greysteil: Ah, sorry, I put a comment on that but it seems like a subsequent push resolved it 🤦‍♂️ 

Rails appears to have accidentally removed the default alt in v5.2.1 (I think they intended only to deprecate it). The logic here is a straight copy from what Rails used to do, and future-proofs us.

julianguyen: Ohhh wow, that's silly of them to do that. Thanks for heads up and explaining!

julianguyen:
FYI the failing test you encountered in CircleCI is unrelated to what you worked on and is being fixed :) https://github.com/ifmeorg/ifme/pull/1021

julianguyen:
Another thing I just realized you need to update is the Rails version on this Wiki page: https://github.com/ifmeorg/ifme/wiki/Installation

greysteil:
Awesome! I saw that was failing in other test runs so decided I could ignore it. Was thinking I should put in a PR to make that status check fail if test are failing, though (separately).

I don't have the ability to merge so will have to leave it to you to shepherd this one out!

greysteil:
> Another thing I just realized you need to update is the Rails version on this Wiki page

Oh good point - done!

julianguyen:
@greysteil That would be great if you made a PR for that status fail check. Thanks so much, I'll merge this one now.

PR #1022, opened 2018-10-01
Add set -e to Circle config to fail build on failing tests

julianguyen:
# Description 

Failing test suites do not fail the CircleCI build, this started happening after I re-wrote the config to support sending code coverage to Code Climate. 

The cause of the issue is because I'm using multi-line commands to execute tests now. To fix this issue I just include `set -e` before the commands (see: https://support.circleci.com/hc/en-us/articles/115015733328-Tests-fail-but-job-finishes-successfully)

# Test Coverage

🚫  N/A

I realized I never added you @greysteil as a member of the Github org so I can't add you as a reviewer. Just sent you an invite! Could you review this?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,7 @@ jobs:
       - run:
           name: Run React Jest Tests
           command: |
+            set -e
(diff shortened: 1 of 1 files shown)


julianguyen:
Thanks so much for the review @greysteil :)

PR #1038, opened 2018-10-01
move method reduce size and add test

DamonClark:
Fixes [#1014](https://github.com/ifmeorg/ifme/issues/1014)

Moved the moments_stats method from the application_controller into its respective helper. 
Reduced size of moments_stats by putting some logic into private methods 
Moved tests related to this code into new file moments_helper_spec.rb

### DIFF
--- a/app/helpers/moments_helper.rb
+++ b/app/helpers/moments_helper.rb
@@ -100,6 +100,26 @@ module MomentsHelper
   # rubocop:enable MethodLength
 
   # rubocop:disable MethodLength
+  def moments_stats
+    total_count = current_user.moments.all.count
+    monthly_count = current_user.moments.where(
+      created_at: Time.current.beginning_of_month..Time.current
+    ).count
+
+    return '' if total_count <= 1
+
+    result = '<div class="center stats">'
+    total_moment
+
+    if total_count != monthly_count
+      monthly_moment
+    end
+
+    result + '</div>'
+  end
+  # rubocop:enable MethodLength
+  #
+  # rubocop:disable MethodLength
   def moments_viewers_input
     input = {}
     if @viewers.present?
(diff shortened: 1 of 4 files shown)


app/helpers/moments_helper.rb
```diff
@@ -150,5 +170,24 @@ def data_for(item)
       @moment.strategy
     end
   end
+
+  private
```
julianguyen: As Codeclimate identified this is redundant, there's already an existing private modifier in the file, so you'll want to put these new private methods under that.

spec/helpers/moments_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+describe MomentsHelper do
+  describe 'moments_stats' do
```
julianguyen: It looks like the tests you moved into this file don't work, I highly recommend running this spec locally on your machine instead of relying on CircleCI to run it for you. From the root folder of the project simply run `rspec ./spec/helpers/moments_helper_spec.rb`

julianguyen: It looks like user1 cannot be found. That's because user1 is undefined. If you take a look at the original application_controller_spec.rb you moved this test from you'll see user1 comes from `let(:user1) { create(:user1) }`

app/helpers/moments_helper.rb
DamonClark: Ahhh! I didn't notice the private modifier on line 18 that already existed. I went ahead and removed the redundant private modifier and moved these new methods to be placed directly underneath the original one.

julianguyen:
Closing this as it's re-opened here: https://github.com/ifmeorg/ifme/pull/1064

PR #1040, opened 2018-10-01
Secure Notifications Controller

### LINKED ISSUE #1025: Secure notifications controller
# Description 

There was some really awesome progress being made in this [PR](https://github.com/ifmeorg/ifme/pull/316) to make the notifications controller more secure and add better test coverage. It never got finished, it would be great if someone could finish it up :)

🚫

tryantwit:
Additionally find_by will either return an ActiveRecord instance or
return nil so we can do away with the rescue and rubocop disable.

# Description 

Supercedes old pull request of #316 to secure notifications controller. Added additional Notification filter param of user_id in the before_action method `set_notification`. Using `Notification.find_by` returns nil if nothing is found so additionally allowed removal of the rescue in `set_notification` since it was doing the same return as in the `destroy` method anyway.

## Corresponding Issue
Fixes #1025.

# Test Coverage

🚫 
Did not expand on test coverage with this commit, only ensured that current test suite still runs and passes

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -62,6 +62,9 @@ class NotificationsController < ApplicationController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_notification
-    @notification = Notification.find_by(id: params[:id], user_id: current_user.id)
+    @notification = Notification.find_by(
+      id: params[:id],
+      user_id: current_user.id
+    )
   end
 end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -7,12 +7,7 @@ class NotificationsController < ApplicationController
   # DELETE /notifications/1
   # DELETE /notifications/1.json
   def destroy
-    notification = Notification.find_by(
-      id: params[:id],
-      user_id: current_user.id
-    )
-
-    notification.destroy if notification.present?
+    @notification.destroy if @notification.present?
 
     respond_to do |format|
       format.html { redirect_back(fallback_location: notifications_path) }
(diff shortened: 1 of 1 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -7,12 +7,7 @@ class NotificationsController < ApplicationController
   # DELETE /notifications/1
   # DELETE /notifications/1.json
   def destroy
-    notification = Notification.find_by(
-      id: params[:id],
-      user_id: current_user.id
-    )
-
-    notification.destroy if notification.present?
+    @notification.destroy if @notification.present?
```
julianguyen: Should we still ensure that the user who the notification belongs to can only destroy it?

tryantwit: I belive it should be doing that unless I'm missing something about the way things are structured. The before_action is setting the notification to be deleted based on the id being passed in and the id of the current user. A user shouldn't be able to access a notification that doesn't have their user_id and be unable to destroy it.

Though if I'm missing something let me know and I can make sure that's the end behavior.

tryantwit: Alright everything should be working like I thought. The Notification to be deleted is being selected on two criteria. The first is the `id` passed into the controller, the second is the id of the current_user object that is provided by Devise on login. In this way a Notification is only eligible for selection to be deleted if the id that is passed in exists and the `user_id` of that Notification matches the id of the currently signed in user.

There are other ways to write this same logic that make it more explicit that a check against the current_user is being done.

```ruby
notification = Notification.find_by(id: params[:id])
notification.destroy if notification.user_id == current_user.id 
```
or
```ruby
notification = current_user.notifications.destroy(Notification.find_by(id: params[:id])
```

Taking a look back through the original issue and pull request it looks like this route has actually been protected from that since Nov 16, 2016 and was modified with this commit 89b01d39616922ae561f2c7e6063f7c313afb662 so at this point this pull request is really only serving as a slight clean up of the controller.

julianguyen: Thanks again for looking into this and clarifying! In there case, `@notifications` works and it looks like we already have test coverage for the signed in and not signed in cases.

Issue #1013, opened 2018-09-26
Move get_stories method in application_controller.rb into its own helper. 


julianguyen:
# Description 

Move get_stories method in application_controller.rb into its own helper.  The motivation is to declutter application_controller and have better separation of concerns. Be sure to move the tests too.

Move user_stories in this new helper too and make it private

# Do you want to be the assignee to work on this?

🚫

jaimevelaz:
Hi @julianguyen, I would like to help with this one 😃

julianguyen:
Hey @jaimevelaz, thanks for expressing your interest. We would love your help with this issue. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :)

PR #1041, opened 2018-10-01
#1013 Extract stories helper

jaimevelaz:
# Description 

Move all `get_stories` related code to its own helper and take the opportunity to refactor it a bit.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1013

# Test Coverage

✅


app/controllers/application_controller.rb
```diff
@@ -15,6 +15,7 @@ class ApplicationController < ActionController::Base
   include ActionView::Helpers::DateHelper
   include ActionView::Helpers::TextHelper
   include CommentsHelper
+  include StoriesHelper
```
julianguyen: Is it necessary to include StoriesHelper here if it's being included in more specific controllers?

ahmetabdi:
Shouldn't something like this live in a `presenter` ?

app/controllers/application_controller.rb
jaimevelaz: Thanks for the feedback @julianguyen. You are right, in this case we do not need to include `StoriesHelper` in `ApplicationController` since the method is not used in views (unlike `CommentsHelper` method), I've updated the code to avoid including `StoriesHelper` in `ApplicationController` and use a `FakeStoriesController` in specs to avoid depending on a specific controller.

Feedback welcome

jaimevelaz:
> Shouldn't something like this live in a `presenter` ?

Hi @ahmetabdi, I had seen this kind of code in `presenters` but following project [backend best practices](https://github.com/ifmeorg/ifme/wiki/Backend-Practices#concern-vs-service-vs-helper) I went for a `helper`, I think consistency is more important in this case.

julianguyen:
Hey @ahmetabdi, thanks for bringing up presenters. I'm open to be honest, my understanding was that helpers could be used for presentational methods used in views. Can you tell why you would use a presenter over a helper?

jaimevelaz:
Thanks for the review @julianguyen, should I merge the branch myself?

julianguyen:
@jaimevelaz Yeah feel empowered to merge once everything gets the green light :)

Issue #1015, opened 2018-09-26
Move user_created_data? from application_controller.rb into comments_helper.rb


julianguyen:
# Description 

user_created_data? is just being used by comments_helper.rb let's move that there as a private method

# Do you want to be the assignee to work on this?

🚫

jiradeto:
I am interested

julianguyen:
Hey @jiradeto, thanks for expressing your interest. We would love your help with this issue. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :)

jeccb:
Hi, can I do it too?

julianguyen:
Hey @jeccb thanks for expressing your interest!  I'm not sure if @jiradeto is still interested in working on this. If not, it's up for grabs. Otherwise, there are other issues that are similar in terms of refactoring Rails code. I'm happy to point you to some if you can't find anything.

Please also check out [next steps](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack) for joining our contributor community :)

julianguyen:
Alternatively you two could do some remote pairing on this issue. Two heads are always better than one :)

asarbu:
hi 👋 I created a MR with the changes you requested

PR #1047, opened 2018-10-02
Move user_created_data? from application_controller.rb into comments_helper.rb

asarbu:
# Description 

Move user_created_data? from application_controller.rb into comments_helper.rb

## More Details

Move user_created_data? from application_controller.rb into comments_helper.rb because the action is used only in the helper class.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1015

# Test Coverage

✅ 
🚫


app/helpers/comments_helper.rb
```diff
@@ -84,4 +84,20 @@ def comment_for(model_name)
   rescue ActiveRecord::RecordInvalid
     respond_not_saved
   end
+
+  # rubocop:disable MethodLength
+  def user_created_data?(id, data_type)
```
julianguyen: So it looks like `user_created_data?` is only used in `comment_deletable?`. I would put a private above this method.

asarbu: sure 👍 . Do you think there might be other methods inside this helper that need to be private ?

julianguyen: I would double check to make sure, I don't know off the top of my head!

asarbu: I looked into the file and moved the methods that were privat under that scope. Let me know what you think

Issue #1050, opened 2018-10-02
Add missing test coverage for medication_refill_helper.rb


julianguyen:
# Description 

Womp womp we're missing test coverage here. Let's go back and add it!

# Do you want to be the assignee to work on this?

🚫

HolyWalley:
If you don't mind, I'm in :)

PR #1058, opened 2018-10-02
[i1050] Add missing test coverage for medication_refill_helper.rb

HolyWalley:
# Description 

Add tests for the medication_refill_helper.rb

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1050

# Test Coverage

✅

### DIFF
--- a/spec/helpers/calendar_helper_spec.rb
+++ b/spec/helpers/calendar_helper_spec.rb
@@ -9,7 +9,7 @@ describe CalendarHelper do
     it 'when add_to_google_cal is true without a refill date' do
       medication.add_to_google_cal = true
       medication.refill = nil
-      expect(new_cal_refill_reminder_needed?(medication)).to eq(false)
+      expect(new_cal_refill_reminder_needed?(medication)).to eq(nil)
     end
 
     it 'when add_to_google_cal is true with a refill date' do
--- a/spec/helpers/calendar_helper_spec.rb
+++ b/spec/helpers/calendar_helper_spec.rb
@@ -9,7 +9,7 @@ describe CalendarHelper do
     it 'when add_to_google_cal is true without a refill date' do
       medication.add_to_google_cal = true
       medication.refill = nil
-      expect(new_cal_refill_reminder_needed?(medication)).to eq(nil)
+      expect(new_cal_refill_reminder_needed?(medication)).to eq(false)
     end
 
     it 'when add_to_google_cal is true with a refill date' do
--- a/spec/helpers/medication_refill_helper_spec.rb
+++ b/spec/helpers/medication_refill_helper_spec.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe MedicationRefillHelper, type: :helper do
+  include CalendarHelper
+  include Devise::Test::ControllerHelpers
+
+  describe '#save_refill_to_google_calendar' do
+    let(:user) { FactoryBot.create(:user1) }
+    let(:medication) { FactoryBot.create(:medication, user_id: user.id) }
+    let(:exception_text) { 'RESCUE INVOKED' }
+
+    before do
+      allow_any_instance_of(helper.class).to receive(:return_to_sign_in).and_return(exception_text)
+      sign_in user
+    end
+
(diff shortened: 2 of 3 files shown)


spec/helpers/calendar_helper_spec.rb
```diff
@@ -9,7 +9,7 @@
     it 'when add_to_google_cal is true without a refill date' do
       medication.add_to_google_cal = true
       medication.refill = nil
-      expect(new_cal_refill_reminder_needed?(medication)).to eq(false)
+      expect(new_cal_refill_reminder_needed?(medication)).to eq(nil)
```
julianguyen: Why was this changed to nil?

spec/helpers/medication_refill_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:user) { FactoryBot.create(:user1) }
+    let(:medication) { FactoryBot.create(:medication, user_id: user.id) }
+    let(:exception_text) { 'RESCUE INVOKED' }
+
+    before do
+      allow_any_instance_of(helper.class).to receive(:return_to_sign_in).and_return(exception_text)
+      sign_in user
+    end
+
+    context 'when current_user.google_oauth2_enabled? && new_cal_refill_reminder_needed?(medication)' do
```
julianguyen: I would write this out as a sentence instead of referring to methods in the context

spec/helpers/medication_refill_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      before do
+        allow_any_instance_of(User).to receive(:google_oauth2_enabled?).and_return(true)
+        allow_any_instance_of(helper.class).to receive(:new_cal_refill_reminder_needed?).and_return(true)
+      end
+
+      it { expect(helper.save_refill_to_google_calendar(medication)).to eq(exception_text) }
+    end
+
+    context 'until current_user.google_oauth2_enabled? && new_cal_refill_reminder_needed?(medication)' do
+      it { expect(helper.save_refill_to_google_calendar(medication)).to eq(true) }
```
julianguyen: Similar comment as above

spec/helpers/calendar_helper_spec.rb
HolyWalley: so, in ruby nil is a negative value in bool expressions. for example

if nil
  true
else
  false
end

#=> false

but nil != false and this is why test is falling down with false instead of nil.

spec/helpers/medication_refill_helper_spec.rb
HolyWalley: I agree. I'll fix it in a few hours

spec/helpers/calendar_helper_spec.rb
HolyWalley: we can also change `#new_cal_refill_reminder_needed` to this:

```
!!(medication.add_to_google_cal && medication.refill &&
  (medication.add_to_google_cal_changed? || medication.refill_changed?))
```

it will have a similar result, but I'm not sure that this is good practice.

julianguyen: Hmm, ideally that method should return a boolean. Returning either true or nil isn't very consistent to me. 

Double nots are messy too, as you pointed out, what about `nil?`

spec/helpers/medication_refill_helper_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:user) { FactoryBot.create(:user1) }
+    let(:medication) { FactoryBot.create(:medication, user_id: user.id) }
+    let(:exception_text) { 'RESCUE INVOKED' }
+
+    before do
+      allow_any_instance_of(helper.class).to receive(:return_to_sign_in).and_return(exception_text)
+      sign_in user
+    end
+
+    context 'when the user has the google oauth2 enabled and he need a new refill reminder needed' do
```
julianguyen: Let's remove the "he" pronoun to be neutral. You can replace it with "they" instead.

spec/helpers/medication_refill_helper_spec.rb
julianguyen: Similar pronoun comment above

HolyWalley:
@julianguyen done :)

Issue #1049, opened 2018-10-02
Adding missing test coverage for invitations_helper.rb


julianguyen:
# Description 

Womp womp we're missing test coverage here. Let's go back and add it!

# Do you want to be the assignee to work on this?

🚫

gdestree:
hi @julianguyen I would be interested in working on this one.

imi56:
I would like to work on this.

PR #1059, opened 2018-10-02
[i1049] small refactoring

HolyWalley:
# Description 

Your helper looks like a simple database association, so, I replaced it.

## More Details

As I can see that only user can have invited_by_id, and I think it will not be changed in the future. Another way is replace User.where(id: resource.invited_by_id).first.try(:name) by User.find_by(id: resource.invited_by_id)&.name but I think my way will be the best.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1049

# Test Coverage

🚫 it's just a database association :)

### DIFF
--- a/app/helpers/invitations_helper.rb
+++ b/app/helpers/invitations_helper.rb
@@ -1,7 +1,0 @@
-# frozen_string_literal: true
-
-module InvitationsHelper
-  def inviter_name(resource)
-    User.where(id: resource.invited_by_id).first.try(:name)
-  end
-end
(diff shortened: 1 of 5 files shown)


HolyWalley:
I'll take a look at this in a few hours.

HolyWalley:
@julianguyen I'm not sure why, but 
`bundle exec rspec ./spec/controllers/strategies_controller_spec.rb:295`
gives the same error on the master branch. can you check it locally? 

if I replace
`another_user = create(:user)`
but
`another_user = create(:user1)`
it works. 

But it feels strange that this issue has appeared now.

HolyWalley:
super strange, but after merge the build turned green

Issue #1012, opened 2018-09-26
Move tag_usage method in application_controller.rb into its own helper. 


julianguyen:
# Description 

Move tag_usage method in application_controller.rb into its own helper.  The motivation is to declutter application_controller and have better separation of concerns. Be sure to move the tests too.

Let's move data_included? as well as a private method in the new helper.

# Do you want to be the assignee to work on this?

🚫

Atul9:
@julianguyen I would like to work on this

julianguyen:
Hey @Atul9, thanks for expressing your interest. We would love your help with this issue. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :)

PR #1065, opened 2018-10-02
#1012 Move tag_usage method in application_controller.rb into its own…

jiradeto:
# Description 

Move tag_usage method in application_controller.rb into its own helper.

## More Details

- create new helper function called tags_hepler.rb
- move function tag_usage into the new helper
- move private function data_included into the new helper
- move the test 

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1012

# Test Coverage

✅


app/helpers/tags_helper.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+module TagsHelper
+  # rubocop:disable MethodLength
+  def tag_usage(data_id, data_type, user_id)
+    result = []
+    moments = Moment.where(user_id: user_id).order('created_at DESC')
+    if data_type == 'category'
+      strategies = Strategy.where(user_id: user_id).order('created_at DESC')
+      [moments, strategies].each do |records|
```
jm96441n: can both branches of this conditional be extracted to private methods that return the proper array and assign that to the `result` variable?
so something like:
```
if data_type == 'category'
   result  = get_moments
elsif data_type.in?(%w[mood strategy])
   result = get_moods
end
```
or something along those lines, maybe not those method names specifically but that's the idea

julianguyen:
@jiradeto Any updates on this PR? Thanks again for working on it!

app/controllers/application_controller.rb
```diff
@@ -145,38 +146,10 @@ def most_focus(data_type, profile_id)
...
-      end
-    elsif data_type.in?(%w[mood strategy])
-      moments.find_each do |m|
-        result << m if data_included?(data_type, data_id, m)
-      end
-    end
-    result
-  end
-  # rubocop:enable MethodLength
+  # rubocop:enable MethodLength
```
julianguyen: Very minor, but it looks like there's trailing whitespace on this line, can you remove it?

app/helpers/tags_helper.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+module TagsHelper
+  # rubocop:disable MethodLength
```
julianguyen: I think you can remove this disable comment. You can double check by running rubocop locally.

spec/controllers/application_controller_spec.rb
```diff
@@ -13,47 +13,36 @@
...
     it_behaves_like :most_focus, :strategy
-  end
+  end  
 
-  describe 'tag_usage' do
-    it 'is looking for categories tagged nowhere' do
-      new_category = create(:category, user_id: user1.id)
-      result = controller.tag_usage(new_category.id, 'category', user1.id)
-      expect(result[0].length + result[1].length).to eq(0)
+  describe 'moments_stats' do
```
julianguyen: You can actually get rid of the tests related to moments_stats since it was moved to moments_helper_spec.rb.

Issue #840, opened 2018-03-05
Refactor classes and methods that are too long


nshki:
From #839, many classes and methods were wrapped with `# rubocop:disable MethodLength` and `# rubocop:disable ClassLength` to temporarily make Rubocop happy. Those are all great indicators of refactoring opportunities.

Let's break those classes and methods up into smaller pieces. Could be a good opportunity to apply [DHH's controller development methodology](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/) for controllers?

Files that could use this attention:
- [x] `controllers/application_controller.rb`
- [x] `controllers/allies_controller.rb`
- [x] `controllers/groups_controller.rb`
- [x] `controllers/moments_controller.rb`
- [x] `controllers/categories_controller.rb`
- [x] `controllers/omniauth_callbacks_controller.rb`
- [x] `controllers/strategies_controller.rb`
- [x] `controllers/moods_controller.rb`
- [x] `controllers/meetings_controller.rb`
- [x] `helpers/application_helper.rb`
- [x] `mailers/application_mailer.rb`
- [x] `models/user.rb`

stefannibrasil:
I and @aSquare14 are working on #795 and I did a small refactor on the `controllers/ominiauth_callbacks_controller.rb` and on `model/user.rb`. We are planning to open the PR this week.

But I would like to work more on refactoring others too when we finish this one :)

keitheous:
I would like to have a go at some of these. Let me get the system up and running, then I'll dive into this. Thanks =)

Bennet11:
Hi there, i would also like to try and help on some of these. Is someone already assigned to this?

nshki:
Hi @Bennet11, both @stefannibrasil and @keitheous are working on some at the moment. Perhaps you all could coordinate! We have a Slack channel, which may be the best way to communicate with other contributors. If you'd like to be a part of it, please email join.ifme@gmail.com.

keitheous:
Awesome! Lets get in touch on slack so we do not git conflict. Will you notify me when you're on slack @Bennet11 ? 

I have plans to work with @tsara27 this Sunday on some of the controllers here.

Bennet11:
@nshki Sorry I was not able to respond immediately. I was able to get in touch with Julia last night and she already invited me to join the slack channel

@keitheous that's great to hear! I'm on slack right now my name there is @JohnV

nshki:
@rhiroyuki Go for it!

[...]

PR #1076, opened 2018-10-03
Refactor meetings controller part 1

hmasila:
# Description 
Refactor the meetings controller class to make it thinner and also make the methods shorter

## More Details
Everybody loves clean code.

image

## Corresponding Issue

# Screenshots

# Test Coverage

✅ 
🚫

### DIFF
--- a/app/controllers/meetings_controller.rb
+++ b/app/controllers/meetings_controller.rb
@@ -23,6 +23,7 @@ class MeetingsController < ApplicationController
     comment_for('meeting')
   end
 
+  # rubocop:disable MethodLength
   def delete_comment
     comment_exists = Comment.exists?(id: params[:commentid])
     is_my_comment = Comment.user_comment?(params[:commentid], current_user)
@@ -35,8 +36,7 @@ class MeetingsController < ApplicationController
     end
 
     if comment_exists && ((is_my_comment && is_member) || is_my_meeting)
-      CommentNotificationsService.remove(comment_id: params[:commentid],
-                                         model_name: 'meeting')
+      remove_notification
     end
     head :ok
   end
@@ -55,6 +55,7 @@ class MeetingsController < ApplicationController
 
   # POST /meetings
   # POST /meetings.json
+  # rubocop:disable MethodLength
   def create
     @meeting = Meeting.new(meeting_params)
     leader?(Group.find(meeting_params[:group_id]))
@@ -103,6 +104,7 @@ class MeetingsController < ApplicationController
     end
   end
 
+  # rubocop:disable MethodLength
   def join
     group_id = Meeting.where(id: params[:meeting_id]).first.group_id
     meeting_member = MeetingMember.where(
@@ -142,6 +144,7 @@ class MeetingsController < ApplicationController
     end
   end
 
+  # rubocop:disable MethodLength
   def leave
     meeting_name = Meeting.where(id: params[:meeting_id]).first.name
     group_id = Meeting.where(id: params[:meeting_id]).first.group_id
@@ -188,6 +191,7 @@ class MeetingsController < ApplicationController
 
   # DELETE /meetings/1
   # DELETE /meetings/1.json
+  # rubocop:disable MethodLength
   def destroy
     leader?(@meeting.group)
     # Notify group members that the meeting has been deleted
@@ -235,6 +239,13 @@ class MeetingsController < ApplicationController
       meeting: meeting,
       type: type,
       members: members
+    )
+  end
+
+  def remove_notification
+    CommentNotificationsService.remove(
+      comment_id: params[:commentid],
+      model_name: 'meeting'
     )
   end
 end
--- a/app/controllers/meetings_controller.rb
+++ b/app/controllers/meetings_controller.rb
@@ -204,7 +204,7 @@ class MeetingsController < ApplicationController
   # Use callbacks to share common setup or constraints between actions.
   def set_meeting
     @meeting = Meeting.friendly.find(params[:id])
-  rescue StandardError
+  rescue RecordNotFound
     redirect_to_path(groups_path)
   end
 
@@ -239,9 +239,8 @@ class MeetingsController < ApplicationController
   end
 
   def remove_notification(comment, meeting)
-    if (my_comment?(comment) && member_for(meeting)) || leader_for(meeting)
-      remove_notification!
-    end
+    remove_notification! if (my_comment?(comment) && member_for(meeting)) ||\
+    leader_for(meeting)
   end
 
   def my_comment?(comment)
--- a/app/controllers/meetings_controller.rb
+++ b/app/controllers/meetings_controller.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# rubocop:disable ClassLength
 class MeetingsController < ApplicationController
   include CommentsHelper
 
--- a/app/controllers/meetings_controller.rb
+++ b/app/controllers/meetings_controller.rb
@@ -27,7 +27,7 @@ class MeetingsController < ApplicationController
   # rubocop:disable MethodLength
   def delete_comment
     comment = Comment.find_by(id: params[:commentid])
-    is_my_comment = comment.present? && comment.comment_by == current_user.id
+    is_my_comment = comment.present? && (comment.comment_by == current_user.id)
 
     if comment.present?
       meeting_id = comment.commentable_id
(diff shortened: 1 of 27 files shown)


app/controllers/meetings_controller.rb
```diff
@@ -45,58 +26,40 @@ def comment
...
+      meeting_id = Comment.find(params[:commentid]).commentable_id
+      meeting = Meeting.find(meeting_id)
+      is_my_meeting = MeetingMember.leader?(current_user, meeting)
+      is_member = MeetingMember.member?(current_user, meeting)
     end
 
     if comment_exists && ((is_my_comment && is_member) || is_my_meeting)
-      CommentNotificationsService.remove(comment_id: params[:commentid],
-                                         model_name: 'meeting')
+      remove_notification
```
carlqt: correct me if I'm wrong but in this action, we want to delete a `comment` if it exists AND the comment is the current_user's comment AND the current_user is a member of the meeting right?

app/controllers/meetings_controller.rb
```diff
@@ -45,58 +26,40 @@ def comment
...
-      is_member = MeetingMember.where(
-        meeting_id: meeting_id,
-        user_id: current_user.id
-      ).exists?
-    else
-      is_my_meeting = false
-      is_member = false
+      meeting_id = Comment.find(params[:commentid]).commentable_id
+      meeting = Meeting.find(meeting_id)
+      is_my_meeting = MeetingMember.leader?(current_user, meeting)
```
carlqt: We should be asking the `MeetingMember` object about its state and not the class. At the top of my head, I would do something like:
```ruby
meeting_member = meeting.members.find_by(user: current_user)
```
Chaining methods like this ensures that the record returned is a member of the meeting. Just refactor if gets too long.

Then I would add the methods `#leader?` to the `MeetingMember` model.

app/models/comment.rb
```diff
@@ -44,6 +44,20 @@ def create_from!(params)
         viewers: viewers
       )
     end
+
+    def meeting_comments(meeting)
```
julianguyen: I would unit test these new methods

app/models/group_member.rb
```diff
@@ -32,4 +32,19 @@ def destroy_meeting_memberships
     # are readonly
     MeetingMember.where(id: meeting_membership_ids).destroy_all
   end
+
+  def self.member?(user, group)
```
julianguyen: I would unit test these new methods

app/models/meeting_member.rb
```diff
@@ -18,4 +18,19 @@ class MeetingMember < ApplicationRecord
   belongs_to :meeting, foreign_key: :meeting_id
   belongs_to :user, foreign_key: :user_id
   belongs_to :group_member, foreign_key: :user_id
+
+  def self.member?(user, meeting)
```
julianguyen: I would unit test these new methods

app/controllers/meetings_controller.rb
julianguyen: That sounds correct to me ^

app/controllers/meetings_controller.rb
```diff
@@ -45,58 +26,40 @@ def comment
 
   # rubocop:disable MethodLength
   def delete_comment
-    comment_exists = Comment.where(id: params[:commentid]).exists?
-    is_my_comment = Comment.where(
-      id: params[:commentid],
-      comment_by: current_user.id
-    ).exists?
+    comment_exists = Comment.exists?(id: params[:commentid])
```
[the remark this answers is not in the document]
hmasila: yeah, that makes sense 👍

app/controllers/meetings_controller.rb
```diff
@@ -45,58 +26,40 @@ def comment
...
+      remove_notification
     end
-
     head :ok
   end
-  # rubocop:enable MethodLength
 
   # GET /meetings/new
   def new
-    @group_id = params[:group_id]
```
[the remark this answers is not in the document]
hmasila: I am not sure I fully understand what you mean here. Could you possibly expound or rephrase?

app/controllers/meetings_controller.rb
hmasila: I am not sure what to do with this comment. Is there something I should add or change?

carlqt: Sorry for the confusion, I'm just trying to clarify what this action is supposed to do as I feel there's definitely room for improvements. I apologise that I'm unable to point it out right.

Will let you know once I got an idea and if this PR isn't merged by then.

app/controllers/meetings_controller.rb
hmasila: I am trying to convert 

```ruby
MeetingMember.find_by(
        id: current_user.id, leader: true
      )
```
into the above format but it fails because the result of `meeting.members.find_by(current_use.idr)` is`User` not `MeetingMember` - so determining if it's leader or not might be a little bit hard

app/controllers/meetings_controller.rb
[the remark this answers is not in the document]
julianguyen: Oh sorry, I meant that you can get rid of the reference to @group_id in meetings_helper.rb since it's no longer defined.

app/controllers/meetings_controller.rb
```diff
@@ -45,72 +26,52 @@ def comment
 
   # rubocop:disable MethodLength
   def delete_comment
-    comment_exists = Comment.where(id: params[:commentid]).exists?
-    is_my_comment = Comment.where(
-      id: params[:commentid],
-      comment_by: current_user.id
-    ).exists?
+    comment = Comment.find_by(id: params[:commentid])
+    is_my_comment = comment.present? && (comment.comment_by == current_user.id)
```
jm96441n: can we extract this boolean into it's own method and add a `?` to the end of `is_my_comment` to specify it's a boolean value?

app/controllers/meetings_controller.rb
```diff
@@ -45,72 +26,52 @@ def comment
...
+      is_my_meeting = MeetingMember.find_by(
+        id: current_user.id, leader: true
+      )
+      is_member = meeting.members.find_by(id: current_user).present?
     end
 
-    if comment_exists && ((is_my_comment && is_member) || is_my_meeting)
-      CommentNotificationsService.remove(comment_id: params[:commentid],
-                                         model_name: 'meeting')
+    if comment.present? && ((is_my_comment && is_member) || is_my_meeting)
```
jm96441n: same here, can we extract this boolean statement into it's own method so this is a bit more clear as to the intent of this

app/controllers/meetings_controller.rb
```diff
@@ -259,29 +210,48 @@ def destroy
   private
 
   # Use callbacks to share common setup or constraints between actions.
-  # rubocop:disable RescueStandardError
   def set_meeting
     @meeting = Meeting.friendly.find(params[:id])
-  rescue
+  rescue StandardError
```
jm96441n: are we rescuing that the record is not found? Can we rescue that error instead so we're not potentially swallowing other errors?

app/controllers/meetings_controller.rb
hmasila: yeah, that sounds like a good idea

app/controllers/meetings_controller.rb
hmasila: Honestly, I am not sure what we are rescuing here. It was a general `rescue` before so I changed it to `StandardError` which is no better but more of a preference issue. Maybe @juliamji could provide us with some answers

julianguyen: I think you're referring to the wrong person... but yeah looks like the record not being found is being rescued.

hmasila: oops, my bad. Okay in that case we can rescue from `RecordNotFound` instead

app/controllers/meetings_controller.rb
```diff
@@ -259,29 +201,60 @@ def destroy
...
-      leader: true
-    )
-    return if group_member.exists?
+  def leader?(group)
+    return if GroupMember.find_by(
+      user_id: current_user.id, group_id: group.id, leader: true
+    ).present?
 
-    redirect_to_path(group_path(group_id))
+    redirect_to_path(group_path(group.id))
```
jm96441n: I know this was here when you started this refactor, but can this redirect be part of the code path in the destroy method? This reads initially that the `leader?` method should return a boolean, so maybe have it just run the query in the guard clause and then the redirect logic can go in the destroy method

PR #1078, opened 2018-10-03
Fix lingering categories

### LINKED ISSUE #1077: Lingering categories bug
# Description 

Categories are prone to linger on Strategy objects after being deleted, which causes error screens indefinitely for a user when trying to access their dashboard.

# Reproduction Steps

1. Login
2. Create a new Strategy and attach a Category to it
3. Delete that Category
4. Access the Strategies screen

✅

nshki:
# Description 

Fixes lingering categories bug.

## More Details

- Adds a test to check for lingering categories on Strategy objects
- Implements fix for the test
- Bypasses lingering categories in UI
- Adds rake task to clean up lingering categories

## Corresponding Issue

Fixes #1077 

# Test Coverage

✅

### DIFF
--- a/app/controllers/concerns/shared.rb
+++ b/app/controllers/concerns/shared.rb
@@ -48,8 +48,12 @@ module Shared
 
   def shared_destroy(model_object, model_name)
     moments = current_user.moments.all
+    strategies = current_user.strategies.all
     moments.each do |moment|
       update_moment(model_object, model_name, moment)
+    end
+    strategies.each do |strategy|
+      update_strategy(model_object, model_name, strategy)
     end
     model_object.destroy
     redirect_to_path(index_path(model_name))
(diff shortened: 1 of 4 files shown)


lib/tasks/cleaner.rake
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+namespace :cleaner do
```
julianguyen: I wonder if we want this to be a one-time migration instead of a more "permanent" rake task 🤔 Folks who are more experienced with Rails, can you share your thoughts?

app/controllers/concerns/shared.rb
```diff
@@ -64,6 +68,13 @@ def update_moment(model_object, model_name, moment)
     Moment.find_by(id: moment.id).update(params)
   end
 
+  def update_strategy(model_object, model_name, strategy)
```
julianguyen: Not mandatory: update_moment and update_strategy look pretty similar, we could combine them and have some conditional logic for whether to use Moment.find_by or Strategy.find_by

Issue #1052, opened 2018-10-02
Add missing test coverage for reminder_helper.rb


julianguyen:
# Description 

Womp womp we're missing test coverage here. Let's go back and add it!

# Do you want to be the assignee to work on this?

🚫

ScotianMan:
this wasnt assigned so Ive knocked it out while poking around this morning.

PR #1079, opened 2018-10-03
Add specs for reminder_helper, refactor methods in reminder_helper fo…

ScotianMan:
# Description 

Adds a new spec file for the reminder_helper as well as a small refactor of reminder_helper to make for easier testing

## Corresponding Issue

#1052 

# Test Coverage

✅


app/helpers/reminder_helper.rb
```diff
@@ -4,22 +4,31 @@ module ReminderHelper
...
+      reminders = format_reminders(names_arr)
     end
 
     reminders.html_safe
   end
 
   private
 
+  def reminders?(data)
+    data.methods.include?(:active_reminders) && data.active_reminders&.any?
```
julianguyen: Could you explain why you're also checking for data.methods to include active_reminders?

spec/helpers/reminder_helper_spec.rb
```diff
@@ -0,0 +1,57 @@
+describe ReminderHelper do
+  name1 = 'name1'
+  name2 = 'name2'
+  reminder_names = [
+    name1, name2
+  ]
+
+  describe '#reminders?' do
+    it 'returns false when data has no active_reminder method' do
+      expect(reminders?({})).to be_falsey
```
julianguyen: Why not explicitly check for whether it's true or not?

spec/helpers/reminder_helper_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    it 'returns false when data has no active reminder' do
+      user = FactoryBot.create(:user)
```
julianguyen: You could move up the creation of a user into a let

app/helpers/reminder_helper.rb
ScotianMan: Just a sanity check so this method does not have to rely on someone passing the correct data into it. If the data in this case doesnt have a method `active_reminders` the safe nav `&` didnt catch it. This could be caught in other specs but since I wasnt sure I opted to over test it here

spec/helpers/reminder_helper_spec.rb
ScotianMan: Bad habit honestly. Since we are infact no interested in the nil case here checking true is the better choice hehe

spec/helpers/reminder_helper_spec.rb
ScotianMan: moved

app/helpers/reminder_helper.rb
```diff
@@ -4,22 +4,31 @@ module ReminderHelper
...
+      names_arr = data.active_reminders.map(&:name)
+      reminders = format_reminders(names_arr)
     end
 
     reminders.html_safe
   end
 
   private
 
+  def reminders?(data)
```
jm96441n: can this be renamed so it's a bit more clear as to the intent of the method? This reads to me as "does reminders include this data" and it took a little bit to unpack it

ScotianMan: It was called `has_reminders?` and rubocop yelled at me to remove the has on methods returning bools

jm96441n: yeah, rubcop will do that, since this is checking for active reminders maybe `data_has_active_reminders?` or something like that?

jm96441n: Or maybe change the parameter to be something a bit more descriptive than `data` (tbh I'm not too familiar with the code so `data` might be the most descriptive name for it), but something to make this a bit more human readable

ScotianMan: Ya I didnt look too far into why rubocap complained yet. I only hopped on this repo yesterday. Changing the `data` name would be getting a little out of the scope of the issue since this was just for specs. I kinda pushed it refactoring some of the helper to make the specs more concise but I agree `data` is a bit ambiguous. It appears that data can either me medications or strategies. All for debating the semantics of the method name I made though lol

jm96441n: I just jumped into the repo yesterday as well! Rubocop usually yells about methods that being with `has`, since this appears to be checking if data has any active reminders, maybe this should be `active_reminders?(data)` or `contains_active_reminders?` or `active_reminders_present?`, something to convey that this is checking that the data argument contains active reminders

ScotianMan: haha nice. I diden even realize it was hacktoberfest, GH has been a disorganized mad house this week 😂 

I think `active_reminders?` is totally appropriate and will appease rubocop 👍

Issue #1075, opened 2018-10-03
Only deploy to production (Heroku) via CircleCI if a git tag exists


julianguyen:
# Description 

This is related to discussions involving [staging servers](https://github.com/ifmeorg/ifme/issues/1055). Before we set up staging servers, we should improve our production deployment process. Currently, we deploy every time there's a push to master. While this was okay in the past, there are certain times of the year where we get frequent (almost daily) merges to master (yay!)... which means A LOT of deploys. Sometimes we don't want to deploy right away if the change is minor. Setting up tags offers flexibility, we can decide on the spot whether we want to deploy the last x changes without having to work against a fixed deploy schedule.

We also record a [manual changelog](https://github.com/ifmeorg/ifme/wiki/Changelog) for major updates and often forget to update it. Using git tags and releases will also tackle that problem.

Let's also enforce [Semver](https://semver.org/) when it comes to versioning.

# Do you want to be the assignee to work on this?

🚫 This would be a great opportunity for someone who's interested! I'm happy to help (if needed) as I have experience writing CircleCI configs.

tryantwit:
I'll take a crack at this. CircleCI 2.0 has support for [Git Tag Filters](https://circleci.com/docs/2.0/configuration-reference/#tags) so should be pretty easy.

PR #1082, opened 2018-10-03
Create two separate workflows for circleci.

tryantwit:
# Description

Workflow that deploys requires a git tag to run. Adding a tag filter
would require all other tasks to have a tag as well. Splitting into two
workflows, the build-test flow can work on all branches/commits
while build-test-deploy only works when a git tag matching vX.X.X is pushed.

## More Details
[Git Tag Documentation](https://circleci.com/docs/2.0/workflows/#git-tag-job-execution)

According to the documentation if using tag filters and a job requires any other job directly or indirectly, those jobs must also require regular expression filters for tags.

## Corresponding Issue

#1075 

# Screenshots

# Test Coverage

🚫

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,7 +130,7 @@ jobs:
             git push https://heroku:$HEROKU_API_KEY@git.heroku.com/ifme.git master
 workflows:
   version: 2
-  build-deploy:
+  build-test:
     jobs:
       - setup-coverage
       - build:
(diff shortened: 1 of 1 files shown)


julianguyen:
Thanks sooo much for working on this! Question, does the regex for the tag version just mean we can only have single digits e.g. 2.2.0 but not 13.2.0?

.circleci/config.yml
```diff
@@ -139,9 +139,30 @@ workflows:
...
+              only: /^v[0-9]+\.[0-9]+\.[0-9]+/
       - deploy:
           requires:
             - upload-coverage
           filters:
             branches:
-              only: master
+              ignore: /.*/
+            tags:
+              only: /^v[0-9]+\.[0-9]+\.[0-9]+/
```
julianguyen: Do we just want to have tags in the deploy job? Could you clarify why it's in the other jobs?

julianguyen: Thanks for explaining!

PR #1083, opened 2018-10-03
Target Rails version

huyvohcmc:
# Description 

- Target Rails version to 5.2
- Don't require rubocop by default

# Test Coverage

✅

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,6 @@
 AllCops:
   TargetRubyVersion: 2.3
+  TargetRailsVersion: 5.2
   Exclude:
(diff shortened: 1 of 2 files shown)


julianguyen:
Could you explain why we would not want to require rubocop by default?

camillevilla:
Hi @huyvohcmc! Pinning a Rails version in our Rubocop config adds a source of potential drift over time. For example, when we upgrade to the next minor Rails version (which will probably happen automatically, via Dependabot), someone will need to remember to manually bump our Rubocop target rails version as well. We'll be sticking with our current config for now.

Issue #1054, opened 2018-10-02
Add missing test coverage for group_notifier.rb


julianguyen:
# Description 

Womp womp we're missing test coverage here. Let's go back and add it!

# Do you want to be the assignee to work on this?

🚫

julianguyen:
Hey @jm96441n, thanks for expressing your interest. We would love your help with this issue. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :) After that you can assign yourself to this issue.

PR #1087, opened 2018-10-03
[Issue #1054] Added spec coverage for group_notifier

jm96441n:
added specs for group notifier

# Description 

This PR introduces spec coverage for the `group_notifier` service.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1054

# Screenshots

N/A

# Test Coverage

✅

### DIFF
--- a/spec/services/group_notifier_spec.rb
+++ b/spec/services/group_notifier_spec.rb
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 describe GroupNotifier do
-  let(:current_user) { double(User, name: 'user', id: 1) }
-  let(:group) { double(Group, name: 'group', id: 1) }
+  let(:current_user) { build(:user) }
+  let(:group) { build(:group) }
   let(:type) { 'notifier' }
-  let(:recipient) { double(User, id: 2) }
+  let(:recipient) { build(:user) }
   let(:recipients) { [recipient] }
-  let(:new_notifications) { [double(Notification)] }
+  let(:new_notifications) { [build(:notification)] }
   let(:client) { double('client') }
   let(:mailer) { double('mailer') }
   let(:data) do
--- a/spec/services/group_notifier_spec.rb
+++ b/spec/services/group_notifier_spec.rb
@@ -20,6 +20,10 @@ describe GroupNotifier do
   end
   subject { GroupNotifier.new(group, type, current_user) }
 
+  after do
+    subject.send_notifications_to(recipients)
+  end
+
   it 'creates push notifications' do
     allow(subject).to receive(:create_notification)
     allow(subject).to receive(:send_email_notification)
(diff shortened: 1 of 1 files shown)


spec/services/group_notifier_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+describe GroupNotifier do
+  let(:current_user) { double(User, name: 'user', id: 1) }
```
julianguyen: You can actually use existing FactoryBot values for user and group, but doubles do work too

jm96441n: that's true, I usually prefer using doubles/mocks but if it's preferred I can use FactoryBot here

spec/services/group_notifier_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      uniqueid: "notifier_#{current_user.id}"
+    }.to_json
+  end
+  subject { GroupNotifier.new(group, type, current_user) }
+
+  it 'creates push notifications' do
+    allow(subject).to receive(:create_notification)
+    allow(subject).to receive(:send_email_notification)
+    allow(Notification).to receive_message_chain(:where, :order).and_return(new_notifications)
+    allow(Pusher).to receive(:[]).with("private-#{recipient.id}").and_return(client)
```
julianguyen: Could this be `Array` instead of `:[]`?

spec/services/group_notifier_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    expect(mailer).to receive(:deliver_now)
+  end
+
+  it 'creates notifications' do
+    allow(subject).to receive(:push_notifications)
+    allow(subject).to receive(:send_email_notification)
+    expect(Notification).to receive(:create).with(user_id: recipient.id, uniqueid: "notifier_#{current_user.id}", data: data)
+  end
+
+  after do
```
julianguyen: I usually put before and after blocks at the top of the file. I've never seen an after at the end, but it makes sense actually.

spec/services/group_notifier_spec.rb
jm96441n: TBH I'm not sure, I'll give it a try

jm96441n: Actually TIL so ruby access hashes using `.[]('key-here')` as a method:
screenshot from 2018-10-04 20-44-46

jm96441n: works the same for arrays as well

spec/services/group_notifier_spec.rb
julianguyen: Let's replace things with FactoryBot where it already exists. Thanks!

jm96441n: Can do! I'll make those changes tonight

julianguyen:
Thanks so much for working on this! Awesome work :) Feel free to merge whenever ready!

jm96441n:
Awesome, thanks for everyone's feedback on this!

Issue #1066, opened 2018-10-02
Create a bot or find an existing one that comments on first time comments in Github issues


julianguyen:
# Description 

Right now we manually reply to people who respond to issues with "I wanna work on this!". Let's find a way to automate this. The message could be something like:

Hey `@some-user`, thanks for expressing your interest. We would love your help with this issue. Please follow these [next steps](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack) to join our contributor community.

The existing welcome bots we have set up just comment on first time PRs, issue creations, and PR merges. We use Probot and they're awesome :)

If you work on this bot, you can totally open source this beyond our project and feature your bot on Probot 😉 

# Do you want to be the assignee to work on this?

🚫

chischaschos:
I wanna work on this!

austinmckinley:
I remember seeing this a couple days ago, I think this might do the trick? https://github.com/apps/the-welcome-bot

julianguyen:
Hey @chischaschos if you're planning on working on this, can you please assign yourself to this issue. If you're unable to do so then follow these [next steps](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

chischaschos:
@austinmckinley, @julianguyen  Yeah I'd like to do it, I want to learn about Probot.

grust101:
@chischaschos, I've assigned you and myself to this issue. My thoughts are to write a probot script that is listening for an event where someone comments using the phrase "I want to work on this" and then use will auto comment with instructions to join the slack/etc and can also assign them to the issue. Thoughts?

ScotianMan:
well... Ive just kept picking away lol I believe the bot is ready... with a couple caveats and things I want to work on so this can be used for any repo but I already abstracted much of the functionality. 2 things of note are the message is hardcoded right now and specific to this repo as well as the bot will continue to post the message when a non member continues to comments until they sign up. Both of these tweaks can probably happen later tonight or tomorrow but I think its ready to give it a test on a repo thats not my test repo. App is living on Heroku and is accepting webhooks. As far as a PR goes I can gear up some docs and instructions 👍 here is a quick video https://screencast.com/t/qcfy3yPFG

grust101:
[...]

PR #1089, opened 2018-10-04
1066-add-membership-bot

ScotianMan:
# Description 

Adds a config var so member-bot can use a custom link to a join us page

## Corresponding Issue

#1066 

# Screenshots
this is not of the PR work but of the functionality of the bot
 https://screencast.com/t/qcfy3yPFG

# Test Coverage

🚫 

there is only a config var added. The testing really needs to be added to the member-bot it self. Ill work on that at some point

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -25,3 +25,7 @@ newPRWelcomeComment: >
 # Comment to be posted to on pull requests merged by a first time user
 firstPRMergeComment: >
   Thank you for merging this pull request with us! If you haven't already, in another pull request, please [add yourself](https://github.com/ifmeorg/ifme/wiki/Contributor-Blurb) to our [Contribute](https://www.if-me.org/contribute) page.
+
+# Link for member-bot to use for joining us
+howToJoinLink: >
+  https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack


.github/config.yml
```diff
@@ -25,3 +25,7 @@ newPRWelcomeComment: >
 # Comment to be posted to on pull requests merged by a first time user
 firstPRMergeComment: >
   Thank you for merging this pull request with us! If you haven't already, in another pull request, please [add yourself](https://github.com/ifmeorg/ifme/wiki/Contributor-Blurb) to our [Contribute](https://www.if-me.org/contribute) page.
+
+# Link for member-bot to use for joining us
+howToJoinLink: >
```
hmasila: I think we can also include some good message like 
> "Thank you for showing interest in joining our community. For further steps visit ...."

ScotianMan: definitely 👍 although that is a change I will need to make over in the member-cop repo before it can be added here for the app to use. Id like to get this PR in and tested on a live repo before I start shining up member-cop

ScotianMan: Here is a video. I can @ mention people and use html https://www.screencast.com/t/qcfy3yPFG

ScotianMan: changing this message text is no longer something that should be done in this PR. Probably best I make a PR in the member-cop repo with the update to use custom messages with the link then we will need a quick PR here to change this file to member-bots new requirements.

julianguyen: Nice! That's cool you can @ people :)
Do you have to add the message in the member-cop repo because @ people?

ScotianMan: correct. Although when I change it I plan to allow people to placeholder in some things like that. Maybe somthing like "welcome [@user] please read [@link] for more info"

julianguyen: I wonder if you can concatenate what's in the member-cop repo and what's in this config file.
I don't want to limit your bot to just displaying our welcome message. You could just have `@user` in your member-cop repo and then our welcome message can be in this config?

julianguyen: Maybe have a flag in the config too? Like if includeUser is true then concatenate so that the final message is `@user Thanks for expressing your interest. We would love your help with this issue. Please follow these [next steps](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack) to join our contributor community.`

ScotianMan: yup its doable 👍 Once this is merged and I know its working I can fine tune those things with a PR in member-cop and 1 here to change the config. Right now member-bot has a generic welcome message that works for any repo, grabs the user from the webhook and the link from the config file concats some things and posts the comment

ScotianMan:
can we get this merged in @julianguyen? Ive starting working on some customization to member-bot but in the mean time the link isnt working and member-bot is posting msgs to users that they cant follow to the next steps

Issue #995, opened 2018-09-23
Add ability to copy to clipboard when selecting the secret share input field


julianguyen:
# Description 

Secret share is awesome feature our Rails Girls Summer of Code 2017 students @SophieRoseMcDonald and @jnamster worked on! It adds the ability to generate a private URL to your Moment post that you can share with anyone who is not signed into the app. You're also able to remove the generated URL when you want it to expire.

We want to add the ability when the user clicks into the input field to have the URL automatically copied to the user's clipboard. You'll want to create a special type in our Input component (React) to support copy to clipboard. You use our Tooltip component to display something like "Copied to clipboard" (please use i18n) once the user clicks the input field.

# Screenshots

![image](https://user-images.githubusercontent.com/3010728/45747192-e3344e80-bbb9-11e8-9fb6-97e9252502cc.png)

# Do you want to be the assignee to work on this?

🚫

namila:
I will work on this and make a PR

julianguyen:
Hey @namila, thanks for expressing your interest. We would love your help with this issue. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :)

ScotianMan:
Anything happening with this? Ill knock this out this morning 👍

ScotianMan:
not exactly as requested but it works and gets the job done for now and we can discuss.

dandida95:
i would like to contribute

PR #1096, opened 2018-10-05
995-click-to-copy-input

ScotianMan:
# Description 

Adds click to copy functionality to the secret share url created on the moments page. The back end is set up for i18n and the var is set for the front end but react is not playing nicely so for now the confirmation toast will be in english but the tool tip is international providing someone translates. I added a dependency for `react-toastify` but I couldnt find we had any toast implemented at all and using a tooltip didnt feel/look right. 

## Corresponding Issue

#995 

# Screenshots

https://screencast.com/t/k4vbGGmAVsYd

# Test Coverage

✅ however a feature test confirming the toast firing and clipboard copy could be added.

### DIFF
--- a/app/assets/stylesheets/secret_share/secret_share_content.scss
+++ b/app/assets/stylesheets/secret_share/secret_share_content.scss
@@ -7,11 +7,9 @@
 }
 
 .secretShare {
-  form {
-    #secretShareLink {
+  #secretShareLink {
       background-color: $white;
       color: $mulberry;
       font-size: 1em;
-    }
   }
 }
--- a/app/assets/stylesheets/secret_share/secret_share_content.scss
+++ b/app/assets/stylesheets/secret_share/secret_share_content.scss
@@ -5,3 +5,13 @@
   background: $white-80;
   @include dashboardSectionStyles();
 }
+
+.secretShare {
+  form {
+    #secretShareLink {
+      background-color: $white;
+      color: $mulberry;
+      font-size: 1em;
+    }
+  }
+}
--- a/app/helpers/moments_helper.rb
+++ b/app/helpers/moments_helper.rb
@@ -29,7 +29,8 @@ module MomentsHelper
         name: 'secretShareInput',
         readOnly: true,
         value: secret_share_url(moment.secret_share_identifier) || nil,
-        dark: true
+        dark: true,
+        copyOnClick: true
       }
     ], action:  moment_path(moment) }
   end
--- a/app/helpers/moments_helper.rb
+++ b/app/helpers/moments_helper.rb
@@ -15,7 +15,34 @@ module MomentsHelper
     edit_form_props(moment_form_inputs, moment_path(@moment))
   end
 
+  def secret_share_url_props
+    { inputs: [
+        {
+          id: 'secretShareLink',
+          type: 'text',
+          name: 'secretShareInput',
+          readOnly: true,
+          value: secret_share_url(@moment.secret_share_identifier) || nil,
+          dark: true
+        }
+      ], action:  moment_path(@moment)
+    }
+  end
+
   private
+
+  def copy_input
+    [
+      {
+        id: 'myId',
+        type: 'text',
+        name: 'myName',
+        readOnly: true,
+        value: secret_share_url(@moment.secret_share_identifier) || nil,
+        dark: true
+      }
+    ]
+  end
 
   # rubocop:disable MethodLength
   def moment_form_inputs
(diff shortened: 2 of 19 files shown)


ScotianMan:
Anyone know why there are controller specs in the helpers spec? Its causing a conflict, I can resolve it easily enough but should there be controller specs in a helper file on master?
/cc @julianguyen

client/app/components/Input/InputDefault.jsx
```diff
@@ -79,6 +89,10 @@ export const InputDefault = (props: Props) => {
       }
       ref={myRef}
       aria-label={label}
+      onClick={(e: SyntheticEvent<HTMLInputElement>) => copyToClipBoard(e)}
```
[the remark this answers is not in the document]
julianguyen: We should write an integration test for this in InputDefault.spec.jsx

julianguyen:
As for the merge conflicts, I would just resolve them for now and you can open up a new issue if you'd like to tackle the separation of concerns issue you raised. Thanks!

client/app/components/Input/InputDefault.jsx
```diff
@@ -21,6 +23,14 @@ export type Props = {
   hasError?: Function,
   myRef?: any,
   label?: string,
+  onClick?: Function,
+};
+
+const copyToClipBoard = (e: SyntheticEvent<HTMLInputElement>) => {
+  e.currentTarget.select();
+  document.execCommand('copy');
+  // TODO: Once i18n and react are playing nicely this can be changed
```
julianguyen: Did you encounter issues as well?

client/app/components/Input/InputDefault.jsx
```diff
@@ -60,7 +70,7 @@ export const InputDefault = (props: Props) => {
   } = props;
   if (!REQUIRES_DEFAULT.includes(type)) return null;
   return (
-    <input
+    <><input
```
julianguyen: I'm wondering why `<>` was added.

app/helpers/moments_helper.rb
```diff
@@ -15,8 +15,25 @@ def edit_moment_props
     edit_form_props(moment_form_inputs, moment_path(@moment))
   end
 
+  def secret_share_url_props(moment)
```
jm96441n: why is this here just to call a private method? why not expose that method?

spec/helpers/moments_helper_spec.rb
```diff
@@ -0,0 +1,23 @@
+describe MomentsHelper do
+  let(:user) { FactoryBot.create(:user) }
+  let(:moment) { FactoryBot.create(:moment, :with_secret_share, user: user) }
```
jm96441n: do these need to be persisted? can we use `build` instead and stub the DB calls?

app/assets/stylesheets/secret_share/secret_share_content.scss
```diff
@@ -5,3 +5,13 @@
   background: $white-80;
   @include dashboardSectionStyles();
 }
+
+.secretShare {
+  form {
```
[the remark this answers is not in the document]
ScotianMan: sure does, fixed 😜 That form was overriding everything on me for nearly 20 minutes and when I made the css finally listen to me I never looked back at fixing it. CSS and I have a love hate relationship 😂

client/package.json
```diff
@@ -103,7 +104,7 @@
   },
   "jest": {
     "moduleNameMapper": {
-      "\\.scss$": "identity-obj-proxy"
+      "\\.(css|scss)$": "identity-obj-proxy"
```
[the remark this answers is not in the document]
ScotianMan: The toastify css was not being parsed by Jest, this was why :-)

client/app/components/Input/InputDefault.jsx
ScotianMan: so I was only returning one element. I changed it to a div

app/helpers/moments_helper.rb
ScotianMan: To protect this trivial logic and future logic being used to generate the props as well as to keep the style of the file in line with the other functions in this helper. Is there a reason not to? style or security?

jm96441n: I'd say more in terms of OO design. Generally private methods are in place to hide implementation details that aren't necessary for the public interface of an object, but since the only thing this method does is call a private method (which returns a hash, so the internals of the method are still exposed) it doesn't necessarily need to be private. Also, in ruby, `private` methods shouldn't really be used as a form of security since they aren't really private as in other OO languages

julianguyen:
Just checking in - any updates on this?

julianguyen:
I'm going to finish this issue and merge it into `finish-click-to-copy-input`

PR #1103, opened 2018-10-05
Setup Sentry for better error monitoring

julianguyen:
# Description 

We currently use Papertrail on production. While that is helpful for full error logs it would be great to also pool together error in easier to read/organize/categorize reports and give access to that account to multiple people in our org (we can't do that with Papertrail in Heroku)

## More Details

- Covers monitoring for both Rails and React
- In addition to this setup, already created an account and set up the environment variable in Heroku
- Remove some unnecessary comments

# Test Coverage

🚫 Tested manually

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,6 +23,7 @@ class ApplicationController < ActionController::Base
                        if: proc { |c| c.request.format == 'application/json' }
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :if_not_signed_in, unless: :devise_controller?
+  before_action :set_raven_context, if: proc { Rails.env.production? }
 
   # i18n
   before_action :set_locale
@@ -229,6 +230,11 @@ class ApplicationController < ActionController::Base
 
   def user_moments(user_id)
     Moment.where(user_id: user_id)
+  end
+
+  def set_raven_context
+    Raven.user_context(id: current_user.id)
+    Raven.extra_context(params: params.to_unsafe_h, url: request.url)
   end
 end
 # rubocop:enable ClassLength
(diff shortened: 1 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -230,5 +231,10 @@ def user_strategies(user_id)
   def user_moments(user_id)
     Moment.where(user_id: user_id)
   end
+
+  def set_raven_context
+    Raven.user_context(id: current_user.id)
+    Raven.extra_context(params: params.to_unsafe_h, url: request.url)
```
julianguyen: Using to_unsafe_h was suggested in the Sentry official setup docs lol.

config/environments/production.rb
```diff
@@ -123,4 +123,8 @@
   # subsequently, this will cause the SSL certs registered for our private DNS address to fail.
   config.action_controller.default_url_options = { host: primary_domain }
   config.action_controller.asset_host = primary_domain
+
+  Raven.configure do |config|
```
denysmedynskyi: What do you think about putting this block and https://github.com/ifmeorg/ifme/pull/1103/files#diff-58da95d5e273d16d883532efa37ca499 together. 
So we would have:
```
Raven.configure do |config|
  config.dsn = ENV['SENTRY_DSN']
  config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
end
```

app/controllers/application_controller.rb
julianguyen: Hmm I don't know if we should expose email and id together in the error logs. Even though our app isn't officially HIPAA compliant, I don't want potentially sensitive user information to show up in the logs that is identifiable. But then again, only specific/requested contributors in our org would have access... 🤔

config/environments/production.rb
julianguyen: Ooh, that would technically deduce the need for having that initializer file. Good idea!

Issue #1104, opened 2018-10-05
Flakey feature tests


jm96441n:
# Description 

We have a few feature specs that will fail due to the header menu being hidden in a hamburger menu when the window is not large enough. 

# Reproduction Steps

add to your `spec_helper.rb` file (this will force the window to resize to a width that will hide the header)
```
config.before(js: true) do
  page.current_window.resize_to(500, 1_200
end
```
# Screenshots

N/A

# Do you want to be the assignee to work on this?

✅

PR #1106, opened 2018-10-06
#1104 - Flakey Feature Tests

jm96441n:
# Description 

I noticed a few feature tests would fail if the browser window opened was not at a certain size due to certain elements that would be visible on a larger window were hidden into menus at smaller resolutions. This PR moves to either open the header for scenarios where that is an issue, or clicks an element on the page that is there at all screen sizes.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1104

# Screenshots

N/A

# Test Coverage

✅

### DIFF
--- a/spec/features/persisting_browser_locale_after_sign_in_spec.rb
+++ b/spec/features/persisting_browser_locale_after_sign_in_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.feature 'Persisting browser locale after sign in', type: :feature do
-  scenario 'When user sign in, sign out, then change local', js: true do
+  scenario 'When user sign in, sign out, then change local', js: true, header: true do
     user = create :user
 
     login_as user
@@ -12,8 +12,13 @@ RSpec.feature 'Persisting browser locale after sign in', type: :feature do
       'your mental health.'
     )
 
+    open_header_if_hidden
+
     within('#header') { find('a[href="/users/sign_out"]').click }
-    expect(find('#header')).to have_content('SIGN IN')
+
+    open_header_if_hidden
+
+    expect(find('#header')).to have_content(/sign in/i)
 
     change_language('es')
 
--- a/spec/features/toggle_language_spec.rb
+++ b/spec/features/toggle_language_spec.rb
@@ -51,11 +51,13 @@ describe 'ToggleLanguage', js: true do
         change_language('en')
       end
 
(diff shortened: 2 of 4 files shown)


spec/features/toggle_language_spec.rb
```diff
@@ -117,14 +119,14 @@
...
 
-      it 'persists locale selection on a different page' do
+      it 'persists locale selection on a different page', feature: true do
         expect(find('.pageTitle')).to have_content en_signed_in_root_title
         change_language('es')
         expect(find('.pageTitle')).to have_content es_signed_in_root_title
 
         change_page(
           lambda {
-            within '.dashboardNav' do
```
jm96441n: The `Dashboard Nav` section is not shown at smaller resolutions. Not sure if this is a bug or not, but the content is still accessible through the main part of the dashboard

spec/features/persisting_browser_locale_after_sign_in_spec.rb
```diff
@@ -12,8 +12,13 @@
...
     )
 
+    open_header_if_hidden
+
     within('#header') { find('a[href="/users/sign_out"]').click }
-    expect(find('#header')).to have_content('SIGN IN')
+
+    open_header_if_hidden
+
+    expect(find('#header')).to have_content(/sign in/i)
```
jm96441n: Tthis text is capitalized when shown in the header and titleized when accessed through the hamburger menu, this searches via a case-insensitive regex to find the the link to sign in

spec/features/toggle_language_spec.rb
julianguyen: Yep, that's correct, it's not visible by CSS in smaller resolutions but the content is technically still accessible.

spec/support/hidden_header_support.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module HiddenHeaderSupport
+  def open_header_if_hidden
```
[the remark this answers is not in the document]
jm96441n: I like that, it's much clearer as to the intent

spec/features/toggle_language_spec.rb
jm96441n: these failures here with the dashboard actually aren't related to the header, after getting the parts where accessing the header was working I ran into issues on the dashboards after signing in, let me make a quick gif to show where this was occurring

PR #1108, opened 2018-10-06
Switch component and fix i18n in client

julianguyen:
# Description 

- Switch component (a11y friendly)
- Update forms
- Remove react-intl because there are problems with it working in Rails
- Leverage existing i18n-js setup to get i18n to client
- Replace TODOs in client for i18n

## Corresponding Issue

#780

# Screenshots

capture

image

# Test Coverage

✅


client/app/stories/Checkbox.stories.jsx
```diff
@@ -7,7 +7,7 @@ function handleCheckboxClick(allChecked) {
   window.alert(`Here's an example of an action: ${allChecked}`);
 }
 
-storiesOf('Checkbox', module).add('Checkbox', () => (
+storiesOf("Checkbox (don't use)", module).add('Checkbox', () => (
```
julianguyen: Was created for RGSoC, needs to be integrated in new Input component

config/locales/en.yml
```diff
@@ -11,6 +11,10 @@ en:
  click_here: "click here"
  less: " [Less]"
  language: 'Language'
+ true: 'yes'
```
julianguyen: I was encountering some weird behaviour where when I specified yes and no to be the i18n keys, they got converted to true and false.

julianguyen: Just saw found this: https://github.com/svenfuchs/i18n/issues/94
Could work if I did `'yes': 'yes'` for instance 🤔

client/app/components/Input/InputSwitch.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+export type State = {
+  checked: boolean,
+  key?: string,
+};
+
+export class InputSwitch extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = { checked: !!props.checked };
```
nshki: This is actually just a question out of curiosity on my part, not a suggestion or anything! What's the advantage of using `!!` here when that particular prop is defined as a boolean via Flow?

client/app/components/Input/InputSwitch.scss
```diff
@@ -0,0 +1,38 @@
+@import "../../styles/_global.scss";
+
+.switch {
+  @include setFontSize($size-20);
```
nshki: Was thinking we can bump this down a tad bit! Was initially thinking 14px but let me know if that seems too tiny.

Was also thinking of bumping the letter spacing by a tad too -- was thinking 0.95em.

client/app/components/Input/InputSwitch.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  &On {
+    @include linearTransition(0.5s);
+    background: $key-lime;
+  }
+
+  &Off {
+    @include linearTransition(0.5s);
+    background: $mulberry-70;
+    justify-content: flex-end;
```
nshki: Was actually wondering what you think about having the off state have a left-aligned toggle knob, and the on state have a right-aligned toggle knob. 🤔

Was a bit unaccustomed to seeing it in reverse.

client/app/components/Input/InputSwitch.jsx
julianguyen: Good question, it's because this.props.checked can actually be undefined or a boolean. For the generic Input component, this component is a child of, checked and most of the props are optional. this.state.checked in this particular component is a boolean so I needed a way of converting undefined to false.

client/app/components/Input/InputSwitch.scss
julianguyen: I can switch it, now that you point it out I think you're correct. https://www.w3schools.com/howto/howto_css_switch.asp also handles it the way you mentioned.

client/app/components/Input/InputSwitch.scss
julianguyen: Haha not sure if 0.95em was a typo but this is what it looks like 
image

On the desktop, I made the font 16px and on mobile, it will be 14px.
image

Setting it 14px on desktop was a tad too small for me.

client/app/components/Input/InputSwitch.jsx
nshki: Ahh, I see, thanks for the detailed explanation! I completely overlooked the `?` in the definition, so I didn't consider the possibility of the prop being undefined.

client/app/components/Input/InputSwitch.scss
```diff
@@ -0,0 +1,38 @@
+@import "../../styles/_global.scss";
+
+.switch {
+  @include setFontSize($size-16);
```
nshki: Ha, sorry 0.95em was completely a typo. Definitely meant 0.095em (Figma doc outputs 0.0944444em).

PR #1113, opened 2018-10-08
Rescue from specific errors

hmasila:
# Description 

Rescuing Exception can cause very unexpected behaviour.

StandardError is the parent of most Ruby and Rails errors. If you catch StandardError you’re not introducing the problems of rescuing Exception, but it is not a great idea.
## More Details

Rescuing all application-level errors might cover up unrelated bugs you don’t know about.

The safest approach is to rescue the error (or errors) you are expecting and deal with the consequences of that error inside the rescue block.

In the event of an unexpected error in your application you want to know that a new error has occurred and deal with the consequences of that new error inside its own rescue block.

Being specific with rescue means your code doesn’t accidentally swallow new errors. You avoid subtle hidden errors that lead to unexpected behaviour for your users and bug hunting for you.

## Corresponding Issue

# Screenshots

# Test Coverage

✅ 
🚫


app/helpers/application_helper.rb
```diff
@@ -8,13 +8,11 @@ def html_options
     { class: 'htmlOptions' }
   end
 
-  # rubocop:disable RescueStandardError
   def i18n_set?(key)
     I18n.t key, raise: true
-  rescue
+  rescue StandardError
```
hmasila: What is the error we should be rescuing from here?

app/helpers/medication_refill_helper.rb
```diff
@@ -3,21 +3,16 @@
...
-      CalendarUploader.new(args).upload_event
-    rescue
-      return_to_sign_in
-    else
-      true
-    end
+    args = calendar_uploader_params(medication)
+    CalendarUploader.new(args).upload_event
+    true
+  rescue StandardError
```
hmasila: What error are we expecting here?

app/helpers/application_helper.rb
nshki: Seems like there could be multiple different options here: https://guides.rubyonrails.org/i18n.html#using-different-exception-handlers, but since we're passing a key here, I think it'd be safe to assume we want to handle the `MissingTranslationData` exception.

app/helpers/medication_refill_helper.rb
nshki: After searching around https://github.com/googleapis/google-api-ruby-client/, which is the gem we're using within `CalendarUploader`, it seems that there are two possible types of exceptions thrown: `Google::Apis::ServerError` and `Google::Apis::ClientError`.

Not entirely certain which is expected to be thrown here, so would love @julianguyen to weigh in.

julianguyen: Great call out! Hmmmmm I'm not sure whether this would be a ServerError (server error) or ClientError (request invalid). It's technically possible it could be either depending on the situation. Maybe ask the folks in that repo?

https://github.com/googleapis/google-api-ruby-client/search?q=ClientError+ServerError&unscoped_q=ClientError+ServerError

hmasila: We could add both errors to cater for the two specific occasions. From my understanding, 

`Google::Apis::ServerError` 
> An error occurred on the server and the request can be retried
In short, they messed up and we can retry

`Google::Apis::ClientError`
> The request is invalid and should not be retried without modification
In short, we messed up and need to upload valid date

nshki: Handling both sounds good to me. 😄

hmasila: One more question, why are we redirecting to `sign_in` if the upload failed? Does it mean the user is not logged in?

app/helpers/date_time_helper.rb
```diff
@@ -4,7 +4,7 @@ module DateTimeHelper
   def format_date(date_str)
     begin
       date_formatted = date_str.to_date
-    rescue StandardError
+    rescue ArgumentError
```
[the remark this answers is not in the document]
nshki: I think handling `ArgumentError` as well as `NoMethodError` here would be good, in case someone passes the wrong number of arguments or passes in an argument that doesn't support `.to_date`.

app/helpers/medication_refill_helper.rb
nshki: That's a great question. I would assume this helper is only used where users are definitely signed in already. Will defer to @julianguyen.

app/models/user.rb
```diff
@@ -141,14 +143,13 @@ def set_defaults
...
-  OAUTH_TOKEN_URL = 'https://accounts.google.com/o/oauth2/token'
-
   def update_access_token
     refresh_token_params = { 'refresh_token' => refresh_token,
                              'client_id'     => nil,
                              'client_secret' => nil,
                              'grant_type'    => 'refresh_token' }
-    response = Net::HTTP.post_form(User::OAUTH_TOKEN_URL, refresh_token_params)
+    uri = URI.parse(User::OAUTH_TOKEN_URL)
+    response = Net::HTTP.post_form(uri, refresh_token_params)
```
hmasila: This method never succeeds. We might not have seen the errors before because of the rescue from Standard error. I am getting 

```{"error"=>"invalid_request",
 "error_description"=>"Could not determine client ID from request."}```

Possibly because we don't have `client_id` and `client_secret`. @julianguyen do we have these credentials for testing?

File: app/models/user.rb [hunk 5]
```diff
@@ -141,14 +143,13 @@ def set_defaults
...
 
-  OAUTH_TOKEN_URL = 'https://accounts.google.com/o/oauth2/token'
-
   def update_access_token
     refresh_token_params = { 'refresh_token' => refresh_token,
                              'client_id'     => nil,
                              'client_secret' => nil,
                              'grant_type'    => 'refresh_token' }
-    response = Net::HTTP.post_form(User::OAUTH_TOKEN_URL, refresh_token_params)
+    uri = URI.parse(User::OAUTH_TOKEN_URL)
```
hmasila: The URL has to be converted to URI before passing it through Net::HTTP

app/models/user.rb
julianguyen: Good find :O I'll set up values for GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET in our CircleCI environment variables.

julianguyen: Just set it up, try running again 🤞

julianguyen:
Looks like the issues with missing API keys has been fixed, we're not getting failing tests related to that 😅

app/controllers/categories_controller.rb
```diff
@@ -90,13 +90,11 @@ def quick_create
   private
 
   # Use callbacks to share common setup or constraints between actions.
-  # rubocop:disable RescueStandardError
   def set_category
     @category = Category.friendly.find(params[:id])
-  rescue
+  rescue ActiveRecord::RecordNotFound
```
jm96441n: :clap: :clap:

app/helpers/medication_refill_helper.rb
```diff
@@ -3,21 +3,18 @@
...
-      args = calendar_uploader_params(medication)
-      CalendarUploader.new(args).upload_event
-    rescue
-      return_to_sign_in
-    else
-      true
-    end
+    args = calendar_uploader_params(medication)
+    CalendarUploader.new(args).upload_event
+    true
```
jm96441n: is the `true` here necessary? it seems like a bit of a code smell to be explicitly returning `true`

hmasila: @julianguyen ☝️ I share the same sentiments here. What do you think?

app/models/user.rb
hmasila: So, can I change 

```
'client_id'     => ENV['GOOGLE_CLIENT_ID'],
'client_secret' => ENV['GOOGLE_CLIENT_SECRET']
```

app/models/user.rb
```diff
@@ -141,14 +143,13 @@ def set_defaults
...
-                             'client_id'     => nil,
-                             'client_secret' => nil,
-                             'grant_type'    => 'refresh_token' }
-    response = Net::HTTP.post_form(User::OAUTH_TOKEN_URL, refresh_token_params)
+    params = { 'refresh_token' => refresh_token,
+               'client_id'     => ENV['GOOGLE_CLIENT_ID'],
+               'client_secret' => ENV['GOOGLE_CLIENT_SECRET'],
+               'grant_type'    => 'refresh_token' }
+
+    response = Net::HTTP.post_form(URI.parse(OAUTH_TOKEN_URL), params)
```
hmasila: Actually, I can just stub this method. But i'll let the environment variables remain

app/helpers/medication_refill_helper.rb
julianguyen: I don't think so, but if you look at medications_controller.rb, you'll see the use of `return unless save_refill_to_google_calendar(@medication)` in the create and update methods. So I think simply calling save_refill_to_google_calendar without returning works.

app/models/user.rb
hmasila: there are a couple of errors

1. `{"error"=>"invalid_request", "error_description"=>"Missing required parameter: refresh_token"}` - we don't validate the presence of `refresh_token` so this can be possible

2. `{"error"=>"invalid_request", "error_description"=>"Could not determine client ID from request."}` - if I pass in some refresh token, then this is the next error and it goes without saying the next one is for the secret

3. If I put an unrecognizable client id or secret, I get `{"error"=>"invalid_client", "error_description"=>"The OAuth client was not found."}`

4. If we ignore the above and go to L155 `new_expiration_time = Time.zone.now + decoded_response['expires_in']` then we get `NoMethodError: undefined method 'round' for nil:NilClass` because `decoded_response['expires_in']` is nil

app/helpers/medication_refill_helper.rb
```diff
@@ -3,21 +3,16 @@
...
-      return_to_sign_in
-    else
-      true
-    end
+    args = calendar_uploader_params(medication)
+    CalendarUploader.new(args).upload_event
+  rescue Google::Apis::ClientError
+    return_to_sign_in
+  rescue Google::Apis::ServerError
+    redirect_to_medication(medication)
```
hmasila: I removed the recursive call here because we don't know how long the server will be down. And because this is not a background job, there is no need to halt everything else and wait for it to pass

app/models/user.rb
julianguyen: Are the errors you mentioned no longer appearing anymore?

spec/helpers/medication_refill_helper_spec.rb
```diff
@@ -22,11 +26,60 @@
...
+        before do
+          allow_any_instance_of(CalendarUploader).to receive(:upload_event).and_return(true)
+        end
+
+        it { expect(helper.save_refill_to_google_calendar(medication)).to eq(true) }
+      end
     end
 
-    context 'when the user has not google oauth2 enabled and/or they no need a new refill reminder' do
+    context 'when the user has not google oauth2 enabled and/or they dont need a new refill reminder' do
```
julianguyen: Spellcheck: replace with don't

spec/models/user_spec.rb
```diff
@@ -136,7 +136,7 @@
   end
 
   describe '#available_groups' do
-    it "returns the groups that allys belong to and the user doesn't" do
+    it 'returns the groups that allys belong to and the user doesnt' do
```
julianguyen: Why was the apostrophe removed?

spec/models/user_spec.rb
```diff
@@ -150,4 +150,53 @@
...
+
+      it 'returns a new access token' do
+        expect(user.update_access_token).to eq('MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI3')
+      end
+    end
+
+    context 'when request is unsuccessful' do
+      before do
+        response = {
+          'error': 'inavalid request',
```
julianguyen: Spellcheck: replace with invalid

app/models/user.rb
hmasila: I believe they are still there. I stubbed the request so tests might not show it

spec/models/user_spec.rb
hmasila: I replaced double strings with single for consistency but can revert if necessary

spec/models/user_spec.rb
hmasila: This is the response we get from google oauth. I think we might want to be consistent with their error response

julianguyen: LOL really, wow. That makes sense to be consistent, but they should really fix that. You could probably file a PR if that code is open source.

julianguyen:
Is this ready to merge? :)

hmasila:
@julianguyen Yes it should be. I talked to @nshki to give a final review. Seems they haven't gotten the time yet

spec/models/user_spec.rb
```diff
@@ -5,8 +5,8 @@
 # Table name: users
 #
 #  id                     :integer          not null, primary key
-#  email                  :string           default(""), not null
-#  encrypted_password     :string           default(""), not null
+#  email                  :string           default('), not null
```
nshki: This is super minor, but is this supposed to be `''`?

spec/models/user_spec.rb
```diff
@@ -5,8 +5,8 @@
...
+#  encrypted_password     :string           default('), not null
```
nshki: Same thing here, is this supposed to be `''`?

spec/models/user_spec.rb
```diff
@@ -136,7 +136,7 @@
   end
 
   describe '#available_groups' do
-    it "returns the groups that allys belong to and the user doesn't" do
+    it "returns the groups that allys belong to and the user does'nt" do
```
nshki: Minor typo. Should be `doesn't`.

spec/models/user_spec.rb
hmasila: lol, I now get what you meant. Damn it!!  Okay `invalid` it is 🤣

PR #1114, opened 2018-10-08
Refactor Omniauth Callbacks Controller

hmasila:
# Description 
A minor change that will make the `google_oauth2` method shorter.
Move the user definition to it's own method and refer to it from `google_oauth2`

## Corresponding Issue

# Screenshots

# Test Coverage

✅ 
🚫


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,20 +1,20 @@
...
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
-  # rubocop:disable MethodLength
   def google_oauth2
-    @user = User.find_for_google_oauth2(
-      request.env['omniauth.auth'],
-      current_user
-    )
-
-    if @user
+    if user
```
jm96441n: Can we use `user.present?` or something like that to make it a bit more explicit what this is doing

spec/support/stub_google_oauth2.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module StubGoogleOauth2
+  def stub_env_for_omniauth
+    request.env["devise.mapping"] = Devise.mappings[:user]
```
julianguyen: Nitpick: Let's use single quotes and instead of using arrow notation for the hash object, let's use colon notation.

PR #1117, opened 2018-10-09
Add tests for meeting controller

hmasila:
# Description 

## Corresponding Issue

# Screenshots

# Test Coverage

✅ 
🚫


jm96441n:
For the specs, can we used the `build` function that Factory Bot provides to speed these up a bit? Since we're not testing ActiveRecord's ability to persist we should focus on ensuring that our code hits the proper AR methods (something like `expect(Class).to receive(:create).with(args)`) so that we can isolate the functionality we're testing

spec/controllers/meetings_controller_spec.rb
```diff
@@ -248,4 +248,134 @@
       end
     end
   end
+
+  describe 'POST #create' do
+    let!(:user) { create(:user) }
+    let!(:group) { create(:group) }
```
jm96441n: Do these need to be persisted? Can we use mocks and stub the DB calls instead?

spec/controllers/meetings_controller_spec.rb
```diff
@@ -248,4 +248,134 @@
...
+    context 'when the user is logged in' do
+      include_context :logged_in_user
+      
+      context 'user is the group leader' do
+        it 'deletes the meeting members' do
+          expect { delete :destroy, params: { id: meeting.id } }
+            .to change(MeetingMember, :count).by(-1)
+        end
+
+        it 'deletes the meeting' do
```
hmasila: I have tried 

```ruby
it 'deletes the meeting' do
    expect(meeting).to receive(:destroy)
    delete :destroy, params: { id: meeting.id }
end
```
and 
```ruby
it 'deletes the meeting' do
    expect(meeting).to receive(:destroy)
    delete :destroy, params: { id: meeting.id }
end
``` 
but all fail with 
```shell
expected: 1 time with any arguments
received: 0 times with any arguments
```

Where I'm I going wrong? @jm96441n

hmasila: @jm96441n if you get time please respond to this. Thanks

jm96441n: Let me pull down your branch and give it a try, that seems like it should be working but there might be a difference in the instance of the meeting being retrieved in the controller and the instance being stubbed here. This is an instance where mocks/stubs might make this a bit easier by using a mock in the setup (`build :meeting`) and stubbing the find method on the `Meeting` class you can ensure that you're working with the same object in the test and the controller. I'll post back later tonight when I have a chance to check this out though

jm96441n: Okay, so the issue comes from the fact that `@meeting` in the controller and `meeting` in the spec refer to the same database row, but are two different instantiated ruby objects. The fix would be to stub out the find method on `Meeting` so that it returns the `meeting` object instantiated in the spec (example is the spec on line 356):
```
it 'deletes the meeting' do
  # Stub the 'set_meeting' method to return the object we instantiate in our spec
  allow(Meeting).to receive_message_chain(:friendly, :find).and_return(meeting)
  expect(Meeting).to receive(:destroy)
  delete :destroy, params: { id: meeting.id }
end
```

jm96441n: Doing this also allows us to use 
```
let(:meeting) { build(:meeting) }
```
instead of create so we can remove the DB hits in our unit tests and get closer to isolated unit tests where we're testing only the logic we write

hmasila: Great! Thanks @jm96441n

Issue #1116, opened 2018-10-08
Convert toggle_locale.js to React


julianguyen:
# Description 

Convert toggle_locale.js to React

# Do you want to be the assignee to work on this?

✅

jm96441n:
I could possibly take this on, let me look a little more into it. Is `toggle_locale.js` a js library or something that was built?

julianguyen:
@jm96441n I'm actually almost done (just have to write tests), thanks for offering though :) I'll add you as a reviewer when I'm done.

jm96441n:
sounds good to me!

jm96441n:
Oops, my bad, totally missed that you were assigned to this!

julianguyen:
@jm96441n Would you be interested in https://github.com/ifmeorg/ifme/issues/1118? :D

jm96441n:
I think I can handle that!

PR #1119, opened 2018-10-09
 [#1116] Convert toggle_locale.js to React

julianguyen:
# Description 

- Convert toggle_locale.js to React, refactoring all over the place
- Remove contributors listed twice between blurbs.json and contributors.json
- Change default colour scheme for Input component

Once merged, I'll update the developer config steps for adding a new language https://github.com/ifmeorg/ifme/wiki/Translations. It's simpler now, you just have to update `config.i18n.available_locales array` in `config/application.rb`.

## Corresponding Issue

#1116

# Screenshots

capture

image

# Test Coverage

✅


app/controllers/meetings_controller.rb
```diff
@@ -69,6 +68,7 @@ def create
       render :new
     end
   end
+  # rubocop:enable MethodLength
```
julianguyen: Rubocop complained that the disable comments weren't closed with enable ones.

app/controllers/pages_controller.rb
```diff
@@ -91,19 +83,14 @@ def fetch_posts
...
+        link_name: post[1]['title'],
+        link: "https://medium.com/ifme/#{post[1]['uniqueSlug']}",
+        author: parse_author(post)
       )
     end
     posts
   end
 
-  def set_posts
-    non_medium_posts = JSON.parse(File.read('doc/pages/posts.json'))
```
julianguyen: Removed this since we link directly to our Medium blog

client/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "build-i18n": "rm app/libs/i18n/translations.js app/libs/i18n/default.js; rake react_on_rails:locale",
-    "build:test": "rm -rf public/webpack-test && NODE_ENV=test webpack --config webpack.config.js --display-error-details",
+    "build:test": "rm -rf ../public/webpack-test && NODE_ENV=test webpack --config webpack.config.js --display-error-details",
```
julianguyen: Whoops wrong path lol

config/locales/en.yml
```diff
@@ -223,7 +223,7 @@ en:
   index:
    title: 'Medications'
    subtitle: 'Keep track of your medications, including dosage and refill information.'
-   total: 'Total:'
+   total: 'Total'
```
jm96441n: I don't think the quotes here are necessary, AFAIK in locale files quotes are only necessary for interpolation

spec/controllers/pages_controller_spec.rb
```diff
@@ -94,28 +94,30 @@
     end
   end
 
-  describe 'GET #toggle_locale' do
+  describe 'POST #toggle_locale' do
     context 'When user is signed in' do
       let(:user) { create(:user) }
```
jm96441n: I know this was here previously, but do we need to seed a user here? Can me mock it out using `build` as opposed to `create` so we're not hitting the db?

client/app/components/Input/InputTextarea.scss
```diff
@@ -34,4 +34,28 @@
...
+      }
+    }
+  }
+  .content {
+    color: $white;
+    a {
+      color: $cornflower !important;
+    }
+    b, strong {
+      color: $white !important;
```
nshki: Any way we can avoid using `!important` here and a few other spots in this file? Would prioritize heavier specificity over using these, but no worries if it's necessary!

julianguyen: Good call out and I agree with you. Unfortunately, I have to override this with `!important` because of Pell Editor womp womp.

nshki: Pell Editorrrrrrrr womp.

julianguyen:
@nshki Great comment about the select highlight area! So the best I could do is get the highlight to match the entire height of the container because select is annoying and you can can't customize padding. We still want to to have left and right padding, so this is the best I could do.

I also explored a solution using JS, but it turns out you can't programatically trigger a select menu to open using JS. https://medium.com/browserquirks/browserquirk-programmatically-opening-a-select-box-4ca745a8468f

image

julianguyen:
@nshki Got it working 🎉 
image

app/assets/images/caret_down.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" data-prefix="fas" data-icon="caret-down" class="svg-inline--fa fa-caret-down fa-w-10" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512"><path fill="white" d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"></path></svg>
```
julianguyen: Directly downloaded svg from Font Awesome, I can't find a way to import SVGs to SCSS files

nshki: Oh interesting. Does referencing the SVG path in `url()` not do anything?

julianguyen: There's no way to get the path for the SVG at compile time that I found :/

nshki: Oohhh I see. Well. Not the end of the world. It might actually be better to keep the SVG directly in the markup in case we want to apply transitions to it in the future.

julianguyen: Good point! I'll poke around try that instead.

julianguyen: @nshki Updated!

Issue #507, opened 2017-07-21
Can allies sign in with Google?


shelleyvadams:
When my family member followed the link in her ally invitation email to accept the invitation, she didn't see any way to sign up with her Google account, just a form to create a password.

Is it possible for an ally to sign in with Google, or only with a password?

julianguyen:
Excellent question. Didn't realize this bug, but yeah, you can only sign up via password when you are given an allies invitation. Once you create an account with a password, and that account is a google account, you should be able to log in via Google auth. But we should allow users to join right away with Google auth.

jnamster:
![](https://github.trello.services/images/mini-trello-icon.png) [(3) Add Google auth support to accepting ally invites](https://trello.com/c/GczM7BHF/60-3-add-google-auth-support-to-accepting-ally-invites)

SophieRoseMcDonald:
![](https://github.trello.services/images/mini-trello-icon.png) [(3) Add Google auth support to accepting ally invites](https://trello.com/c/GczM7BHF/60-3-add-google-auth-support-to-accepting-ally-invites)

alexiamcdonald:
Let me know if you need any help. @abrophy and I are working on other Google related issues.

julianguyen:
@alexiamcdonald No one is working on this at the moment, so you could take it on :D

alexiamcdonald:
Ok I'll take a look tomorrow. Maybe I can knock two birds out with one stone. 😄

PR #1122, opened 2018-10-09
#507 Accept invites through google oauth.

dipil-saud:
# Description 

Accept invitations with google oauth.

## #507 

# Screenshots
image

# Test Coverage

✅

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -3,6 +3,8 @@
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def google_oauth2
     if user.present?
+      user.accept_invitation! if invitation_token
+
       flash[:notice] = I18n.t('devise.omniauth_callbacks.success',
                               kind: t('omniauth.google'))
       sign_in_and_redirect @user, event: :authentication
@@ -11,10 +13,17 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     end
   end
 
+  private
+
   def user
     @user ||= User.find_for_google_oauth2(
       request.env['omniauth.auth'],
       current_user
     )
+  end
+
+  # returns the invitation token passed in with the callback url
+  def invitation_token
+    request.env.dig('omniauth.params', 'invitation_token')
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -98,9 +98,9 @@ class User < ApplicationRecord
   def self.find_for_google_oauth2(access_token, _signed_in_resource = nil)
     data = access_token.info
     user = find_or_initialize_by(email: data.email) do |u|
-      u.name = data.name
       u.password = Devise.friendly_token[0, 20]
     end
+    user.name ||= data.name
 
     user.update!(
       provider: access_token.provider,
--- a/app/views/users/invitations/edit.html.erb
+++ b/app/views/users/invitations/edit.html.erb
@@ -21,3 +21,11 @@
 
   <%= f.submit t("devise.invitations.edit.submit_button"), class: 'buttonGhostM smallMarginTop' %>
 <% end %>
+<div class="center">
+  <div class="smallMarginTop">
+    <%= button_to t('devise.shared.sign_up_google'),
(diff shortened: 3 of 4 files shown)


app/views/users/invitations/edit.html.erb
```diff
@@ -21,3 +21,15 @@
 
   <%= f.submit t("devise.invitations.edit.submit_button"), class: 'buttonGhostM smallMarginTop' %>
 <% end %>
+<div class="center">
+  <div class="smallMarginTop">
+    <div class="smallMarginBottom">
```
julianguyen: I don't think we need the nested div with smallMarginBottom, wouldn't the sign up with Google button be the only button?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -3,6 +3,7 @@
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def google_oauth2
     if user.present?
+      user.accept_invitation! if invitation_token
```
julianguyen: Hmm, what happens if user.accept_invitation doesn't execute successfully? I think we should still sign in the user via Google OAuth but I'm wondering if it's worth displaying some sort of error message telling them the invitation didn't go through 🤔

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -1,49 +1,44 @@
 # frozen_string_literal: true
 
 RSpec.describe OmniauthCallbacksController, type: :controller do
+  shared_examples 'successful sign in with oauth details' do
```
julianguyen: Nice!

app/controllers/omniauth_callbacks_controller.rb
dipil-saud: `accept_invitation!` just sets the `invitation_accepted_at` and clears the token so if the user is valid upto that point, it will not error out. The method does not raise an exception(although the `!` at the end of it is misleading), so the the user sign in will go through.

dipil-saud: There however is one catch here. Since the user can use any google account to login, the email for the google account used can be different from the email used in the invitation. I've not put in any restrictions so the user is able to sign-up even if that's the case.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -11,10 +12,17 @@ def google_oauth2
     end
   end
 
+  private
```
hmasila: 👍

app/models/user.rb
```diff
@@ -98,9 +98,9 @@ def available_groups(order)
   def self.find_for_google_oauth2(access_token, _signed_in_resource = nil)
     data = access_token.info
     user = find_or_initialize_by(email: data.email) do |u|
-      u.name = data.name
       u.password = Devise.friendly_token[0, 20]
```
hmasila: what happens to the previous password in a case where the user exists?

dipil-saud: The block is only called to initialize a new user if the user does not exist.

app/controllers/omniauth_callbacks_controller.rb
julianguyen: I think that's fine, it's possible that the invitation was originally sent to a non-Google account and the person wants to use their Google account instead. Thanks for explaining. 

Yeah the `!` is misleading :/ What happens if you exclude it?

dipil-saud: That's just how the method is named in `devise_invitable`. Excluding the `!` calls the method which sets the `invitation_accepted_at` but doesn't save the model.

julianguyen: Ah okay, let's keep what we have here then! Thanks for clarifying!

app/models/user.rb
julianguyen: Actually, I know I just approved the PR, but can you confirm why you moved the logic for setting u.name outside of the initialize block?

dipil-saud: A user is created with the email but without a name when an invitation is sent, so the name needs to be set for that user. It won't update if a name is present, so existing functionality is preserved.

Issue #971, opened 2018-09-18
Cleanup code duplication warnings identified by Codeclimate


julianguyen:
# Description 

Cleanup code duplication warnings identified by Codeclimate

https://codeclimate.com/github/ifmeorg/ifme/issues?category=duplication&engine_name%5B%5D=structure&engine_name%5B%5D=duplication

# Do you want to be the assignee to work on this?

🚫

DRBragg:
I love me some refactoring and I'm happy to help. Just sent you an email request to join.

DRBragg:
Removing myself from this for now, most of the CC issues are for React and the RoR issues is just the common_fields method being repetitive. I'd rather leave this open for another beginner who is more comfortable with JS than Ruby so they can get a PR. I can circle back around later.

julianguyen:
@DRBragg That's very thoughtful of you to do that. There are other more advanced RoR and React issues if you're interested!

DanielSauve:
Hey there! I would gladly pick this up, I sent an email request to join!

PR #1126, opened 2018-10-09
Issue 971 duplication

DanielSauve:
# Description 

Cleaned up the code duplication identified by code climate, mainly wrote functions to generate objects that are used in the UI since many of the UI fields are similar and popped up as duplicated code.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/971

# Test Coverage

✅


app/helpers/medications_helper.rb
```diff
@@ -229,6 +143,50 @@ def days_checkbox
...
-end
 
+  # rubocop:disable MethodLength, LineLength
+  def medication_unit_field(type)
+    {
+      id: "medication_#{type}_unit",
+      type: 'select',
+      name: "medication[#{type}_unit]",
+      dark: true,
+      value: @medication["#{type}_unit"] || t('medications.units.tablets.other'),
```
julianguyen: I would just put `t('medications.units.tablets.other'),` on the next line and remove the disable LineLength comment.

app/helpers/viewers_helper.rb
```diff
@@ -49,4 +49,32 @@ def get_viewers(data, data_type, obj)
     end
     []
   end
+
+  # rubocop:disable MethodLength
+  def get_viewers_input(viewers, name, translation_name, obj)
+    input = {}
+    if viewers.present?
+      checkboxes = []
```
julianguyen: We should move the logic for populating checkboxes into another helper.

config/locales/de.yml
```diff
@@ -211,6 +211,7 @@ de:
    refill_hint: 'Wann dieses Rezept erneuert werden muss'
    refill_reminder_hint: "Wir senden dir eine Erinnerungsmail in der Woche bevor dieses Arzneimittel aufgefüllt werden muss."
    daily_reminder_hint: "Wir senden dir eine tägliche Erinnerungsmail zum Einnehmen dieses Arzneimittels"
+   total: 'Gesamt'
```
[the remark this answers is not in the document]
DanielSauve: The main reason I added the new translation was due to the former translation having a ":" at the end, while every other title field on the form doesn't contain a ":"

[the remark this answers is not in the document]
julianguyen: Ah, I see! In a recent commit that was merged to master, we actually remove the colon from the total. Can you try rebasing and using that existing value instead?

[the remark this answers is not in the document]
julianguyen: Oh it looks like `total: 'Gesamt:'` still exists
You can just remove the colon from here!

app/helpers/moments_helper.rb
```diff
@@ -90,7 +90,7 @@ def moment_form_inputs
         checkboxes: checkboxes_for(@strategies),
         formProps: quick_create_strategy_props
       },
-      moments_viewers_input,
+      get_viewers_input(@viewers, 'moment', 'moments', @moment),
```
jm96441n: Are these ivars coming from the controller? If so can we pass the ivars in from the view file as arguments to the method that calls this? [This blog post](http://www.carlosramireziii.com/stop-using-instance-variables-in-partials.html) references using ivars in partials, but I feel the same logic applies to helpers as well

julianguyen:
Sorry for the merge conflicts with the translation files, I recently reformatted them. Let me know if you run into any problems, happy to help!

DanielSauve:
I'm going to look at cleaning up all the things addressed by the feedback, should be done by the end of tonight!

DanielSauve:
Sorry about this taking a little bit longer than expected, I haven't had access to my Mac the past couple days, and you were right about the Windows setup being difficult 😛

DanielSauve:
I'll address the last two comments when I get a chance to use my Mac again 😅

app/helpers/viewers_helper.rb
DanielSauve: How would you recommend I go about doing this?

julianguyen: I would make a helper method that returns the value of checkboxes and it would contain lines 58-65.

DanielSauve:
I'm not sure why merging the changes from the upstream master into my PR caused some of the tests to fail, any ideas on what went wrong?

julianguyen:
@DanielSauve This might be an issue that's only happening in the Circle build. I'm also encountering this in https://github.com/ifmeorg/ifme/pull/1177/files. Looking into it now. When you run the tests locally do they pass for you?

julianguyen:
@DanielSauve I fixed the issue, can you rebase from master? :)

DanielSauve:
Perfect! That seems to have fixed it!

Issue #968, opened 2018-09-18
Improve code quality and readability in /app/views/shared/_stats.html.erb


julianguyen:
# Description 

When working on implementing the redesigns I noticed that `/app/views/shared/_stats.html.erb` was pretty messy to look at (did not have time to clean it up myself). There's a lot of logic in this view when it should be moved to a helper and possibly subviews.

imi56:
I would love to work on this.

julianguyen:
Hey @aSquare14, if you don't mind I'm going to assign you and @prateksha from this issue just because you two have multiple open PRs that still need to be wrapped up. Happy to sync with you after if you want to work on a refactoring related issue.

julianguyen:
Hey @imi56, thanks for expressing your interest. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :) After that, assign yourself to the issue.

OPARA-PROSPER:
@julianguyen Can i work on this issue?

julianguyen:
@OPARA-PROSPER 	Of course! Go ahead and assign yourself to the issue :)

julianguyen:
@OPARA-PROSPER Good question, so that file is quite lengthy for a view file. We could start off by moving lines 20-42 to another view partial.

There's some conditional logic that is repeated throughout the file, so you could move it into a helper for e.g. `!local_assigns[:profile] || local_assigns[:profile] == current_user.id`

julianguyen:
I would also look for opportunities to refactor code e.g.
- Replace uses of where with find
- Replace `local_assigns[:data_type] == 'category' || local_assigns[:data_type] == 'mood' || local_assigns[:data_type] == 'strategy'` with `%w(category mood strategy).include? local_assigns[:data_type]`

OPARA-PROSPER:
@julianguyen Thanks for the feedback, i really appreciate them!

Can you give me a little bit of time to wrap my head around this concepts, i'd need to dig deep because i'd love to understand what i'm doing better, so i can learn better.

if i eventually venture into ruby (Which probably looks so), that will be great for me. <3

julianguyen:
@ScotianMan ^ any ideas why this bot keeps posting? 🙃

julianguyen:
@OPARA-PROSPER Of course, take your time and feel free to ask questions on Slack too.

Here are some articles I found on refactoring Rails views which may be helpful learning material:
[...]

PR #1142, opened 2018-10-11
/app/views/shared/_stats.html.erb refactor

OPARA-PROSPER:
# Description 

I refactored the _stats.html.erb partial view like @julianguyen suggested on issue #968
## More Details

Refactoring this partial view involved:
- Creating another partial view in the _shared folder_ i named the view __focus.html.erb__ 
- Replacing local_assigns[:data_type] == 'category' || local_assigns[:data_type] == 'mood' || local_assigns[:data_type] == 'strategy' with %w(category mood strategy).include? local_assigns[:data_type] 
## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/968
# Screenshots

# Test Coverage

✅ 
🚫


app/views/shared/_stats.html.erb
```diff
@@ -17,29 +17,7 @@
...
-              <% elsif local_assigns[:data_type] == 'strategy' && (!local_assigns[:profile] || local_assigns[:profile] == current_user.id) %>
-                <%= viewers_hover(get_viewers_for(Strategy.where(id: key).first, 'strategy'), Strategy.where(id: key).first) %>
-              <% elsif local_assigns[:data_type] == 'strategy' && local_assigns[:profile] %>
-                <div class="someFocusTextName">
-                  <%= Strategy.where(id: key).first.name %>
-                </div>
-              <% end %>
-            </div>
-          </div>
+          <%= render 'shared/focus' %>
```
julianguyen: Does the page render properly out of curiosity?
In our codebase, we've been doing `<%= render partial: '/shared/stats', locals: { data_type: 'category' } %>` for example.

app/views/shared/_focus.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="someFocus">
+    <div class="someFocusText">
+      <% if local_assigns[:data_type] == 'category' && (!local_assigns[:profile] || local_assigns[:profile] =current_user.id) %>
```
julianguyen: `local_assigns[:profile] =current_user.id` should be `local_assigns[:profile] == current_user.id`

Looks like this an issue in multiple places.

OPARA-PROSPER: Yah, just noticed.. I fixed this

app/views/shared/_stats.html.erb
julianguyen: You should really be running the app locally on your machine while making changes to it. Unfortunately, we don't have automated tests for everything. Things like UI often times have to be tested manually.

Are you using Windows or Linux? I recommend using Linux and following our [Installation](https://github.com/ifmeorg/ifme/wiki/Installation) steps for that.

OPARA-PROSPER: I'm using a windows, still working to get a dual boot (Really need to keep using windows for now), Once i get it i'd be sure to install ruby on rails

julianguyen: Sounds good! I also recommend using Linux in VirtualBox. It takes a lot less time than doing a dual boot installation.

Setting up Ruby on Rails on Windows is possible but not as straightforward compared to Linux or macOS.

app/views/shared/_focus.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="someFocus">
+    <div class="someFocusText">
+      <% if local_assigns[:data_type] == 'category' && (!local_assigns[:profile] || local_assigns[:profile] ==current_user.id) %>
+        <%= viewers_hover(get_viewers_for(Category.where(id: key).first, 'category'), Category.where(id: ke.first) %>
```
jm96441n: Can we move these AR queries into the helper method?

jm96441n:
Can we move the AR queries that are present in the view into the controller? Running AR queries in the view breaks the MVC pattern (where the View is only responsible for displaying objects queried from the controller)

julianguyen:
@OPARA-PROSPER Hey, I saw that you were setting up your dev environment! Any luck and are you still working on this issue?

Issue #1144, opened 2018-10-12
Flakey test in strategies_controller_spec.rb related to `ActiveRecord::RecordNotUnique`


julianguyen:
# Description 

I notice this is a frequent flakey test that fails on CircleCI. Sometimes I see it happen randomly when I run rspec in my dev environment. Maybe there's a race condition happening? O.o

```
spec/controllers/strategies_controller_spec.rb
Failure/Error: another_user = create(:user)

ActiveRecord::RecordNotUnique:
  PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "users_pkey"
  DETAIL:  Key (id)=(1) already exists.
  : INSERT INTO "users" ("email", "encrypted_password", "created_at", "updated_at", "name", "uid") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"
./spec/controllers/strategies_controller_spec.rb:298:in `block (5 levels) in <top (required)>'
------------------
--- Caused by: ---
PG::UniqueViolation:
  ERROR:  duplicate key value violates unique constraint "users_pkey"
  DETAIL:  Key (id)=(1) already exists.
  ./spec/controllers/strategies_controller_spec.rb:298:in `block (5 levels) in <top (required)>'
```

# Do you want to be the assignee to work on this?

🚫

jaimevelaz:
I can take a look at this one 😁

PR #1145, opened 2018-10-12
[#1144] Fix flakey test in strategies_controller_spec

jaimevelaz:
# Description 

`StrategiesControllerSpec` had a flaky spec caused by hardcoding active records ids on fixtures, to avoid this kind of issues I replaced all hardcoded IDs and let the database assign them.

## More Details

Checking [CircleCI history builds](https://circleci.com/gh/ifmeorg/ifme/4565) I found that the seed that made the spec fail is `54890` which now works fine. The flaky spec is reproducible using that seed `rspec --seed 54890 spec/controllers/strategies_controller_spec.rb` in `master`

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1144

# Test Coverage

Same

### DIFF
--- a/spec/controllers/strategies_controller_spec.rb
+++ b/spec/controllers/strategies_controller_spec.rb
@@ -62,7 +62,7 @@ describe StrategiesController do
 
         it 'renders no content for a json request' do
           get :show, format: 'json', params: { id: id }
-          expect(response.body).to eq('')
+          expect(response.body).to be_empty
         end
       end
     end
@@ -90,7 +90,7 @@ describe StrategiesController do
       context 'when the comment is saved' do
         it 'responds with an OK status' do
           post :comment, params: valid_comment_params
-          expect(response.status).to eq(200)
+          expect(response).to be_successful
         end
       end
 
@@ -130,7 +130,7 @@ describe StrategiesController do
 
           it 'renders nothing' do
             get :delete_comment, params: { commentid: comment.id }
-            expect(response.body).to eq('')
+            expect(response.body).to be_empty
           end
         end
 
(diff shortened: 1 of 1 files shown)


spec/controllers/strategies_controller_spec.rb
```diff
@@ -263,12 +257,6 @@
           post :create, params: strategy_params
           expect(response).to redirect_to(strategy_path(assigns(:strategy)))
         end
-
-        it 'redirects to the strategy show' do
```
jaimevelaz: This spec was almost the same as line 262, take the opportunity to remove it.

spec/controllers/strategies_controller_spec.rb
```diff
@@ -90,7 +90,7 @@
       context 'when the comment is saved' do
         it 'responds with an OK status' do
           post :comment, params: valid_comment_params
-          expect(response.status).to eq(200)
+          expect(response).to be_successful
```
hmasila: I think we should go with the specific status here. Either 204 or 200

spec/controllers/strategies_controller_spec.rb
```diff
@@ -209,42 +204,41 @@
   end
 
   describe 'GET edit' do
-    let(:user)        { create(:user, id: 1) }
-    let!(:strategy1)  { create(:strategy, user_id: 1, id: 1) }
-    let!(:strategy2)  { create(:strategy, user_id: 2, id: 2) }
+    let(:another_user) { create(:user) }
+    let!(:strategy1)   { create(:strategy, user: user) }
```
hmasila: 👍

spec/controllers/strategies_controller_spec.rb
jaimevelaz: I see your point, changed it back to check the expected status code. Thanks for the feedback

PR #1146, opened 2018-10-12
Fix flakey group spec

jaimevelaz:
# Description 

After looking `CircleCi` build to gather information to fix https://github.com/ifmeorg/ifme/pull/1145 I found that `spec/models/group_spec.rb` also had a flakey spec that failed due the same reason (hardcoded IDs and specs order). 

I think we should create an issue to review hardcoded IDs in fixtures to avoid flakey specs in the future.

## More Details

I took the opportunity to normalize the boolean assertions rspec format, I went for `be_truthy` and `be_falsey` but I can change it back if we think `be false` and `be true` is more expressive.

## Corresponding Issue

No issue but related to https://github.com/ifmeorg/ifme/issues/1144

# Test Coverage

Same

### DIFF
--- a/spec/controllers/groups_controller_spec.rb
+++ b/spec/controllers/groups_controller_spec.rb
@@ -87,7 +87,7 @@ RSpec.describe GroupsController, type: :controller do
       put :update, params: { id: group.id, group: { leader: [user.id] } }
 
       non_leader.reload
-      expect(non_leader.leader).to be true
+      expect(non_leader.leader).to be_truthy
     end
   end
 end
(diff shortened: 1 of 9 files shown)


jaimevelaz:
>I think to be true or to be false should work in cases where the attribute to be checked can only exist as a boolean. eg take_medication_reminder.active? or meeting.leader? Using is falsey or truthy could create a loophole because be_truthy does not actually check for the Boolean value true. It checks to see if the value is not nil and it's not false. In that case, if it exists with any value, then the test will pass, which shouldn't be the case.

Good point, both assertions are not exactly the same thing, to avoid blocking this simple issue with this I've reverted the formatting change done at left only the brittle spec change.

Issue #1053, opened 2018-10-02
Adding missing test coverage for category.rb


julianguyen:
# Description 

Womp womp we're missing test coverage here. Let's go back and add it!

# Do you want to be the assignee to work on this?

🚫

Atul9:
I would like to work on this

julianguyen:
Hey @Atul9, thanks for expressing your interest. We would love your help with this issue. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps :) After that, assign yourself to the issue.

julianguyen:
@felipemrvieira Can you follow the next steps above and then assign yourself to this issue? Thanks.

thovt93:
@julianguyen I would like to work on this

PR #1067, opened 2018-10-03
Add test for category.rb(#1053)

felipemrvieira:
# Description 
Add tests for the category.rb

## More Details

- Remove category factory from spec/factories/factories.rb and create own  factory on spec/factories/category.rb.
- Create category model spec: spec/models/category_spec.rb
- Test attribute validation
- Test User relation

## Corresponding Issue

#1053 

# Test Coverage

✅

PR #1150, opened 2018-10-13
Adding missing test coverage for category.rb

thovt93:
# Description 

Add tests for the category.rb

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1053

# Screenshots

# Test Coverage

✅


spec/models/category_spec.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.describe Category, type: :model do
+  it { is_expected.to validate_presence_of(:user_id) }
```
jm96441n: this should also validate presence of name, as well as test the `self.link` method

Gemfile.lock
```diff
@@ -447,6 +447,8 @@ GEM
       rubyzip (~> 1.2, >= 1.2.2)
     sentry-raven (2.7.4)
       faraday (>= 0.7.6, < 1.0)
+    shoulda-matchers (3.1.2)
```
julianguyen: Why is this necessary? Can't we use existing matchers?

julianguyen:
Closing this PR, will finish up #1067 based on the comments and merge it.

PR #1152, opened 2018-10-13
Add visually hidden label to language select tag

### LINKED ISSUE #1151: Missing label for language dropdown on home page
# Description 

`select` input on home page needs associated label to describe its purpose.

# Reproduction Steps

Perform a DevTools accessibility audit.

# Screenshots

screen shot 2018-10-13 at 8 22 18 am

✅

digilou:
# Description 

- Added label to language select tag in footer
- Created `visually-hidden` class to maintain the design, yet provide a better experience for screen reader users

## Corresponding Issue

Closes #1151


digilou:
Gah! Didn't think to fetch upstream. Re-working this...

app/views/shared/_footer.html.erb
```diff
@@ -27,7 +27,8 @@
 
   <div class="footerExtras">
     <div>
-      <%= select_tag 'locale', options_for_select(@locales.collect{ |l| [l[:name], l[:locale]] }, @locale), 'aria-label': t('language') %>
+      <%= label_tag 'locale', "Choose a Language", class: 'visually-hidden' %>
+      <%= select_tag 'locale', options_for_select(@locales.collect{ |l| [l[:name], l[:locale]] }, @locale) %>
```
jm96441n: while we're making changes here can we change the iterator from `l` to `locale` or something similar for improved readability

digilou: That is a great idea. I'll do that while I'm in the code.

app/assets/stylesheets/application/shared.scss
```diff
@@ -96,3 +96,15 @@
     }
   }
 }
+
+/* visible to screen readers */
+.visually-hidden {
```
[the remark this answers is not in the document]
digilou: Thanks for the instruction! I do like consistency in a codebase.

digilou:
I'm certain I didn't use the React component correctly. I could use some instruction on that, as I'm not sure if `InputLabel` is the correct one I should be using. Also, I'm having trouble testing on my local machine. I get a 500 error when running the Rails server:
"500 Internal Server Error
If you are the administrator of this website, then please read this web application's log file and/or the web server's log file to find out what went wrong."

client/app/widgets/ToggleLocale/index.jsx
```diff
@@ -36,7 +37,12 @@ const onChange = (e: SyntheticEvent<HTMLInputElement>) => {
 export const ToggleLocale = (props: Props) => {
   const { locale, locales } = props;
   return (
-    <div aria-label={I18n.t('language')}>
```
julianguyen: I was looking into this further and there's a simpler way of doing implementing this without having to introduce more CSS to hide elements.

We would want to extend our Input component to add a prop called hiddenLabel: boolean. If it's true then we don't want to render InputLabel. But we do want to put an aria-label on the input/select element itself.

Let me know if you want to discuss this further :)

digilou: The reason why the CSS is that way is to make it hearable by screen readers, but not visible to mar the design. Hiding the label altogether and replacing it with an aria label is a solution, but screen reader support for ARIA is not consistent, sadly. When a label is possible, I shoot for a label.

On the note of this component, I see an error on the live website:
screen shot 2018-10-16 at 6 53 50 am

digilou:
> @digilou When you look at the rails log, what's the specific error message you get?

It appears to stem from me not having a local database setup for this:
`ActiveRecord::NoDatabaseError (FATAL:  database "ifme_development" does not exist`

client/app/widgets/ToggleLocale/index.jsx
digilou: That being said, I will definitely test the `aria-label` solution with a couple of screen readers!

julianguyen: Thanks for explaining! :)

It looks like that issue you're getting on the live site has to do with our cache which I'll look into. Can you try doing a hard refresh? Thanks for the heads up.

julianguyen:
@digilou The database error you're getting means you haven't set up the database yet. Have you set up Postgres and run migrations? Let us know if you need any help in the #setup channel on Slack

https://github.com/ifmeorg/ifme/wiki/Installation#postgresql
https://github.com/ifmeorg/ifme/wiki/Installation#running-the-app

client/app/widgets/ToggleLocale/index.jsx
digilou: Whoa...
image
I think #1162 targets the white text bug. 
I'm seeing this on a Windows 10 OS and Firefox 62.0.3 browser.

digilou: "Combobox. English. Collapsed." is what NVDA, Firefox, Windows 10 says. I almost wonder if putting `aria-label` on the input, rather than the div, would help the chances of "Languages" being read aloud.

digilou:
I have not! But that makes a LOT more sense. 🤦‍♀️ Thanks for the wiki instruction links!

client/app/widgets/ToggleLocale/index.jsx
julianguyen: That makes sense to me! ^

app/models/strategy.rb
```diff
@@ -55,4 +55,8 @@ def published?
   def comments
     Comment.comments_from(self)
   end
+
+  def comments
```
julianguyen: Looks like this is a duplicate of what exists a few lines above. Let's not include this change in this PR.

app/assets/stylesheets/application/shared.scss
```diff
@@ -96,3 +96,15 @@
     }
   }
 }
+
+/* visible to screen readers */
+.visuallyHidden {
```
julianguyen: Let's move this to `client/app/components/Input/Input.scss`

client/app/widgets/ToggleLocale/index.jsx
```diff
@@ -36,7 +37,12 @@ const onChange = (e: SyntheticEvent<HTMLInputElement>) => {
...
+    <div>
+      <InputLabel
```
julianguyen: Instead of using InputLabel here, let's create a prop in the Input component called hideLabel that's either undefined or a boolean. Then based on that in InputLabel which is called by displayLabel(), apply the visuallyHidden class to InputLabel if hideLabel is true. 

By applying this to the generic Input component, we can reuse this same logic for other Input components that have hidden labels.

Issue #1148, opened 2018-10-12
Enforce password strength for new users


julianguyen:
# Description 

We use Devise for authentication, let's setup password strength for new users who sign up. This [post](https://github.com/plataformatec/devise/wiki/How-To:-Set-up-simple-password-complexity-requirements) should be helpful.

For existing users, let's display a modal prompt (using the Modal component) telling them to `We have enforced more security measures for users. Please %{update your password}` where `%{update your password}` has a link to `edit_user_registration_path`. This prompt will only disappear once the password meets requirements.

# Do you want to be the assignee to work on this?

🚫

LMulvey:
https://github.com/dropbox/zxcvbn - slick low-cost library for doing password strength estimations. I've used it a bunch in the past 🤗

JEscalanteGT:
I would like to work in this issue. How does look a strong password? (Min characters, it should contain numbers, etc)

PR #1153, opened 2018-10-13
Add validation password

yasirazgar:
# Description 
Added validation for password strength, checks the requirements during sign up and also prompts the already registered user to update the password.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/1148

# Screenshots
password validation

# Test Coverage

✅


yasirazgar:
@julianguyen, Just added server side validation and please check the UI for error message and let me know if changes are needed.
Working on prompt for existing users.

spec/models/user_spec.rb
```diff
@@ -135,6 +135,30 @@
     end
   end
 
+  describe '#validations' do
+    context 'password' do
+      let(:user) { create :user }
+
+      context 'with valid one' do
+        it 'saves without any errors' do
```
jm96441n: These seem like they should integration tests if there's any persistence happening. You can also use `build :user` instead and check if `user.valid?` to avoid hitting the DB in unit tests

jm96441n:
I believe this should also contain some feature (integration) testing around the UI prompts as wel

config/locales/en.yml
```diff
@@ -613,6 +613,7 @@ en:
       meeting_notify_text: 'Meeting updates'
       google_auth1: 'If you no longer want to be authenticated with Google, please enter a new email and password.'
       google_auth2: 'Since you are authenticated with Google, you do not need to enter your password.<br>If you want to change this, please enter a new email and password.'
+      password_complexity_error: 'Should contain 8 to 128 character, including 1 uppercase, 1 lowercase, 1 digit and 1 special character'
```
julianguyen: This needs to be translated to our other locales.

Let's also use the following copy instead:
`should contain 8 to 128 characters, including 1 uppercase, 1 lowercase, 1 digit, and 1 special character`.

julianguyen:
Part of this issue is also prompting existing users to go into their Account settings and update their password? Will this be done in another pull request?

yasirazgar:
@julianguyen I will do the prompt part in this PR itself, I just want to confirm the UI and english translation.

julianguyen:
Awesome!

So in en.yml and other translation files we would want translations for:
```
update_your_password_prompt: 'We have enforced better security measures for our users. Please %{update_your_password_link}'
update_your_password: 'Update your password'
```

Then the value would be something like: `t('...update_your_password_prompt', update_your_password_link: <%= link_to t('...update_your_password'), edit_user_registration_path %>)`

yasirazgar:
@julianguyen Looks like we can't get the users password (as string) once it is encrypted and saved, while using devise, which is needed for comparing the password strength for the existing user.
There maybe some complex algorithms to decrypt but still its not good to access the users password.

I think we can use following ways to handle the issue,

1. Send an email (once) to all the existing user which tells the new password criteria and asks them to,  
 "update the password, if there existing password doesn't meet the new criteria or 
 ignore this email, if your existing password has already met the criteria" ( probably using a rake task).
2. We can show a small notice on the home page for a month or so, common to all the users asking them to update the password.

Or please let me know if you have any other idea.

julianguyen:
Oooh, a really good point about existing encrypted passwords (which is good we do that haha).

I think the simplest thing to do is to show the prompt the next time they log in, and if the password is changed after that then we don't need to show the prompt again.

yasirazgar:
Ok for this,
1. We may need to add a boolean field to track when the user updates the password and avoid showing the prompt to them, 
2 . Also some user may not update the password if they already have a strong password, and we will be prompting the user unnecessarily and forcing them to do a unnecessary update.
Is that ok.

julianguyen:
I think it would be okay to force someone who already has a strong password to update it again. It never hurts to refresh passwords 😆 

We could also ask them to change their password every x months, that way we don't need to store anything extra in the DB. Thoughts? @mialahmoka @sexybiggetje

julianguyen:
https://www.networkworld.com/article/3104015/security/periodic-password-changes-good-or-bad.html

Hmm it looks like there's pros and cons.

An alternative to saving something in the DB, is just to make it cookies based. It is possible for someone to clear their cookies and see the prompt again, but again, no harm in changing your password to another strong password.

In the future, we could look into 2FA through a [gem](https://github.com/tinfoil/devise-two-factor) like this.

mialahmoka:
First thing I think is necessary to say in case there's any confusion is that passwords should never be encrypted, they should always be hashed and salted with SHA-256. If passwords are being stored in an encryption algorithm, that's something that needs to be changed. Second thing, I apologize if I read this wrong @yazirasgar but if you're saying having a "strong" password as determined by a small set of requirements should mean that they can't change their password, I strongly disagree. There's a lot of reasons for someone to need to change their password. 

I have a blog post that may contain some relevant information. It's mostly aimed at user password management, but it might give some insight. 
https://mokasec.wordpress.com/blog/

julianguyen:
Both @yasirazgar and I used the word encrypted incorrectly it seems. Sorry, I am a security n00b haha. I looked it up and Devise uses [BCrypt](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/encryptor.rb) to hash their passwords.

julianguyen:
@mialahmoka So do you recommend asking a user to change their password once to meet our small and arbitrary set of requirements for a "strong password" or should it be something done periodically, e.g. every 6 months?

sexybiggetje:
As long as the password change policy doesn't affect external users (google).
Instead of storing a boolean if the user changed the password, store a date/time for when the user has last changed it. For the rest I agree this is a good policy. Whatever period is required for password changes that's anyone's guess. And does it take in to account the password reusability? E.g. have I used this password before?

mialahmoka:
@julianguyen in theory it's a decent idea, however the issue I could see is with salts. If the salt is generated upon initial password input, there would be no way to verify that the password has or hasn't been reused.

yasirazgar:
@mialahmoka, Sorry for using the wrong term, yes the passwords are not encrypted there are hashed using BCrypt in devise, and I think there is way to compare the new passwords with the previous hashed_password i will check, we can store the last three or five encrypted password and compare them with new one entered by the user.

yasirazgar:
@julianguyen I think this one getting bit complex, I will leave this issue till it is completed, ie new user part is completed. And this can be merged.
and will create a new issue to handle the existing user, with all the things we discussed and work on it.

yasirazgar:
Created a new ticket, https://github.com/ifmeorg/ifme/issues/1168, please check the description and update it if I am missing anything.

julianguyen:
Thanks for creating the follow-up issue @yasirazgar :) Sounds like a good plan to me.

spec/models/user_spec.rb
```diff
@@ -135,6 +135,38 @@
...
+        it 'saves without any errors even if the password strength is less' do
+          user.password = 'warsdasdf'
+          user.token = 'access token'
+          expect(user.valid?).to be true
+        end
+      end
+
+      context 'with valid one' do
+        it 'saves without any errors' do
+          user.password = 'waspAr$0'
```
julianguyen: Should you also test the case where user.password is blank because of `return if password.blank? || (password =~ PWD_REGEX)`

yasirazgar:
@julianguyen I have added the missing test and rebased this branch to master.
I think this can be merged after review.

spec/models/user_spec.rb
```diff
@@ -139,6 +139,21 @@
     context 'password' do
       let(:user) { build :user }
 
+      context 'when password is blank' do
+        before do
+          user.save
+        end
+        # password is just a attribute in user, it is not a field
+        # so, when password is nil, when it is not set
+        # devise will take care of nil/blank password while registering a user
```
julianguyen: Good idea on the comment!

yasirazgar: Fixed typos in the comment 😬

app/models/user.rb
```diff
@@ -117,11 +117,9 @@ def self.find_for_google_oauth2(access_token, _signed_in_resource = nil)
   # rubocop:enable MethodLength
 
   def google_access_token
-    if !access_expires_at || Time.zone.now > access_expires_at
-      update_access_token
-    else
-      token
-    end
+    return token if access_expires_at && access_expires_at > Time.zone.now
```
jm96441n: this boolean looks a bit weird, why are we doing comparing something to itself?

app/models/user.rb
```diff
@@ -160,6 +158,14 @@ def update_access_token
 
   private
 
+  def password_complexity
+    return if google_oauth2_enabled?
+    return if password.blank? || (password =~ PWD_REGEX)
```
jm96441n: can we extract this boolean into a separate method for readability?

spec/models/user_spec.rb
```diff
@@ -135,6 +135,53 @@
     end
   end
 
+  describe '#validations' do
+    context 'password' do
+      let(:user) { build :user }
+
+      context 'when password is blank' do
+        before do
+          user.save
```
jm96441n: do we need to save the user here? we should be able to validate the user before persisting

app/models/user.rb
hmasila: I think it checks if `access_expires_at` is present and if it is present, it checks that the expiration time has not reached yet. If both are true then we return the token; otherwise, the token needs to be updated

app/models/user.rb
hmasila: @jm96441n I think it is okay the way it is since it is not a complex condition

spec/models/user_spec.rb
```diff
@@ -135,6 +135,53 @@
...
+        end
+        # password is just an attribute in user, it is not a field
+        # so, password is nil, when it is not set
+        # devise will take care of nil/blank password while registering a user
+        it 'should not throw any error' do
+          user = User.first
```
hmasila: I think we can have something in the lines of

```ruby
user = User.new
user.password = '' # invalid state
expect(user.valid?).to be false # run validations
expect(user.errors[:email]).to include("can't be blank") # check for presence of error
```
We don't need to save the record to check for it's validity

spec/models/user_spec.rb
```diff
@@ -135,6 +135,53 @@
...
+      end
+
+      context 'with valid one' do
+        it 'saves without any errors' do
+          user.password = 'waspAr$0'
+          expect(user.valid?).to be true
+        end
+      end
+
+      context 'with invalid one' do
```
hmasila: I think we can be more descriptive in this context. We are talking about passwords, so something like

```ruby
context 'when password is invalid'
....
end
```

app/models/user.rb
jm96441n: oh weird, my screen must have cut off the end of the conditional (or reading while tired) I totally read this as
```
access_expires_at && access_expires_at
```
my bad!

app/models/user.rb
jm96441n: that's true, but in terms of readability I think it's easier to read:
```
return if valid_password?
```
rather than
```
return if password.blank? || (password =~ PWD_REGEX)
```
Uncle Bob recommends sticking just about every boolean statement that's part of a conditional into it's own method in the book `Clean Code` just for the sake of making code easier to read
https://simpleprogrammer.com/refactoring-boolean-conditions-into-methods/

spec/models/user_spec.rb
yasirazgar: This will not work in our case, this will still throw error from devise, and also this way it is bit confusing as we are checking errors on email, I have updated the test to use something similar and avoided saving the record.

yasirazgar:
@julianguyen @hmasila I have addressed all the comments and moved the password validation part into separate concern, as got warning from codeclimate for user class length and also few more methods related password validation may get added by #1153.

app/models/concerns/password_validatable.rb
```diff
@@ -0,0 +1,20 @@
...
+  private
+
+  def password_complexity
+    return if valid_pwd?
+
+    err_mess = I18n.t('devise.registrations.password_complexity_error')
+    errors.add(:password, err_mess)
+  end
+
+  def valid_pwd?
```
hmasila: Opinion: Write full words instead of short forms
```suggestion
  def valid_password?
```

app/models/concerns/password_validatable.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module PasswordValidatable
+  extend ActiveSupport::Concern
+
+  PWD_REGEX = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).*$/
+
+  private
+
+  def password_complexity
+    return if valid_pwd?
```
hmasila:
```suggestion
    return if valid_password?
```

app/models/concerns/password_validatable.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module PasswordValidatable
+  extend ActiveSupport::Concern
+
+  PWD_REGEX = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).*$/
```
hmasila:
```suggestion
  PASSWORD_REGEX = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).*$/
```

app/models/concerns/password_validatable.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  private
+
+  def password_complexity
+    return if valid_pwd?
+
+    err_mess = I18n.t('devise.registrations.password_complexity_error')
```
hmasila:
```suggestion
    error_message = I18n.t('devise.registrations.password_complexity_error')
```

app/models/concerns/password_validatable.rb
```diff
@@ -0,0 +1,20 @@
...
+    errors.add(:password, err_mess)
```
hmasila:
```suggestion
    errors.add(:password, error_message)
```

app/models/concerns/password_validatable.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module PasswordValidatable
```
hmasila: The name of this module sounds like a polymorphic relationship. 
```suggestion
module PasswordValidator
```

app/models/user.rb
```diff
@@ -44,6 +44,8 @@
 #
 
 class User < ApplicationRecord
+  include PasswordValidatable
```
hmasila:
```suggestion
  include PasswordValidator
```

app/models/user.rb
```diff
@@ -160,6 +159,10 @@ def update_access_token
 
   private
 
+  def google_access_token_expired?
```
hmasila: 👍

spec/models/user_spec.rb
```diff
@@ -135,6 +135,47 @@
...
+  describe '#validations' do
+    context 'password' do
+      let(:user) { build(:user, password: nil) }
+
+      context 'when password is blank' do
+        it 'should throw only password error from devise' do
+          expect(user.valid?).to be false
+
+          expect(user).to have(1).error_on(:password)
+          expect(user.errors[:password]).not_to include(I18n.t('devise.registrations.password_complexity_error'))
```
hmasila: 👍

app/models/concerns/password_validatable.rb
yasirazgar: There is already a method `valid_password?` in devise, which conflicts with this method, I will change other suggested method and variable names.

jm96441n: I would then use a different method name, and a comment explaining the method

jm96441n: Especially if there is already a `valid_password` method in devise, it's going to be easy to get these two confused

spec/models/user_spec.rb
jm96441n: The description for this test no longer matches what the test does here. We also need some feature specs around this functionality

yasirazgar:
Updated variable and method names, 
@jm96441n functional/integration tests for this seems repetitive to me, all these user creation/registration part is taken care by devise, additional strength user' s password which is added is already tested at model level in user model what do you think.
@julianguyen If everything is good can you merge this branch, I need these changes for #1153.

julianguyen:
@yasirazgar I think it's fine not to have integration tests here. Could you just fix the merge conflict issue and it should be good to merge. Thanks!

yasirazgar:
@julianguyen, resolved conflicts.

Issue #1162, opened 2018-10-14
Text in language selection dropdown shows up as white on a white background


CSumm:
# Description 
Text in language selection dropdown shows up as white on a white background.

# Reproduction Steps

Enter If-Me.org.
Scroll down to footer where language dropdown is located.
Open language dropdown by clicking on it.

# Screenshots
![accessibilityissue_language dropdown](https://user-images.githubusercontent.com/31711543/46921317-e576a680-cfc7-11e8-803e-d0b6f36aa3e3.jpg)

# Do you want to be the assignee to work on this?

✅

CSumm:
This seems to occur only on a PC. Tried on Ubuntu and didn't receive this error.

gpetrioli:
The real issue here is that `option` element does not support opacity. Now, the `select` element is styled with opacity (*it is white but with and alpha channel of 0.1 out of 1*) but the `option` discard the alpha channel and show as white. 

We would need to define a specific color, but stylistically it cannot match the `select` because, as it is semi-transparent, it is defined by the background image in the footer.

We could use media queries to approximate a color based on what is shown behind the select box but we can never be exact.

Let me know if you want me to proceed with this and if the proposed solution is adequate.

julianguyen:
@gpetrioli You're spot on 👌 @CSumm started a [PR](https://github.com/ifmeorg/ifme/pull/1167) for this. It would be great if you could help review it and provide suggestions.

CSumm:
I think his explanation of how to fix it seems like the answer. We could work together on this issue if you would like.

CSumm:
I think working on the PR together would be best. The PR I wrote was rather sloppy and was close to irreversibly bad.

gpetrioli:
Ok, after some crossbrowser testing i see that most browsers do not show transparent options (*when the select box is open*) but they are just not affected because they also do not change the color of the `option` elements. 

So just setting `color: initial` fixes the problem everywhere.

PR #1167, opened 2018-10-15
Added operating system recognition to override UI

CSumm:
# Description

Added operating system recognition to override UI. Used window.navigator.appVersion to check what OS the browser is running on. This returns a string of the OS which is always truthy, and logically and it to the two classNames.

## More Details

The issue fixes the problem on Windows browser UI that places white text on a white background.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1162

# Test Coverage

✅


package.json
```diff
@@ -1,7 +1,9 @@
 {
   "name": "ifme",
   "private": true,
-  "dependencies": {},
+  "dependencies": {
```
julianguyen: We don't need this dependency

client/package.json
```diff
@@ -42,7 +42,9 @@
     "react-dom": "^15.5.4",
     "react-intl": "^2.1.5",
     "react-lazyload": "^2.3.0",
-    "react-on-rails": "9.0.3",
+    "react-on-rails": "8.0.6",
```
julianguyen: Let's unstage this change

client/app/stories/SideNav.stories.jsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
```
julianguyen: Let's unstage this change

client/app/components/SideNav/SideNav.scss
```diff
@@ -0,0 +1,52 @@
+@import "../../styles/_base.scss";
```
julianguyen: Let's unstage this change

client/app/components/Dropdown/Dropdown.scss
```diff
@@ -20,7 +20,6 @@
   appearance: $appearance;
 }
 
-
```
julianguyen: Let's unstage this change

client/app/components/Dropdown/DropdownFactory.jsx
```diff
@@ -23,7 +23,7 @@ export const DropdownFactory = (variationClassName: string) => {
     value: propValue,
   }: Props) => (
     <select
-      className={`${css.dropdown} ${variationClassName}`}
+      className={window.navigator.appVersion && `${css.dropdown} ${variationClassName}`}
```
julianguyen: Could you clarify that this change does?

julianguyen: It also looks like you haven't rebased from master in a while, this file actually no longer exists.

client/app/components/SideNav/index.jsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  handleCloseClick() {
+    this.setState({ showMenu: false });
+  }
+
+  render() {
+    return (
+      <Router>
+        <div className={navColor}>
```
jm96441n: it it possible to build these iteratively rather than hardcoding all these `li` elements?

julianguyen: This file shouldn't be included in this PR. It also looks like you haven't rebased from master in a while.

client/app/components/Dropdown/DropdownFactory.jsx
CSumm: I was attempting to get the OS on which the browser was running and whichever it was also use the css. However, it was guesswork since I never used OS compatibility for javascript before.

CSumm:
I'm going to get on this this weekend hopefully. I would also like some help. I read a comment on the issue I opened up and would like to ask the person for some help since he knows a lot more.

Issue #1174, opened 2018-10-16
Write tests for profile_controller


gr1d99:
# Description 
It seems we don't have tests for **profile_controller**

# Do you want to be the assignee to work on this?

✅

PR #1176, opened 2018-10-16
Add tests for profile_controller.rb

gr1d99:
# Description 
This PR adds test for `profile_controller.rb`

## Corresponding Issue
[#1174](https://github.com/ifmeorg/ifme/issues/1174)

# Screenshots

# Test Coverage

✅

### DIFF
--- a/app/controllers/profile_controller.rb
+++ b/app/controllers/profile_controller.rb
@@ -9,11 +9,9 @@ class ProfileController < ApplicationController
     user = current_user if user.nil?
 
     # Determine how the profile should be displayed based on the user_id
-    @stories = if user == current_user
-                 paginate_stories(current_user)
-               elsif current_user.allies_by_status(:accepted).include?(user)
-                 paginate_stories(user)
-               end
+    if user == current_user || current_user.mutual_allies?(user)
+      @stories = Kaminari.paginate_array(get_stories(user)).page(params[:page])
+    end
 
     @profile = user
   end
--- a/app/controllers/profile_controller.rb
+++ b/app/controllers/profile_controller.rb
@@ -10,10 +10,9 @@ class ProfileController < ApplicationController
 
     # Determine how the profile should be displayed based on the user_id
     @stories = if user == current_user
-                 Kaminari.paginate_array(get_stories(current_user, false))
-                         .page(params[:page])
+                 paginate_stories(current_user)
                elsif current_user.allies_by_status(:accepted).include?(user)
-                 Kaminari.paginate_array(get_stories(user, false))
+                 paginate_stories(user)
                end
 
     @profile = user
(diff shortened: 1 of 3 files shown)


spec/controllers/profile_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+describe ProfileController do
+  describe 'GET #index' do
+    context 'when user is not signed in' do
+      before { get :index, params: { 'uid': 'user-id' } }
+      it { expect(response).to redirect_to(new_user_session_path) }
+    end
+
+    context 'when user is logged in' do
```
julianguyen: We should combine the two duplicate `when user is logged in` contexts

julianguyen: We want to test when a user is logged in and looking at their own profile and when a user is logged in a looking at someone else's profile

app/controllers/profile_controller.rb
```diff
@@ -9,13 +9,13 @@ def index
     user = current_user if user.nil?
 
     # Determine how the profile should be displayed based on the user_id
-    if user == current_user
-      @stories = Kaminari.paginate_array(get_stories(current_user, false))
+    @stories = if user == current_user
+                 Kaminari.paginate_array(get_stories(current_user, false))
```
hmasila: The two statements inside the conditional are similar. 
Maybe we could have

```ruby
if user == current_user || current_user.allies_by_status(:accepted).include?(user)
  @stories = Kaminari.paginate_array(get_stories(user, false)).page(params[:page])
end
```
better enough, we could put `user == current_user || current_user.allies_by_status(:accepted).include?(user)` in a variable

hmasila: From my suggestion my aim was to do away with the elseif block. That is why I did
```ruby
if user == current_user || current_user.allies_by_status(:accepted).include?(user)
  @stories = Kaminari.paginate_array(get_stories(user, false)).page(params[:page])
end
```

We also don't need a separate method for the pagination

app/controllers/profile_controller.rb
```diff
@@ -9,11 +9,9 @@ def index
     user = current_user if user.nil?
 
     # Determine how the profile should be displayed based on the user_id
-    @stories = if user == current_user
-                 paginate_stories(current_user)
-               elsif current_user.allies_by_status(:accepted).include?(user)
-                 paginate_stories(user)
-               end
+    if user == current_user || current_user.allies_by_status(:accepted).include?(user)
```
hmasila: This line is too long

julianguyen: You should be able to use `current_user.mutual_allies?(user)` for the allies check

Issue #1051, opened 2018-10-02
Add missing test coverage for medication_helper.rb


julianguyen:
# Description 

Womp womp we're missing test coverage here. Let's go back and add it!

# Do you want to be the assignee to work on this?

🚫

timorthi:
I can give this a go :)

timorthi:
Did someone already work on this @julianguyen ?

julianguyen:
Actually no it was medication_refill_helper.rb, sorry about that! It's still open!

timorthi:
I won't be able to get on this as soon as I thought I could, so I'm gonna unassign myself and let someone else pick this up. Otherwise I'll reassign myself when I'm ready to work on it :)

haleymnatz:
@julianguyen I'll take this on! I'm excited to have gotten a chance to contribute to ifme again. 🎉 

Also, thanks @timorthi for great communication about your contribution! I assigned myself an issue before and forgot to remove myself when I couldn't get to it :/ So thanks for not doing the same thing!

julianguyen:
Yay @haleymnatz :D Thanks so much. Hope you're doing decent 💜

PR #1179, opened 2018-10-17
Add missing test coverage for medication_helper.rb

wilburhimself:
# Description 

Add missing test coverage for medication helper

## Corresponding Issue

[https://github.com/ifmeorg/ifme/issues/1051](https://github.com/ifmeorg/ifme/issues/1051)

# Test Coverage

✅

### DIFF
--- a/spec/helpers/medications_helper_spec.rb
+++ b/spec/helpers/medications_helper_spec.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+describe MedicationsHelper do
+  let(:current_user) { create(:user) }
+  let(:medication) { create(:medication, user: current_user) }
+
+  before(:each) do
+    @medication = medication
+  end
+
+  def get_field(field_name)
+    @fields.find {|f| f[:id] == field_name}
+  end
+
+  describe 'fields' do
+    before(:each) do
+      @fields = common_fields
+    end
+
+    it 'returns common medication fields' do
+      expect(get_field('medication_name')[:value]).to eq(@medication.name)
+      expect(get_field('medication_strength')[:value]).to eq(@medication.strength)
+      expect(get_field('medication_strength_unit')[:value]).to eq(@medication.strength_unit)
+      expect(get_field('medication_total')[:value]).to eq(@medication.total)
+      expect(get_field('medication_total_unit')[:value]).to eq(@medication.total_unit)
+      expect(get_field('medication_dosage')[:value]).to eq(@medication.dosage)
+      expect(get_field('medication_dosage_unit')[:value]).to eq(@medication.dosage_unit)
+      expect(get_field('medication_refill')[:value]).to eq(@medication.refill)
+      expect(get_field('medication_comments')[:value]).to eq(@medication.comments)
+    end
+
+    context 'when refill reminder is active' do
+      let(:medication) { create(:medication, :with_refill_reminder, user: current_user) }
+      it 'medication_refill_reminder is set' do
+        @medication = medication
+        expect(get_field('medication_refill_reminder')[:value]).to eq(true)
+        expect(get_field('medication_refill_reminder_attributes_id')[:value]).to eq(@medication.refill_reminder.id)
+      end
+    end
+
+    context 'when take medication reminder is active' do
+      let(:medication) { create(:medication, :with_daily_reminder, user: current_user) }
+      it 'medication_take_medication_reminder is set' do
+        @medication = medication
+        expect(get_field('medication_take_medication_reminder')[:value]).to eq(true)
+        expect(get_field('medication_take_medication_reminder_attributes_id')[:value]).to eq(@medication.take_medication_reminder.id)
+      end
+    end
+  end
+
+  describe 'google_fields' do
+    before(:each) do
+      @fields = google_fields
+    end
+
+    it 'adds google field to common fields' do
+      expect(get_field('medication_add_to_google_cal')[:value]).to eq(true)
+    end
+  end
+
+  describe 'days_checkbox' do
+    before { @weekdays = days_checkbox }
+
+    it 'returns weekly dosage checkboxes information' do
+      0.upto(6).each do |i|
+        expect(@weekdays[i][:id]).to eq("medication_weekly_dosage_#{i}")
+        expect(@weekdays[i][:checked]).to eq(@medication.weekly_dosage.include?(i))
+      end
+    end
+  end
+end
--- a/spec/helpers/medications_helper_spec.rb
+++ b/spec/helpers/medications_helper_spec.rb
@@ -18,15 +18,15 @@ describe MedicationsHelper do
     end
 
     it 'returns common medication fields' do
-      expect(get_field('medication_name')[:value]).to eq(@medication.name)
-      expect(get_field('medication_strength')[:value]).to eq(@medication.strength)
-      expect(get_field('medication_strength_unit')[:value]).to eq(@medication.strength_unit)
-      expect(get_field('medication_total')[:value]).to eq(@medication.total)
-      expect(get_field('medication_total_unit')[:value]).to eq(@medication.total_unit)
-      expect(get_field('medication_dosage')[:value]).to eq(@medication.dosage)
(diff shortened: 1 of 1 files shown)


spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  def get_field(field_name)
+    @fields.find {|f| f[:id] == field_name}
+  end
+
+  describe 'fields' do
+    before(:each) do
+      @fields = common_fields
+    end
+
+    it 'returns common medication fields' do
```
julianguyen: So you'll want to test the direct output of new_medication_props and edit_medication_props rather than common_fields which is a private method.

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+describe MedicationsHelper do
+  let(:current_user) { create(:user) }
+  let(:medication) { create(:medication, user: current_user) }
+
+  before(:each) do
+    @medication = medication
```
hmasila: Why are we converting this to an instance method?

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+describe MedicationsHelper do
+  let(:current_user) { create(:user) }
+  let(:medication) { create(:medication, user: current_user) }
+
+  before(:each) do
```
hmasila: Since the value of medication does not change;

`before do` is just fine

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    @medication = medication
+  end
+
+  def get_field(field_name)
+    @fields.find {|f| f[:id] == field_name}
```
hmasila: What's the purpose of this method. It just seems wrong to compare name with id

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(get_field('medication_total_unit')[:value]).to eq(@medication.total_unit)
+      expect(get_field('medication_dosage')[:value]).to eq(@medication.dosage)
+      expect(get_field('medication_dosage_unit')[:value]).to eq(@medication.dosage_unit)
+      expect(get_field('medication_refill')[:value]).to eq(@medication.refill)
+      expect(get_field('medication_comments')[:value]).to eq(@medication.comments)
+    end
+
+    context 'when refill reminder is active' do
+      let(:medication) { create(:medication, :with_refill_reminder, user: current_user) }
+      it 'medication_refill_reminder is set' do
```
hmasila: Try and make this as natural as possible. You could have something like

`it 'sets a medication refill reminder' do `

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        @medication = medication
```
hmasila: I think we can use `medication` as it is without converting it to an instance variable

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        expect(get_field('medication_refill_reminder')[:value]).to eq(true)
+        expect(get_field('medication_refill_reminder_attributes_id')[:value]).to eq(@medication.refill_reminder.id)
+      end
+    end
+
+    context 'when take medication reminder is active' do
+      let(:medication) { create(:medication, :with_daily_reminder, user: current_user) }
+      it 'medication_take_medication_reminder is set' do
+        @medication = medication
```
hmasila: I think we can use `medication` as it is without converting it to an instance variable

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      it 'medication_refill_reminder is set' do
+        @medication = medication
+        expect(get_field('medication_refill_reminder')[:value]).to eq(true)
+        expect(get_field('medication_refill_reminder_attributes_id')[:value]).to eq(@medication.refill_reminder.id)
+      end
+    end
+
+    context 'when take medication reminder is active' do
+      let(:medication) { create(:medication, :with_daily_reminder, user: current_user) }
+      it 'medication_take_medication_reminder is set' do
```
hmasila: Try and make this as natural as possible. You could have something like

`it 'sets a medication take reminder' do`

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        @medication = medication
+        expect(get_field('medication_take_medication_reminder')[:value]).to eq(true)
+        expect(get_field('medication_take_medication_reminder_attributes_id')[:value]).to eq(@medication.take_medication_reminder.id)
+      end
+    end
+  end
+
+  describe 'google_fields' do
+    before(:each) do
+      @fields = google_fields
```
hmasila: I don't think we need an instance variable here

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    it 'adds google field to common fields' do
+      expect(get_field('medication_add_to_google_cal')[:value]).to eq(true)
+    end
+  end
+
+  describe 'days_checkbox' do
+    before { @weekdays = days_checkbox }
```
hmasila: I don't think we need to instantiate an instance variable here

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      let(:medication) { create(:medication, :with_daily_reminder, user: current_user) }
+      it 'medication_take_medication_reminder is set' do
+        @medication = medication
+        expect(get_field('medication_take_medication_reminder')[:value]).to eq(true)
+        expect(get_field('medication_take_medication_reminder_attributes_id')[:value]).to eq(@medication.take_medication_reminder.id)
+      end
+    end
+  end
+
+  describe 'google_fields' do
```
hmasila: Use a `#` for instance methods and `.` for class methods.

This should be `describe '#google_fields' do`

spec/helpers/medications_helper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    before(:each) do
+      @fields = google_fields
+    end
+
+    it 'adds google field to common fields' do
+      expect(get_field('medication_add_to_google_cal')[:value]).to eq(true)
+    end
+  end
+
+  describe 'days_checkbox' do
```
hmasila: Use a `#` for instance methods and `.` for class methods.

This should be `describe '#days_checkbox' do`

hmasila:
```suggestion
  describe '#days_checkbox' do
```

spec/helpers/medications_helper_spec.rb
hmasila:
```suggestion
  describe '#google_fields' do
```

spec/helpers/medications_helper_spec.rb
hmasila:
```suggestion
      it 'sets a medication take reminder' do
```

spec/helpers/medications_helper_spec.rb
hmasila:
```suggestion
      it 'sets a medication refill reminder' do
```

spec/helpers/medications_helper_spec.rb
wilburhimself: because in `medications_helper#common_fields` the medication variable is called as an instance variable. [Example](https://github.com/ifmeorg/ifme/blob/master/app/helpers/medications_helper.rb#L29).

spec/helpers/medications_helper_spec.rb
wilburhimself: This method just finds the field where the field id is `field_name`. The id in this case is the field name. [Example](https://github.com/ifmeorg/ifme/blob/master/app/helpers/medications_helper.rb#L25).

PR #1186, opened 2018-10-17
Tests for MeetingFormHelper and clean up MeetingNotificationsService

julianguyen:
# Description 

- Add tests for MeetingsHelper
- Create MeetingFormHelper
- Clean up MeetingNotificationsService

# Test Coverage

✅


app/helpers/meeting_form_helper.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+module MeetingFormHelper
+  include FormHelper
+
+  def new_meeting_props(group_id)
+    new_form_props(meeting_form_inputs(nil, group_id), meetings_path)
```
hmasila: Opinion: pass objects to methods. In this case 

```ruby
def new_meeting_props(group)
  new_form_props(meeting_form_inputs(nil, group.id), meetings_path)
end
```

app/services/meeting_notifications_service.rb
```diff
@@ -19,49 +18,59 @@ def self.handle_members(args = {})
...
-        user: current_user.name,
-        group_id: meeting.group_id,
-        group: group,
-        typename: meeting.name,
-        type: type,
-        uniqueid: uniqueid
-      }.to_json
-    end
+  def get_notifications(member, data)
+    Notification.create(
```
hmasila: what happens if the notification is not created? I think we need to cater for that

app/services/meeting_notifications_service.rb
```diff
@@ -19,49 +18,59 @@ def self.handle_members(args = {})
...
-      group: group,
+      group: meeting.group.name,
       typename: meeting.name,
       type: type,
       uniqueid: uniqueid
     }.to_json
   end
-  # rubocop:enable MethodLength
+
+  def notification_data
```
hmasila: 👍 nice refactor

app/services/meeting_notifications_service.rb
julianguyen: Thanks for the important call out! Updated!

Issue #1182, opened 2018-10-17
Adding missing tests for CategoriesHelper


julianguyen:
# Description 

Add missing tests! Use as an opportunity to refactor code (including getting rid of Rubocop disable comments like `# rubocop:disable MethodLength`)

# Do you want to be the assignee to work on this?

🚫

Veith3n:
I can do this one 😊

PR #1193, opened 2018-10-18
add tests for categories_helper / resolves #1184

Veith3n:
# Description 
Add tests for categories_helper with little code refactoring
## Corresponding Issue
[#1184](https://github.com/ifmeorg/ifme/issues/1184#issue-370871368)
# Test Coverage
✅

PR #1194, opened 2018-10-18
resolves [#1182] Add tests for moods_helper 

Veith3n:
# Description 
Add tests for categories_helper with little code refactoring
## Corresponding Issue
[#1182](https://github.com/ifmeorg/ifme/issues/1182#issue-370871031)
# Test Coverage
✅

### DIFF
--- a/spec/helpers/moods_helper_spec.rb
+++ b/spec/helpers/moods_helper_spec.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+describe MoodsHelper do
+  subject { TestClass.new(build(:mood, user_id: 999, id: 999)) }
+
+  class TestClass
+    include MoodsHelper
+    include Rails.application.routes.url_helpers
+
+    def initialize(mood)
+      @mood = mood
+    end
+
+    def t(*args)
+      I18n.t(*args)
+    end
+  end
+
+  let(:mood_form_inputs_res) do
+    [
+      {
+        id: 'mood_name',
+        type: 'text',
+        name: 'mood[name]',
+        label: 'Name',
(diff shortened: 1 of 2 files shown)


spec/helpers/moods_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  class TestClass
+    include MoodsHelper
+    include Rails.application.routes.url_helpers
+
+    def initialize(mood)
+      @mood = mood
+    end
+
+    def t(*args)
```
hmasila: Is this really necessary? I haven't seen it's use anywhere in the tests

spec/helpers/moods_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+describe MoodsHelper do
+  subject { TestClass.new(build(:mood, user_id: 999, id: 999)) }
+
+  class TestClass
```
hmasila: I think we can move the initialisation to a before block and do away with the testClass

eg 
```ruby
before do
  @mood = mood
end
```

And tests will be in the format

```ruby
describe '#new_mood_props' do
    it 'calls method with correct params' do
      allow(FormHelper).to receive(:new_form_props).with(mood_form_inputs_res, '/moods').and_return(":new_form_props called")
      expect(FormHelper).to receive(:new_form_props).with(mood_form_inputs_res, '/moods')
      expect(new_mood_props).to eq ":new_form_props called"
    end
  end

hmasila:
This PR is similar to https://github.com/ifmeorg/ifme/pull/1193

PR #1196, opened 2018-10-19
Refactor User model to address method length

ludamillion:
# Description 

Extract updating of Google OAuth2 access token fields on user to private class method. Previously the `find_for_google_oauth2` method was wrapped in a Rubocop exemption. Reducing the line count remove the need for that exemption.

Additionally there was a TODO listed for that method as well to remove an unused parameter. This commit removes that unused param and updates the locations where `find_for_google_oauth2` is called accordingly. 

## Corresponding Issue

This resolves one of the items listed in https://github.com/ifmeorg/ifme/issues/840

# Test Coverage

✅


app/models/user.rb
```diff
@@ -96,25 +96,16 @@ def available_groups(order)
     ally_groups.order(order) - groups
   end
 
-  # TODO: _signed_in_resource is unused and should be removed
-  # rubocop:disable MethodLength
-  def self.find_for_google_oauth2(access_token, _signed_in_resource = nil)
+  def self.find_for_google_oauth2(access_token)
```
hmasila: 👍

ludamillion:
⚖️

PR #1198, opened 2018-10-19
#1181 - Add missing tests for strategies helper

JEscalanteGT:
# Description 

Add missing tests for the helper and refactor some methods to avoir rubocop method lenght and module length alert.

## Corresponding Issue

#1181 

# Test Coverage

✅

### DIFF
--- a/app/helpers/strategies_helper.rb
+++ b/app/helpers/strategies_helper.rb
@@ -1,9 +1,8 @@
 # frozen_string_literal: true
 
-# rubocop:disable ModuleLength
 module StrategiesHelper
-  include CategoriesHelper
   include FormHelper
+  include StrategyFormInputs
   include ViewersHelper
 
   def new_strategy_props(strategy, viewers)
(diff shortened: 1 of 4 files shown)


app/helpers/strategies_helper.rb
```diff
@@ -28,99 +28,20 @@ def edit_strategy_props(strategy, viewers)
...
-    @categories.each do |item|
-      checkboxes.push(
-        id: item.slug,
-        label: item.name,
-        value: item.id,
-        checked: @strategy.category.include?(item.id)
-      )
-    end
-    checkboxes
+    strategy_form_inputs(@strategy, @viewers)[0..1]
```
hmasila: I think this will be more readable if it is 
```suggestion
    [
        build_strategy_name(@strategy),
        build_strategy_description(@strategy)
    ]
```
@julianguyen Correct me if I'm wrong

app/helpers/strategies_helper.rb
```diff
@@ -28,99 +28,20 @@ def edit_strategy_props(strategy, viewers)
...
-      {
-        id: 'strategy_category',
-        type: 'quickCreate',
-        name: 'strategy[category][]',
-        label: t('categories.plural'),
-        placeholder: t('common.form.press_enter'),
-        checkboxes: category_checkboxes,
-        formProps: quick_create_category_props
-      },
+      build_strategy_name(strategy),
```
hmasila: 👍

app/helpers/strategies_helper.rb
julianguyen: Whoa embedded code snippets - that's new
Agreed, I think it's readable the way @hmasila suggested!

app/helpers/strategy_form_inputs.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+
+module StrategyFormInputs
```
julianguyen: Just to be consistent with naming, let's rename this to StrategyFormHelper

app/helpers/strategies_helper.rb
JEscalanteGT: Thanks for the feedback! I will make these changes in the next days

Issue #563, opened 2017-08-25
Add tag filtering to Resources page


julianguyen:
The [Resources](https://www.if-me.org/resources) page is getting long and we want to best help point folks to mental health resources all over the interwebs. Let's add a feature to filter resources by tags (including languages that they're offered in).

You can find the designs for that page here: https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me

rarwolf47:
Is this being worked by anyone? I'd like to try my hand at it.

baohouse:
@rarwolf47 Looks like it's assigned to @katrynmc ; you'll want to sync up with her and see where you can pick up some of the pieces.

katrynmc:
@baohouse I'd say grab it, I haven't had the time lately - hadn't started

andy5995:
@baohouse looks looks like you reassigned it to katrynmc instead of rarwolf47

baohouse:
I can't assign it to @rarwolf47; probably he has to fork the project first?

rarwolf47:
How about now?

baohouse:
@julianguyen Would you have any luck with reassigning this? Not sure why I can't.

julianguyen:
@rarwolf47 You can only assign collaborators of the repo to issues, I sent you an invite! You should accept via email!

rarwolf47:
Done!

aamarill:
@andy5995 I can work on this. Do we know, roughly, how we want this implemented? thanks!

aamarill:
Unfortunately, I can't work on this at the moment. Thank you for expressing interest @seanmfox

julianguyen:
@seanmfox Just case you don't find/notice it, we have an existing custom Input component which you can preview in Storybook that you can use.

seanmfox:
Great, thanks!  In looking at past conversations, what was in mind for how the tag filtering should work?  In finding some past conversations, ideas being tossed around seemed to include dropdown, word cloud, and search box.

PR #1200, opened 2018-10-20
Issue 563 - Resources Filtering

seanmfox:
# Description 

A filter feature has been added to the Resources page allowing a user to narrow resource options based on tags.  This feature uses the `InputTag` component to provide similar functionality as other areas of the site.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/563

# Screenshots

image

image

image

# Test Coverage

✅


julianguyen:
Can you please rebase from master? There was a change made to the default theme colour so that it's white instead of black.

client/app/components/Input/Input.scss
```diff
@@ -168,6 +168,11 @@
   }
 }
 
+.resources {
+  color: #FFFFFF;
```
julianguyen: Let's use `$white` instead

client/app/components/Input/InputCheckbox.jsx
```diff
@@ -60,7 +66,7 @@ export const InputCheckbox = (props: Props) => {
           }
           aria-label={label.replace(/<\/?[^>]+(>|$)/g, '')}
         />
-        <div className={css.checkboxLabel}>{renderHTML(label)}</div>
+        <div className={`${css.checkboxLabel} ${resourcesPageDisplay(resourcesPage)}`}>{renderHTML(label)}</div>
```
julianguyen: It would be better to pass in the css as props instead of checking the via the window.body whether a class exists. You can also override CSS as well.

client/app/widgets/Resources/index.jsx
```diff
@@ -3,25 +3,84 @@ import React from 'react';
...
+  };
+  
+  filterList = (check: Array<*>) => {
+    const { resources } = this.props;
+    const selectedTags = check.filter(c => c.checked === true);
+    const matchingResources = resources.filter(r => {
+      const tagCheck = selectedTags.map((t: Selected)  => 
+        r.tags.concat(r.languages).includes(t.id)
+      );
+      if(tagCheck.includes(false)) {
```
julianguyen: Let's replace this block with `return !tagCheck.includes(false)`

client/app/widgets/Resources/index.jsx
```diff
@@ -3,25 +3,84 @@ import React from 'react';
...
+  
+  render() {
+    const { checkboxes } = this.state;
+    const filteredResources = this.filterList(checkboxes);
+    return (
+      <React.Fragment>
+          <InputTag
+            id='resourceTags'
+            name='resourceTags'
+            placeholder='Press ENTER to add'
```
julianguyen: We can use i18n here, look at other components in client for an example of how to do this.

client/app/widgets/Resources/index.jsx
```diff
@@ -3,25 +3,84 @@ import React from 'react';
...
+
+export class Resources extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = { checkboxes: this.createCheckboxes() };
+  }
+  
+  createCheckboxes = () => {
+    const { resources } = this.props;
+    const tagsList = [...new Set(resources.map(res => res.tags.concat(res.languages)).reduce((acc, val) => acc.concat(val), []))];
```
julianguyen: ESLint probably complained that this line is too long. Let's fix this. Same for the next line.

julianguyen: Does object spread actually work? O.o I thought we were getting errors there.

seanmfox: No errors using it on my side as far as I can tell.  Do you remember what the error was?

client/app/widgets/Resources/index.jsx
```diff
@@ -3,25 +3,93 @@ import React from 'react';
...
+    return (
+      <React.Fragment>
+        <InputTag
+          id="resourceTags"
+          name="resourceTags"
+          placeholder={I18n.t('common.form.press_enter')}
+          checkboxes={checkboxes}
+          onCheckboxChange={box => this.checkboxChange(box)}
+        />
+        <div className={css.gridThree}>
```
julianguyen: Let's also add `css.marginTop` to this div so that there's spacing between the input field and results.

client/app/components/Input/InputTag.jsx
```diff
@@ -12,6 +12,7 @@ export type Props = {
   placeholder?: string,
   checkboxes: Checkbox[],
   onChange?: Function,
+  onCheckboxChange?: Function
```
julianguyen: Just to be consistent with linting, let's add a `,` at the end of what's on line 15.

client/app/widgets/Resources/__tests__/Resources.spec.jsx
```diff
@@ -0,0 +1,73 @@
...
+        tags: ['free', 'blog'],
+        languages: ['English'],
+        type: 'Communities',
+      },
+    ]}
+  />
+);
+
+describe('Resources', () => {
+  it('filter when tag selected', () => {
```
julianguyen: Minor fix: let's replace filter with filters

client/app/widgets/Resources/__tests__/Resources.spec.jsx
```diff
@@ -0,0 +1,73 @@
...
+    wrapper
+      .find('.tagLabel')
+      .at(0)
+      .simulate('click');
+    expect(wrapper.find('.checkboxLabel').text()).toEqual(id);
+    expect(wrapper.find('.resource').length).toEqual(1);
+    expect(wrapper.find('.tag').findWhere(t => t.text() === id).length).toEqual(1);
+  });
+
+  it('unfilter when tag unselected', () => {
```
julianguyen: Minor fix: let's replace unfilter with unfilters

client/app/widgets/Resources/index.jsx
seanmfox: Does the marginTop style already exist?  I can't seem to find it in any of the SCSS files.

julianguyen: Yeah you should see it in margins.scss. you might have to import global.scss from the component file, otherwise it should be importee via Resources.scss

PR #1207, opened 2018-10-24
Remove unnecessary data_type_check variable

gr1d99:
# Description 
Remove unnecessary `data_type_check` variable and rename `data_type_check` method to `valid_data_type?`.

# Test Coverage
🚫

### DIFF
--- a/app/helpers/pages_helper.rb
+++ b/app/helpers/pages_helper.rb
@@ -34,10 +34,6 @@ module PagesHelper
   end
 
   def valid_data_type?(data_type, data)
-    if data_type == 'partners'
-      data['image_link'].is_a?(String)
-    elsif data_type == 'resources'
-      data['tags'].is_a?(Array) && data['languages'].is_a?(Array)
-    end
+    data['image_link'].is_a?(String) if data_type == 'partners'
   end
 end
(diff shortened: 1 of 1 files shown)


app/helpers/pages_helper.rb
```diff
@@ -30,16 +30,10 @@ def valid_hash?(data_type, data)
...
-  def data_type_check(data_type, data)
-    if data_type == 'partners'
-      data_type_check = data['image_link'].is_a?(String)
-    elsif data_type == 'resources'
-      data_type_check = data['tags'].is_a?(Array) &&
-                        data['languages'].is_a?(Array)
-    end
-    data_type_check
+  def valid_data_type?(data_type, data)
+    data['image_link'].is_a?(String) if data_type == 'partners'
```
hmasila: what about if data_type is `resources` why did we stop checking for array?

gr1d99: The call to check `resources` data_type was never invoked anywhere in the application so decided to remove it, but I am open for any suggestions whether I should add another commit to add and implement check for `resources` data types

julianguyen: I'm confused, I think it is necessary if you remove data_type_check(data_type, data) then we won't be able to see content for Partners and Resources

gr1d99: It works as expected @julianguyen, I have tested it myself locally and it still works, let me request on slack for some few folks to test it out

julianguyen: Oh yes you're right about this since we're using the Resources widget now. So only partners matters here. Thanks for clarifying

Issue #1004, opened 2018-09-24
Add content_security_policy.rb initializer.


sexybiggetje:
# Description 

#976 and #1000 (as fixed by @mialahmoka)  leave open the part of adding a Content-Security-Policy file/header.
It should be added and configured properly for the application.

> In addition, the missing Content-Security-Policy header requires the additional file content_security_policy.rb which is missing from config/initializers/.

# Do you want to be the assignee to work on this?

🚫

sexybiggetje:
Maybe this is of some help; https://github.com/twitter/secure_headers

mialahmoka:
This might also be useful:
https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/content_security_policy.rb

mialahmoka:
I also feel like this is relevant. It's old but it could potentially be a threat. It details methods of bypassing HTTP HSTS (the 'max-age' parameter I added at the last line of the header config) via NTP exploits. Theoretically it shouldn't work because attempting to use the http version of the site should throw a 301 redirect, but it could be worth testing just in case.
https://www.google.com/url?sa=t&source=web&rct=j&url=https://www.blackhat.com/docs/eu-14/materials/eu-14-Selvi-Bypassing-HTTP-Strict-Transport-Security-wp.pdf&ved=2ahUKEwjjzpeD-tXdAhVsITQIHXblBtMQFjACegQICRAB&usg=AOvVaw1xu1y_xDSUVFGPI5aUt9PG&cshid=1537871598448

ludamillion:
I can take a pass as this. I'm not super security savvy per se, are there any particular gotchas I should look out for? For example would the default settings from the Twitter tool be appropriate?

sexybiggetje:
It would be your best start. I think it will mostly work, try it and browse around your dev instance.
At first glance it shouldn't cause any *major* issues.

sexybiggetje:
Thanks for this. I'll have to postpone looking at this but maybe @mialahmoka can verify and test?

ludamillion:
This was addressed by https://github.com/ifmeorg/ifme/pull/1208

PR #1208, opened 2018-10-24
Added initial generated version of CSP config

ludamillion:
# Description 

Adds an initialization file for the content security policy following up #976 and #1000. The options to initialize with where generated using the [Laboratory](https://github.com/april/laboratory) add-on for Firefox. 

## Corresponding Issue

This address issue #1004 

# Test Coverage

🚫 

Honestly, because I don't know how one tests this (programmatically at least) and haven't found a pattern to do so in spite of a fair amount of internet searching.


sexybiggetje:
I'm pretty sure that Pusher message is unrelated and because your instance isnt registered on that cluster. That's no problem.

PR #1214, opened 2018-10-27
Increases test coverage for `MeetingsController`

andrezacm:
# Description 

- Increases test coverage for `MeetingsController`
- Fix invalid variable in `MeetingsController#destroy`
- Removes AbstractController::DoubleRenderError warning from `MeetingsController`

# Test Coverage

✅


hmasila:
I had started the PR here but got too busy https://github.com/ifmeorg/ifme/pull/1117. I will close mine and review yours because you got further than me. Good work 🎉

app/controllers/meetings_controller.rb
```diff
@@ -89,9 +90,9 @@ def leave
 
   # DELETE /meetings/1
   def destroy
-    redirect_unless_leader_for(@meeting.group)
+    redirect_unless_leader_for(@meeting.group) && return
     # Notify group members that the meeting has been deleted
-    send_notification(@meeting, group.members, 'remove_meeting')
+    send_notification(@meeting, @meeting.group.members, 'remove_meeting')
```
hmasila: I think this should be `group.group_members`
```suggestion
    send_notification(@meeting, group.group_members, 'remove_meeting')
```

app/controllers/meetings_controller.rb
```diff
@@ -32,7 +32,8 @@ def edit
   def create
     @meeting = Meeting.new(meeting_params)
     @group = Group.find_by(id: meeting_params[:group_id])
-    redirect_unless_leader_for(@group)
+    redirect_unless_leader_for(@group) && return
```
hmasila: I think you can add `&& return` to `new` and `edit` too

andrezacm:
omg I'm sorry didn't notice you were already doing it 😅 
I must have missed your PR, thank you for your reviews

app/controllers/meetings_controller.rb
andrezacm: here I've added the `@meeting` first cause the variable group is not set up in this action and `@meeting` is and you can access the group through it.

And I've used `group.members` cause it sounded better than `group.group_members` and its a valid relationship defined in group model, is basically just a has_many throug group_members. 
https://github.com/ifmeorg/ifme/blob/master/app/models/group.rb#L20

Issue #776, opened 2018-01-03
French translation


panjim:
Hi everyone, i joined the french translation team few days ago.
I saw that many people worked on the translation for a while, but i couldn't find any of the documents to contribute, or at least do some proofreading...

We need those 4 docs:
-devise.fr.yml (from I18n)
-devise_invitable.fr.yml (trad. of devise_invitable.en.yml)
-fr.yml (trad. of en.yml)
-kaminari.fr.yml (trad. of kaminari.en.yml)

if you know something about it please tell me, before i translate everything an other time.

inesbenomar18:
Hi, I'd love to contribute to the French translation team! 👋🏼

hugodahl:
I'd be happy to help with the French translations as well if there is any more need for it

julianguyen:
@hugodahl We definitely need help, it's not being worked on at the moment.

Thanks for expressing your interest. We would love your help with this issue. Please follow these [next steps](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack) to join our contributor community.

simtaxx:
hey, i want to help you for tgis, how can i help ?

julianguyen:
@hugodahl @simtaxx This is our [process](https://github.com/ifmeorg/ifme/wiki/Translations) on translations. Please follow these [next steps](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack) to join our contributor community.

CCoupris:
Hi I translated the file "kaminari.fr.yml" but I'm a new member of GitHub so I don't know exactly how to upload the file in the project. Could you help me please ?

PR #1188, opened 2018-10-17
Add translation devise invitable french

CCoupris:
# Description 

I translated the content into french for "device_instable"

## Corresponding Issue

French translation #776

# Test Coverage

✅

PR #1217, opened 2018-10-28
Translate fr.yml

sylvain-ch21:
# Description 

French translation for fr.yml

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/776


julianguyen:
Looks like https://github.com/ifmeorg/ifme/pull/1188 covered some devise translation files. Changing this PR to merge into fr-translations instead of master.

config/locales/it.yml
```diff
@@ -24,6 +24,7 @@ it:
     it: Italiano
     nb: Norsk
     de: Deutsch
+    fr: Français
```
julianguyen: Looks like we're missing the same value for the other translation files.

Issue #607, opened 2017-09-07
Allies list should show previous invites


andy5995:
On the allies page, I think it would be helpful to show a list of allies that have been previously invited (unless they're in the "accepted" list), along with the expiration date. Perhaps a "*" next to invitations that have passed the expiration date.

Otherwise the user is forced to remember who he or she has emailed.

charlesskariah:
@julianguyen Any suggestions for this? May be I can have a look at this.?

andy5995:
Hi @charlesskariah maybe I can help you with this while you wait to hear from Julia. I created the ticket.

This file is where the list would be displayed:
config/locales/en.yml (along with the associated language files: es.yml and ptbr.yml

You'd be able to get translations from the Slack channels #es-translations and #br-translations

(If #585  is completed before you are done, then you'd also need to get Swedish translations)

To do this patch, I'm not sure **_how to get the info from the database to show recently invited allies and the dates those invites expire_**. We'll find out though, if you don't know either.

Are there any other questions I can help you with? Have you joined the if me Slack chatroom yet? Thanks for having a look at this ticket.

julianguyen:
@charlesskariah So currently our page looks like this: https://github.com/julianguyen/ifme/blob/master/app/views/allies/index.html.erb

We show outgoing and incoming ally requests and accepted allies.

So actually, I'm not sure what else we would add. But maybe we could add email invites in the outgoing section (people who are sent invites to join if me via email). I don't think we need to keep a record after they've accepted how they joined.

andy5995:
I believe we are thinking the same thing, Julia. This is what I wrote in the top post (but maybe I didn't word it very clearly):

> I think it would be helpful to show a list of allies that have been previously invited (unless they're in the "accepted" list), along with the expiration date. Perhaps a "*" next to invitations that have passed the expiration date.

julianguyen:
At the moment, email invitations don't expire. We could make them expire if we wanted to. Anyone in favour of that?

andy5995:
I got the impression they expired:

> Andy Alt has invited you to join http://if-me.org, a community to share mental health experiences with your loved ones and keep track of your moods, medications, and self-care. You can accept the invite through the link below.
> 
> Accept invitation!
> 
> **This invitation will expire on Wednesday September 20, 2017 at 09:37 PM.**

Though in my reply to Charles, I mentioned translations. I think the only translation needed would be for the word _expired_.

julianguyen:
[...]

PR #1218, opened 2018-10-28
Show previously invited allies

sansbrina:
# Description 

Shows previously invited via email but not yet accepted allies, the date
the ally was invited and indicates if the invitation has expired.
The email is listed as a mail_to link.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/607
# Screenshots
screen shot 2018-10-28 at 5 09 04 pm

# Test Coverage

✅ , I updated the spec for the controller but am totally open to adding coverage if necessary!


config/locales/en.yml
```diff
@@ -97,6 +97,9 @@ en:
       title: Allies
       outgoing: 'Outgoing Requests'
       incoming: 'Incoming Requests'
+      invited: 'Invited by email'
+      invited_on: 'Invited on: %{date}'
+      invitation_expired: 'Invitation expired?'
```
julianguyen: For now you can go ahead and get translations for our other languages via Google Translate

app/views/allies/index.html.erb
```diff
@@ -31,7 +31,19 @@
     </div>
   </div>
 <% end %>
-
+<% if @invited_allies.any? %>
+  <div class="title"><%= t('allies.index.invited') %></div>
+  <div class="spacer"></div>
```
julianguyen: We don't have the spacer class anymore. I would use marginTop in the same div where gridThree is.

config/locales/en.yml
julianguyen: I think it would be simpler just to show `invitation_expired: 'Invitation expired'` if that's the case and nothing if it's not.

app/views/allies/index.html.erb
```diff
@@ -31,7 +31,19 @@
...
+  <div class="gridThree">
+  <% @invited_allies.each do |ally| %>
+    <div class="gridThreeItemBoxLight ally">
+      <div><%= mail_to "#{ally.email}" %></div>
+      <div class="subtle"><%=t('allies.index.invited_on', date: format_date(ally.invitation_sent_at)) %></div>
+      <div class="subtle"><%= t('allies.index.invitation_expired') %> <%= ally.invitation_sent_at < 2.weeks.ago ? t('yes_text') : t('no_text') %></div>
```
julianguyen: We should only show this if invitation_accepted_at doesn't exist as well. I think if an invitation is accepted within 2 weeks, it will technically show up here.

I think it's useful just to show pending invitations rather than accepted ones.

sansbrina: Thanks for taking a look at this so quickly! 
I included `invitation_accepted_at: nil` in the query:
https://github.com/ifmeorg/ifme/blob/393edba96b2a9f9da120ec3a44cc20e1e4ea268b/app/controllers/allies_controller.rb#L14-L16
so I don't think they'll show up

PR #332, opened 2016-10-11
Add Docker Compose configuration files

esBeee:
Having Docker and Docker Compose installed, a user is now able to start the server using

```
$ docker-compose up
```

or open a shell using

```
$ docker-compose -f docker-compose.yml -f docker-compose.test.yml run --rm app bash
```

, wherein now also tests will pass, that depend on Selenium and Firefox (also see #317).

PR #887, opened 2018-04-16
Adding Vagranfile and updating .gitignore and Procfile.dev

localhostpedro:
This Vagratnfile will help create detached development environments from the computer, which I'm sure will reduce the amount of misconfiguration for new developers.

PR #1223, opened 2018-10-29
Get docker working again

geppy:
# Description 

Updates the Docker configurations, which have broken over time.

## More Details

- updates `Dockerfile` to use the `ruby:2.3` base image
- combines steps in the `Dockerfile` so the image has fewer layers
- installs managed dependencies outside of `/app`, which is likely to be mounted in dev environments
  - `.yarnrc` tells Yarn to install packages in `/node_modules` ([which will be found by Node's module loader](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders))
  - base `ruby:2.3` image tells Bundler to install gems in `/usr/local/bundle`
- adds `.dockerignore`, so unnecessary files aren't copied into the Docker image
- moves dev environment settings into `docker-compose.overrides.yml`
- adds back in `spec_helper.rb` settings that make Docker-based testing work
  - these settings were originally added in #332, but removed in 69e0c38

# Test Coverage

No coverage changes.


geppy:
@julianguyen You're very welcome!

Do the wiki docs need updating?  [Installation](https://github.com/ifmeorg/ifme/wiki/Installation) could be reorganized, removing the Vagrant bits (since #887  was closed unmerged), and emphasizing that Docker users need to run `docker-compose exec app rake db:create` after `docker-compose up`.

cpretzer:
Nice work on this @geppy!

julianguyen:
@geppy Yeah let's update the docs :D Thanks so much for working on this! I'm sorry for the late response, been busy starting a new job.

cpretzer:
I made an update to the docs here: https://github.com/ifmeorg/ifme/wiki/Installation

julianguyen:
Thanks so much @cpretzer :D

PR #1224, opened 2018-10-30
Refactor routes to use only instead of except

arun1595:
### Description

This is a minor refactor PR which replaces instances of `except` keys with `only`keys  in the route definitions to keep them simple and easy to follow.

Also fixes a minor issue in the `notifications` route definition where `index` and `destroy` routes were included but we didn't have an `index` action inside `NotificationsController`.

Do let me know if I missed something.

Thanks!


hmasila:
This is nice. I am also a fan of `only` so yeeah 🎉

config/routes.rb
```diff
@@ -61,15 +61,15 @@
...
 
-  resources :search, except: [:show, :new, :create, :edit, :update, :destroy] do
+  resources :search, only: :index do
     collection do
       get 'posts'
     end
   end
 
-  resources :notifications, except: [:show, :new, :create, :edit, :update] do
+  resources :notifications, only: :destroy do
```
ludamillion: The NotificationsController does use the index route as a fallback for the redirect_back when rendering an HTML request.

```ruby
  def destroy
    @notification.destroy if @notification.present?

    respond_to do |format|
      format.html { redirect_back(fallback_location: notifications_path) }
    end
  end
```
As such this should be 
```suggestion
  resources :notifications, only: [:destroy, :index] do
```

arun1595: @ludamillion Ah I see. Thanks for letting me know.

Let me update my PR.

PR #1229, opened 2018-10-31
Refactor meeting controller show specs

andrezacm:
# Description 
- Removing duplicated set up of meeting on `MeetingsController#show`
- Refactor meetings controller show specs and add missing behavior

## More Details
There is a before action in MeetingsController that sets up `@meeting` for the show action, although inside show action `@meeting` was being set up again.

I've noticed when a user has not joined a meeting yet and he is a member of the meeting's group,  he has access to the meeting's page but can't comment on it or see other comments. This behavior wasn't present on show specs.

# Test Coverage

✅


app/controllers/meetings_controller.rb
```diff
@@ -6,7 +6,6 @@ class MeetingsController < ApplicationController
 
   # GET /meetings/1
   def show
-    @meeting = Meeting.friendly.find(params[:id])
```
julianguyen: Oooh, nice catch!

PR #1240, opened 2018-11-02
French Translations

julianguyen:
# Description 

Combined efforts for French translations cc: @CCoupris @sylvain-ch21

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/776

# Test Coverage

✅


config/locales/devise.fr.yml
```diff
@@ -0,0 +1,60 @@
+fr:
+  devise:
```
julianguyen: @CCoupris or @sylvain-ch21 could you translate this file into French? :)

config/locales/devise.fr.yml
```diff
@@ -0,0 +1,60 @@
+fr:
```
julianguyen: Taken from https://github.com/tigrish/devise-i18n/blob/master/rails/locales/fr.yml

Issue #14, opened 2015-02-27
Reporting abuse and harassment, blocking allies


julianguyen:
There needs to be a feature in place for doing so and a moderation system for that

# Requirements

- Being able to report a user by clicking a link from their profile or from a comment
- When that happens, that data is saved to a new table in the database (include the reportee user id and reported user id and an optional comments column)
- You would have to create the ability to make a User an admin account, I would use column as a flag in the db for that. 
- Any “admin” user and can go in and open a panel and block users
  - Admin users also see the emails of the users, so that they can email them for any follow-up
  - Make the link to this dashboard obvious in the header. I suggest maybe a smaller header above the current one that appears just for admin users
- Blocking users entails not being able to sign into their account and getting an email and message on the site that says their account has been blocked for violating Terms and Conditions (that is currently being worked on but mention it anyways :P)
- We don’t need to delete the account, so again set a column as a flag for the "block" state

For now, the only way you can add someone as an admin is through rails console. We don't need a UI for that. Definitely add documentation on how to do that in the README.

estermer:
Is this issue still something you need help on? I can maybe try and tackle this

julianguyen:
@estermer Yes that would be great! You can work on this as a more long term project :) We should talk more about this feature on Slack!

estermer:
ok cool

cartothemax:
@estermer  Has any work been done on this issue?  I was browsing through issues and noticed this had been sitting for a bit and thought I could help out with it?  I don't want to step on any toes though.  Thanks!

estermer:
@cartothemax I got busy with work. You are more than welcome to tackle this. Thanks.

cartothemax:
No worries @estermer.  I have had the same challenges myself.  @julianguyen would it be possible to assign this to me?  I am going to finish up https://github.com/julianguyen/ifme/issues/442 and then get this going as well.  Thanks!

jarmstrng:
As this is a good-sized feature, I created a Project for it [here](https://github.com/julianguyen/ifme/projects/8?)

eiguike:
Hello! I would like to help in this feature, how can I accomplish this?

andy5995:
Hi @eiguike ! Thanks for your interest. If you send an email to join.ifme@gmail.com you will receive the details on how to join our _Slack_ development chat room. :)

alvarocasadoc:
Is there anyone still working on this? I think I can help getting it done :)

[...]

PR #1241, opened 2018-11-03
Finishing up reporting feature

julianguyen:
# Description 

Finishing up reporting feature which @aSquare14 and @prateksha worked on for Rails Girls Summer of Code 2018 🎉 

### What I've done

- [x] Add missing test coverage
- [x] Send banned user email notifying whether they've been banned or unbanned
- [x] Cleanup UI
- [x] Fix linting
- [x] Fix translation bugs
- [x] Cleanup ReportsController
- [x] Re-add support for Comments
- [x] Ran `annotate` since there were schema changes

### In another PR
- Notify admin user when new report is submitted

## Additional Details

Do `Rebase and Merge` once pull request is ready to merge

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/14

# Screenshots

capture

image

image

image

# Test Coverage

✅


config/locales/es.yml
```diff
@@ -336,6 +353,14 @@ es:
...
       join_meeting_subject: 'if me | %{user} se ha unido a "%{meeting}" de "%{group}"'
       join_meeting_body: '<p>Para ver "%{meeting}", %{link}!</p>'
+      reported_user_subject: 'if me | Usted ha informado de un usuario'
+      reported_user_body: 'Usted ha informado %{reported_name}. Esto será revisado por un administrador.'
+      reportee_user_subject: 'if me | Has sido reportado'
+      reportee_user_body: 'Has sido reportado. Esto será revisado por un administrador.'
+      add_ban_subject: 'if me | Has sido prohibido'
+      add_ban_body: 'Has sido prohibido por violar a nuestra %{code_of_conduct_link}.'
+      remove_ban_subject: 'if me | Ya no estas prohibido'
+      remove_ban_body: 'Usted fue previamente prohibido por violar a nuestra %{code_of_conduct_link}. Ya no estás prohibido.'
```
alvarocasadoc: I would rephrase this to something like "Anteriormente fuiste baneado/a por violar nuestro %{code_of_conduct_link}. Ya no estás baneado/a" or "El ban aplicado sobre tu cuenta por violar nuestro %{code_of_conduct_link} ha sido cancelado".

app/controllers/profile_controller.rb
```diff
@@ -11,4 +11,45 @@ def index
...
+    return unless user.any?
+
+    if banned
+      BannedMailer.add_ban_email(user.first).deliver_now
+    else
+      BannedMailer.remove_ban_email(user.first).deliver_now
+    end
+  end
+
+  def ban_or_remove(banned)
```
nshki: I wonder if there's a more expressive name we can give this method. The two wrappers above (`ban_user` and `remove_ban`) make this very clear, but as a standalone method, I had to do a double take. Calling `ban_or_remove(false)` is a little confusing, but that could also just be me.

Would something like `ban` be too generic?

app/controllers/profile_controller.rb
```diff
@@ -11,4 +11,45 @@ def index
     @stories = Kaminari.paginate_array(get_stories(user))
                        .page(params[:page])
   end
+
+  def ban_user
+    ban_or_remove(true)
+  end
+
+  def remove_ban
```
nshki: Would like to propose `unban_user` as the action name here. Or alternatively, renaming `ban_user` to `add_ban`.

app/views/profile/index.html.erb
```diff
@@ -36,6 +39,17 @@
         </div>
       <% end %>
     <% end %>
+    <% if current_user.admin? %>
+      <% if @profile.banned?%>
```
nshki: Missing space before `%>`.

julianguyen:
Woo, thank you for the reviews @alvarocasadoc and @nshki 🎉 Especially for improved Spanish translations!

app/controllers/profile_controller.rb
```diff
@@ -11,4 +11,45 @@ def index
...
+    ban(false)
+  end
+
+  private
+
+  def ban_mailer(user, banned)
+    return unless user.any?
```
arun1595: The argument `user`, contrary to what it says, actually holds an `ActiveRecord_Relation` object which is a bit confusing imo. What do you think of calling the `ban_mailer` method with the `User` object instead? This way, `ban_mailer` doesn't need to check if it was passed a non-empty collection and it doesn't need to get the user object with the `user.first` call.

The `ban` and `ban_mailer` methods would look something like

```ruby
def ban(banned)
  return unless current_user.admin
  user = User.find_by(id: params[:user_id])
  return unless user

  user.update(banned: banned)
  ban_mailer(user, banned)
  redirect_to(
    admin_dashboard_path,
    notice_or_alert(
      user,
      banned ? 'user_banned' : 'ban_removed'
    )
  )
end

def ban_mailer(user, banned)
  if banned
    BannedMailer.add_ban_email(user).deliver_now
  else
    BannedMailer.remove_ban_email(user).deliver_now
  end
end
```

And we could go even further and move all the ban and remove ban logic into a Service object.

Thoughts?

app/controllers/profile_controller.rb
```diff
@@ -11,4 +11,45 @@ def index
...
+
+    if banned
+      BannedMailer.add_ban_email(user.first).deliver_now
```
arun1595: Do we really need to pass the whole user object to the mailer method?

AFAIK, the arguments passed to the mailer method will be serialized and stored in the queue to be processed later. Serializing a heavy `User` object is expensive and we should avoid it . Someone let me know if I'm wrong.

I suggest we pass the `user_id` to the mailer method and let it do the job of looking up the `User` object.

@julianguyen  What do you think?

app/controllers/profile_controller.rb
julianguyen: This is a really good call out! Thank you. I think for now we can keep it in the controller, but I'll update things so that the user object is actually the expected type.

app/controllers/profile_controller.rb
julianguyen: That's a good suggestion, will update :D Thanks!

app/helpers/notification_mailer_helper.rb
```diff
@@ -20,7 +20,7 @@ def comment_on_strategy_subject(data)
 
   def comment_on_meeting_subject(data)
     group_id = Meeting.find(data['typeid']).group_id
-    group = Group.where(id: group_id).first.name
+    group = Group.find_by(id: group_id).name
```
hmasila: 👍

app/helpers/notification_mailer_helper.rb
```diff
@@ -51,10 +51,10 @@ def new_group_body(data)
       val('new_group_body'),
       user: data['user'].to_s,
       group: data['group'].to_s,
-      description: Group.where(id: data['group_id']).first.description,
+      description: Group.find_by(id: data['group_id']).description,
```
hmasila: 👍 I hope there will always be a group with this id `data['group_id']` otherwise we will have to deal with `no method error description for nil class`

julianguyen: There should be, but I can double check and fix it in another PR (since it's technically out of scope). Good call out, thanks :)

Issue #1215, opened 2018-10-27
Requesting Japanese Translation


toryalsip:
# Description 
My wife is from Japan and would really love to have this.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

julianguyen:
That would be so awesome to have Japanese translations! Thank you @matcha440!

nshki:
Oh, this is super exciting. If you could keep me looped into subsequent PRs, I would love that! I speak fluent Japanese as well and would love to be an extra pair of eyes where needed.

matcha440:
@nshki that sounds like a great idea to me.  Although I've made a fair amount of progress, I'm sure there are a lot of mistakes too and it is quite a daunting effort.  I'll try and see what I can do to make reviewing what I've got more digestible, maybe take it a section at a time.  Expect to see some PRs start probably later today.

snailscode:
Hi! I'd love to help if this issue still needs Japanese translations! I've been speaking Japanese ever since I was a kid, and have experience in composing/translating my own projects from English > Japanese and vice versa. :)

matcha440:
@snailscode help is welcome and appreciated.  I started on this and made a fair amount of progress last year, but things in my life got hectic and I haven't had an opportunity to get back to it so the the branch I was working on is rather far behind the master branch at this point.  Maybe a good start would be to get the [ja-translations](https://github.com/ifmeorg/ifme/tree/ja-translations) branch rebased? 🤔

snailscode:
@matcha440 Thank you. I hope everything's all set and in order now for you now! 

Wasn't too sure what rebasing was, so I did a quick google search on it and stumbled upon this [article](https://git-scm.com/book/en/v2/Git-Branching-Rebasing)! Are you saying to do something like this? 
```
$ git checkout ja-translations
$ git rebase master
```
Sorry for the silly question! Still pretty new to everything. I'm also a bit lost on where to start in the project, too. 😰

julianguyen:
Hey @snailscode! Not a silly question! Thanks for helping out with this :) Yeah so this branch is on this repo. So you'll want to clone the repository, pull the latest from the master branch, checkout ja-translations, and rebase like you stated above!

[...]

PR #1243, opened 2018-11-04
Translations for ja.yml through line 175

matcha440:
# Description 
I think the first 175 lines here were for the most part fairly straightforward, but I'm sure there are some mistakes.  I put my wife's GitHub handle as co-author as she helped a little with some of these.

## Corresponding Issue

Relates to #1215 

# Test Coverage

✅

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -60,8 +60,8 @@ ja:
       delete: 消す
       edit: 編集する
       edit_instance: '編集%{instance}'
-      join: 登録
-      leave: 離れる
+      join: 参加する
+      leave: 立ち去る
       add: 追加する
       add_all: 全て追加する
       select_all: 全て選択する
@@ -93,17 +93,17 @@ ja:
   allies:
     add_ally: 仲間に加える
     cancel_ally_request: 'リクエストをキャンセルする'
-    accept: 受け取る
-    reject: 拒む
+    accept: 承認する
+    reject: 拒否する
     index:
       title: 仲間
-      outgoing: 出リクエスト
-      incoming: '入リクエスト'
+      outgoing: 申請したリクエスト
+      incoming: '受信したリクエスト'
       invited: '電子メールで招待されました。'
       invited_on: '招待された：%{date}'
       invitation_expired: '招待状の期限が切れました'
       invite: '大切な人を招待しましょう！'
-      none: '仲間を増えるために検索するか、または大切な人を招待することができます。'
+      none: '仲間を増やすために検索するか、または大切な人を招待することができます。'
       view_all_allies: 全ての仲間を見る
       cancel: キャンセル
     form:
@@ -114,12 +114,12 @@ ja:
     edit_category: カテゴリー編集
     new: 新しいカテゴリー
     form:
-      name_hint: '人、場所、物、アクティビティとかをカテゴライズします。'
+      name_hint: '人、場所、物、アクティビティなどをカテゴライズします。'
     label:
       one: 'カテゴリー：'
       other: 'カテゴリー：'
     index:
-      subtitle: '生活は大変なので、人や場所や物をカテゴリーしましょう。'
+      subtitle: '人生は大変なので、人や場所や物をカテゴライズしましょう。'
       premade1_name: 家族
       premade1_description: 家族に関係ある
       premade2_name: 友達
@@ -128,7 +128,7 @@ ja:
       premade3_description: 仕事に関係ある
       premade4_name: 瞑想
       premade4_description: '瞑想と深呼吸とリラックスのこと'
-      instructions: 'カスタムカテゴリーはまだ作りません。<br>後でカスタマイズしたら、この作り置きのカテゴリーを使えます。'
+      instructions: 'カスタム・カテゴリーはまだ作っていません。<br>後でカスタマイズしたら、この作り置きのカテゴリーを使えます。'
   errors:
     home_page: 'ホームページに戻る'
     internal_server_error:
@@ -154,10 +154,10 @@ ja:
       success: 'このグループに入りました。'
       group_not_exists: '存在しないグループが入れません。'
     leave:
-      error: 'グループのリーダーはあなただけから、離れらることができません。'
-      success: '%{group}を離れていました。'
+      error: 'グループのリーダーはあなただけから、立ち去ることができません。'
+      success: '%{group}を立ち去りました。'
       remove_member_success: '%{user}が%{group}から外しました。'
-      remove_member_error: 'リーダーじゃないからメンバーが外せません。'
+      remove_member_error: 'リーダーではないからメンバーが外せません。'
     new: 新しいグループ
     show:
       led_by: 'リーダー：'
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -1,18 +1,18 @@
 ja:
   app_name: 'if me'
-  app_description: 'A community for mental health experiences'
+  app_description: 'メンタルヘルスのコミュニティです'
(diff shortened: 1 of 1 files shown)


config/locales/ja.yml
```diff
@@ -27,152 +27,152 @@ ja:
...
+    faq: よくある質問
+    resources: リーソス
+    code_of_conduct: 行動規範
   common:
-    name: Name
-    you: You
-    title: Title
-    date: Date
-    daily_reminder: 'Daily reminder email'
+    name: 名前
```
matcha440: I noticed when I was previewing this locally that the usage isn't quite what I thought it was.  It's not the name of a person, so 名前 may not be suitable, but I don't know what would fit better here.

julianguyen:
Could you rebase from master? There are some new translations added related to the report feature. Sorry about that!

matcha440:
@julianguyen yeah, I could rebase but I think that would be somewhat disruptive to the flow.  Would it be better if I just merge master into ja-translations (and fix conflicts) and from ja-translations to the branch related to this review?  I'd personally prefer to do any rebasing once we're near the end and I can just tidy up all commits prior to merging into master at that time.

julianguyen:
@matcha440 That works! :)

config/locales/ja.yml
```diff
@@ -27,152 +27,152 @@ ja:
...
-      premade1_name: Family
-      premade1_description: 'Family related'
-      premade2_name: Friends
-      premade2_description: 'Friends related'
-      premade3_name: Work
-      premade3_description: 'Work related'
-      premade4_name: Meditation
-      premade4_description: 'Meditation, deep breathing, and relaxing activities'
-      instructions: 'You haven''t created any custom categories yet.<br>You can also add the following premade categories and customize them another time.'
+      subtitle: '人生は大変なので、人や場所や物をカテゴライズしましょう。'
```
nshki: Hmm, 人生は大変なので is a good translation here, but the whole phrase doesn't have the same ring in Japanese as it does in English. Perhaps just omitting that portion would work better for Japanese.

config/locales/ja.yml
```diff
@@ -27,152 +27,152 @@ ja:
     fr: Français
     ja: 日本語
   navigation:
-    about: About
-    blog: Blog
-    contribute: Contribute
-    partners: Partners
-    privacy: 'Privacy Policy'
+    about: 'プロフィール'
```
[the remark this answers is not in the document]
matcha440: Yeah, and that's what my wife suggested at first too, and then I saw this string is actually referenced in two different places.  One is of course the main navigation bar so "if meにつて" does make sense there but the other is in the section for modifying the Account.  It's about like we have to have two entries to handle the different contexts here but that's going to be a bit more effort.  If we don't have something that makes sense for both then that's what we may have to end up doing though, right?

matcha440:
@nshki thank you for the initial feedback.  I'll get these updated a little later today but had a question on one of them.

config/locales/ja.yml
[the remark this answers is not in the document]
nshki: Ah gotcha. Yeah, unfortunately if we can't find something that'll work in both contexts, we'll have to split this.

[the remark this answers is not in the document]
matcha440: I've put it back and I think it may make sense to open up a separate issue to address splitting it.

matcha440:
Merging things in from master added some new entries, but I'll save those for later to keep the scope of this PR focused.

config/locales/ja.yml
```diff
@@ -27,55 +27,55 @@ ja:
...
+      confirm: 'よろしいですか？'
+      delete: 消す
+      edit: 編集する
+      edit_instance: '編集%{instance}'
+      join: 参加する
+      leave: 立ち去る
+      add: 追加する
+      add_all: 全て追加する
+      select_all: 全て選択する
+      remove: 外す
```
nshki: 消す I think is more commonly used, but I might be wrong.

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
-      premade2_name: Friends
-      premade2_description: 'Friends related'
-      premade3_name: Work
-      premade3_description: 'Work related'
-      premade4_name: Meditation
-      premade4_description: 'Meditation, deep breathing, and relaxing activities'
-      instructions: 'You haven''t created any custom categories yet.<br>You can also add the following premade categories and customize them another time.'
+      subtitle: '人や場所や物をカテゴライズしましょう。'
+      premade1_name: 家族
+      premade1_description: 家族に関係ある
```
nshki: 家族関係, perhaps?

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
+      premade2_name: 友達
+      premade2_description: 友達に関係ある
```
nshki: 友達関係

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
+      premade3_name: 仕事
+      premade3_description: 仕事に関係ある
```
nshki: 仕事関係

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
+      premade4_name: 瞑想
+      premade4_description: '瞑想と深呼吸とリラックスのこと'
```
nshki: 瞑想、深呼吸、とリラックス関係

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
+    singular: グループ
+    plural: グループ
     form:
-      leaders: Leaders
-      only_leader: 'You are the only leader, therefore you can''t leave.'
-      cannot_delete: 'Please note that you cannot delete groups.'
-      error_save: 'We can''t save this group because: '
-      error_edit_permission: 'You must be a leader of a group in order to edit it'
+      leaders: リーダー
+      only_leader: 'リーダーはあなたしかないから、立ち去ることができません。'
```
nshki: リーダーはあなたしかいませんので、立ち去ることができません。

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
+      cannot_delete: 'グループを削除することができません。'
+      error_save: 'グループを保存できない理由は：'
+      error_edit_permission: 'グループのリーダーじゃないから編集できません。'
```
nshki: あなたはグループリーダーではありませんので編集できません。

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
-      group_not_exists: 'You cannot join a group that doesn''t exist.'
+      success: 'このグループに入りました。'
+      group_not_exists: '存在しないグループが入れません。'
     leave:
-      error: 'You cannot leave the group, you are the only leader.'
-      success: 'You have left %{group}'
-      remove_member_success: 'You have removed %{user} from %{group}'
-      remove_member_error: 'You cannot kick a group member because you are not a leader.'
-    new: 'New Group'
+      error: 'グループのリーダーはあなただけから、立ち去ることができません。'
```
nshki: グループリーダーはあなたしかいませんので、立ち去ることができません。

config/locales/ja.yml
```diff
@@ -93,103 +93,103 @@ ja:
...
+      success: '%{group}を立ち去りました。'
+      remove_member_success: '%{user}が%{group}から外しました。'
+      remove_member_error: 'リーダーではないからメンバーが外せません。'
```
nshki: あなたはリーダーではありませんので、メンバーを外せません。

Issue #1168, opened 2018-10-15
Alert existing users about the new password criteria


yasirazgar:
# Description 
Continuation of #1153 
Show modal dialogue on login for existing user to update there password, for every six months, 
For this store a date/time for when the user has last changed password.
Need a new datetime field for this, by default this should be set to updated_at time when creating for the existing user and created_at/Time.now time while creating a new user.
If possible also add an feature for checking previous three password of the user against the new password.

# Reproduction Steps

# Screenshots

# Do you want to be the assignee to work on this?

✅ 
🚫

yasirazgar:
Will try to work and complete it by this weekend.

PR #1249, opened 2018-11-06
1168 - Password change alert

yasirazgar:
# Description 

Added additional validation for password, which disallows previous three password.
Added modal dialogue to remind user to update their password, for every six months.
Pulled ally related methods into separate concern, to fix code climate ClassLength warning.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1168

# Screenshots
Password Modal
password_modal
Used password error
used password error

# Test Coverage

✅


config/locales/en.yml
```diff
@@ -405,6 +405,10 @@ en:
       email_notifications_question: 'How can I modify email notifications?'
       email_notifications_answer: 'Go to the <strong>%{accounts}</strong> page. You can modify your email, password, profile picture, and bio here. You can also delete your account. Once you delete your account, all your data is deleted.'
     home:
+      update_pwd_modal:
+        title: 'Update password'
+        body: 'We have enforced better security measures for our users, please %{update_password_link}'
+        update_link: 'update your password.'
       stories: Stories
```
yasirazgar: @julianguyen, is the text ok, as this will be shown for every six months.

app/models/concerns/password_validator.rb
```diff
@@ -2,18 +2,62 @@
...
       /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[#?!@$%^&*-]).*$/
-    google_oauth2_enabled? || password.blank? || (password =~ password_regex)
+    return true if password =~ password_regex
+
+    errors.add(:password, I18n.t('devise.registrations.password_errors.format'))
+    false
+  end
+
+  def not_in_used_passwords?
+    password_histories.pluck(:encrypted_password).each do |ep|
```
julianguyen: Just for clarity, let's rename `ep` to `password`

app/views/pages/_update_password_modal.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if user_signed_in? && current_user.password_needs_update? %>
+  <%= react_component('Modal', props: {
+    title: t('pages.home.update_pwd_modal.title'),
+    body: t('pages.home.update_pwd_modal.body', :update_password_link => link_to(t('pages.home.update_pwd_modal.update_link'), edit_user_registration_path)),
```
julianguyen: Let's use colon notion instead of arrow notation i.e. `update_password_link: link_to(t('pages.home.update_pwd_modal.update_link'), edit_user_registration_path))`

config/locales/en.yml
julianguyen: Hmmmm I think there's also security issues with changing one's password periodically. I think it would be better to just ask the a user who hasn't had a history at all of changing their password to do so.

julianguyen:
Any updates @yasirazgar ? :)

app/models/concerns/password_validator.rb
yasirazgar: I think using password will conflict with already existing password attr in user, will use old_password.

config/locales/fr.yml
```diff
@@ -405,6 +405,10 @@ fr:
       email_notifications_question: 'Comment puis-je modifier les alertes email?'
       email_notifications_answer: 'Rendez-vous sur la page <strong>%{accounts}</strong>. Vous pouvez modifier votre email, mot de passe, photo de profil, et bio ici. Vous pouvez aussi supprimer votre compte. Une fois votre compte supprimé, toutes vos données sont supprimées.'
     home:
+      update_pwd_modal:
+        title: 'Mettre à jour le mot de passe'
```
mohammedyasir:
```
Psych::SyntaxError: (<unknown>): did not find expected key while parsing a block mapping at line 409 column 9
```
@julianguyen I am getting above error while running `rake react_on_rails:locale`.
I think that Issue is in this line, not sure what I am missing.

julianguyen: So the issue is actually with the next line:
```
'Nous avons mis en place de meilleures mesures de sécurité pour nos utilisateurs, s'il vous plaît %{update_password_link}'
```

The `'` in `s'l` is unescaped. You have to do:
```
'Nous avons mis en place de meilleures mesures de sécurité pour nos utilisateurs, s''il vous plaît %{update_password_link}'
```

Ahh sorry, this is hard the debug :S

yasirazgar:
@julianguyen fixed locales and addressed your comments.

app/models/concerns/password_validator.rb
julianguyen: What about using encrypted_password instead?

app/views/pages/_update_password_modal.html.erb
julianguyen: ^ Please change this :)

config/locales/en.yml
julianguyen: Let's increase this to 12 months!

app/models/concerns/password_validator.rb
yasirazgar: There is a filed with same name in the user model, even though the encrypted_password will be local to the loop, I think will be good if avoid it, what do you think.

app/views/pages/_update_password_modal.html.erb
yasirazgar: Oops forgot this one will update it.

app/models/concerns/password_validator.rb
yasirazgar: Updated it to encrypted_password.

yasirazgar:
@julianguyen, Addressed the latest comments.

julianguyen:
Not sure if you saw already, but it looks like there's a [failing test](https://circleci.com/gh/ifmeorg/ifme/6053?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

app/models/password_history.rb
```diff
@@ -0,0 +1,15 @@
...
+# Table name: password_histories
+#
+#  id                     :integer          not null, primary key
+#  user_id                :integer           default(''), not null
+#  encrypted_password     :string           default(''), not null
+
+class PasswordHistory < ApplicationRecord
+  belongs_to :user
+
+  validates :encrypted_password, presence: true, uniqueness: true
```
julianguyen: There should be a basic unit test for this new model

julianguyen: @yasirazgar I don't think you committed the `password_histories_spec`

yasirazgar: Oops, my bad committed now.

PR #1252, opened 2018-11-07
Add about label to devise.registrations.edit.labels

### LINKED ISSUE #1251: Navigation and Account share the same About text
# Description 
While working on Japanese translations, I found that the text that is used for the navigation menu and the Account edit page both reference the text `navigation.about`.  For most languages this is probably not a big deal, but for others this produces some nonsensical results. (see screenshots below)

My suggestion is to make this a little more granular by giving Account its own About text.

# Screenshots
image

image

✅

matcha440:
# Description 
Adds a new text string so that the Account Edit page can have its own `About` text.

## Corresponding Issue
Fixes #1251 

# Test Coverage

✅

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -655,6 +655,7 @@ fr:
         labels:
           profile_picture: 'Photo de profile'
           remove_picture: 'Effacer la photo'
+          about: 'À notre propos'
           password_for: 'Mot de passe pour votre compte <strong>if me</strong>'
           update: 'Mise à jour'
         placeholders:
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -41,7 +41,7 @@
     <% end %>
   <% end %>
 
-  <div class="label smallMarginTop"><%= t('navigation.about') %></div>
+  <div class="label smallMarginTop"><%= t('devise.registrations.edit.labels.about') %></div>
   <%= f.text_area :about, placeholder: t('devise.registrations.edit.placeholders.about'), 'aria-label': t('devise.registrations.edit.placeholders.about') %>
 
   <div class="label smallMarginTop">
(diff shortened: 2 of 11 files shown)


config/locales/fr.yml
```diff
@@ -655,6 +655,7 @@ fr:
         labels:
           profile_picture: 'Photo de profile'
           remove_picture: 'Effacer la photo'
+          about: 'À notre propos'
```
ludamillion: I think these locales should have the attention of a fluent speaker. For example _À notre propos_ translates as _About us_. Unless I'm mistaken this should be simply _À propos_ or possibly _À moi propos_. I'm not a native French speaker. (I believe this is also a problem with _Über uns_ as well).

Clearly this is an existing issue but excepting the rare cases where someone's preferred pronoun is plural I don't think that the label on the account edit page should be _About us_.

julianguyen: It looks like we use this translation elsewhere: https://github.com/ifmeorg/ifme/blob/ff6c1bca9859ef575abe641cc34ea6c32bc4ba67/config/locales/fr.yml#L29, which was originally translated by French-speaking folks. I know French myself, and this translation is appropriate since it's an about page on our project.

ludamillion: Clearly the translation is correct when appears in the navigation as it pertains to the project. I'm not suggesting the original usage there was incorrect. What I meant is that the translation which appears on the user's account edit page is still _À notre propos_. It seems strange to me to have the about section on the account edit page labeled _About Us_. Again I'm no native speaker but that strikes me as incorrect. Perhaps it is an idiomatic usage I'm not aware of.

For more context here is a screenshot of what I'm referring to.
screenshot_2018-11-10 if me - compte

julianguyen: Aha, you're actually right about the above translation! Sorry for prematurely merging

PR #1259, opened 2018-11-10
* Resolves #1258: Updates Dockerfile to use Alpine Linux

### LINKED ISSUE #1258: Use Alpine Linux for Smaller Docker Image
# Description 
@geppy did great work to improve the Docker experience and to continue that work, I'm changing the Dockerfile to use [Alpine Linux](https://alpinelinux.org/) for a smaller image.

Also, there are some additional steps to initialize the environment to run in Docker which I'll commit to the documentation.

✅

cpretzer:
# Description 
This commit adds Alpine Linux as the base image for the ifme Docker image

## More Details
[Alpine Linux](https://alpinelinux.org/) is a slimmed down version of Linux that is good for running lightweight containers.

I observed that the Ubuntu image was roughly 2GB, and the Alpine Linux version is roughly 1.26GB, so there is a good amount of space savings there.

For the purposes of running the ifme app in a container locally, this is ideal because the Docker vm can quickly run out of space.

There are also some updates to the wiki that I'll add.
## Corresponding Issue
#1258


Dockerfile
```diff
@@ -1,23 +1,23 @@
 # Dockerfile
-FROM ruby:2.3
+FROM ruby:2.3-alpine
```
cpretzer: This is the main change to use Alpine Linux

Dockerfile
```diff
@@ -1,23 +1,23 @@
...
 
 ENV PORT 3000
 ENV PATH /node_modules/.bin:$PATH
 
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - \
-  && apt-get update -qq \
-  && apt-get install -y \
-    build-essential \
-    libpq-dev \
+RUN apk update \
```
cpretzer: Alpine Linux uses a different package manager called apk, so all the apt-get commands must be changed

Dockerfile
```diff
@@ -1,23 +1,23 @@
...
+  && apk add \
     nodejs \
+    curl \
+    npm \
+    build-base \
```
cpretzer: Need to add some additional packages that are not installed by default

Issue #848, opened 2018-03-10
[Component] Pagination Bullets


nshki:
## Overview

We are looking for an awesome human to build this out as a React component, preferably within Storybook. You can view and get details (colors, dimensions, export SVGs, etc.) by visiting [the Figma designs](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618) and creating a free account.

This is part of the app redesign (#691).

## Screenshot from [the Figma document](https://www.figma.com/file/RTjPkO0nSuJcMqqWLjeVdEof/if-me?node-id=65%3A618):

![screenshot 2018-03-10 at 12 06 07 pm](https://user-images.githubusercontent.com/1121087/37244808-3af8c06a-245c-11e8-855e-4794943c9b04.png)

anGie44:
i'd like to claim this issue! how can i join & contribute? sent an email to the join.ifme

cartothemax:
Hey @anGie44! 👋 

Thanks for jumping on this. Be sure to take a look at the contributing doc if you haven't https://github.com/ifmeorg/ifme/blob/master/CONTRIBUTING.md

And I am guessing that someone will get back to you shortly regarding getting added to the Slack channel. Might just be a little delay 😃 

In the mean time though feel free to get going on this work and feel free to ask any questions on the issue. Hope that helps!

nshki:
@anGie44 Great, take it away! I echo what @cartothemax says as well! I believe @julianguyen will add you as a collaborator/member so you can be assigned this task soon. :) Feel free to start though!

prateksha:
@nshki Just wanted to clarify something. When we click/hover on Pagination 2 and 3 , should it change state to look like 1?

ludamillion:
Not sure just how important this still is but I took a swing at it. The PR is here => https://github.com/ifmeorg/ifme/pull/1261

julianguyen:
Closing this cause we got rid of pagination.

PR #1261, opened 2018-11-12
Add React component for pagination bullets

ludamillion:
# Description 

Does what it says on the tin. Adds a React component for pagination bullets.

There was no hover state specified in the Figma designs so I reused the hover states from the buttons which share the same color scheme. 

I'm new to React and mostly implemented this based on analyzing the other components and reading the docs.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/848

# Screenshots

screenshot_2018-11-12 storybook

screenshot_2018-11-12 storybook 1

# Test Coverage

✅


client/app/components/Bullet/Bullet.scss
```diff
@@ -0,0 +1,36 @@
...
+    line-height: 1.2;
+  }
+
+  &:hover {
+    opacity: 0.5;
+  }
+
+  &.normal {
+    color: $white;
+    background-color: rgba(255, 255, 255, 0.1);
```
julianguyen: You can use $white-10 instead :)

client/app/startup/registration.js
```diff
@@ -13,6 +13,7 @@ import { Header } from '../components/Header';
 import { HeaderProfile } from '../components/Header/HeaderProfile';
 import { Resource } from '../components/Resource';
 import { Tag } from '../components/Tag';
+import { Bullet } from '../components/Bullet';
```
julianguyen: You can remove this since the component isn't being used in any Rails views yet.

client/app/components/Bullet/index.jsx
```diff
@@ -0,0 +1,18 @@
...
+  label: string
+};
+
+export const Bullet = (props: Props) => {
+  const {
+    normal, label, active
+  } = props;
+  const labelClassNames = `bullet ${css.bullet} ${active ? css.active : ''} 
+  ${normal ? css.normal : ''}`;
+  return <div className={labelClassNames}><span>{label}</span></div>;
```
julianguyen: So we actually want to build this component using `<li>`, and style it accordingly depending on whether the parent is `<ul>` or `<ol>`.

julianguyen:
I thought about this more, and we actually don't need a React component for this. We just need to style our default CSS for li, ul, and ol.

ludamillion:
I'll take another swing at simply styling the existing pagination elements. It looks like the existing pagination bullets/links are not accounted for in the design docs so I'll take a design pass at that as well. The pagination in the design docs is against the dark background which does not work in the current use case.

ludamillion:
Here's a super quick treatment I threw together based on the other pagination.
quick-treatment

julianguyen:
Nice work with the pagination items! I think you can actually rename that existing React component you created as PaginationItem.

Then for the bullets, just override the default ul, ol, and li styles :)

Thanks for working on this :D

hmasila:
Woow, this is amazing 🎉

paschalidi:
@ludamillion thats an amazing work. I do not know if this is still in progress but if you have no time I could go on and start developing it again. Please let me know how you feel about it. @julianguyen you may know the status of this?

julianguyen:
@paschalidi I don't know the status, unfortunately, but yeah good idea to check in and ask Luke!

Happy to help you look for another issue to start off with :P Let me know on Slack!

ludamillion:
@paschalidi Feel free to take over on this one. I started working on it just as I was starting a new job and haven't really been able to make time for working on ifme yet.

paschalidi:
@ludamillion @julianguyen will work on that only after I finish https://github.com/ifmeorg/ifme/issues/1390  & https://github.com/ifmeorg/ifme/pull/1391

PR #1273, opened 2018-11-20
Change refill column type

gr1d99:
# Description 
- Change `refill` column in `medications` table to `timestamp`
- This change will allow future implementations of complicated `reminders` as indicated in _TODO_

# Test Coverage
🚫

Tests are already written


app/helpers/medications_helper.rb
```diff
@@ -143,7 +143,7 @@ def medication_refill
     {
       type: 'date',
       label: t('medications.form.refill'),
-      value: @medication.refill || nil,
+      value: @medication.refill&.to_date || nil,
```
julianguyen: Will this always convert string date values to the Date object?

gr1d99: Nope, after the changing `refill` column, `refill` values are now retrieved as `DateTime` objects that are aware of `time zones`, therefore, `.to_date` will convert it to `Date` object. This will allow it to be rendered by form `input type='date'`.

In future when there is an iteration to allow users add date and time of refill, we will just get rid of `.to_date`, update `input type='date'` to `input type='datetime'`.

Issue #991, opened 2018-09-22
Fix full screen modal bug in mobile web


julianguyen:
# Description 

In mobile Chrome, there's a bug with our modal where the address bar is blocking the top part of the modal. Let's fix this!

This [post](https://teamtreehouse.com/community/position-fixed-css-bug-in-chrome-and-firefox-for-android) may give you some insights on how to fix it.

# Reproduction Steps

1. Open mobile Chrome
2. Open any modal (Notifications for example)

# Screenshots

![image](https://user-images.githubusercontent.com/3010728/45913290-810b6180-bde4-11e8-9ce9-429ae6c74b5b.png)

# Do you want to be the assignee to work on this?

🚫

mialahmoka:
I'd like to work on this

julianguyen:
Hi @mialahmoka :) Thanks for expressing your interest, we would love for your help. Here are next steps for contributing: https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack

julianguyen:
@vishalsodani Yep, once you contact us and we add you to our GitHub org, you can self-assign.

lsurasani:
Hey! Unsure if @vishalsodani is working on it but I'd like to take a look if not!

vishalsodani:
> Hey! Unsure if @vishalsodani is working on it but I'd like to take a look if not!
@lsurasani 
Please go ahead. I am not taking up this issue.

julianguyen:
@lsurasani Oh no sorry to hear. Did you try [remote debugging](https://developers.google.com/web/tools/chrome-devtools/remote-debugging/) via Chrome by any chance?

lsurasani:
I did! But it wasn't giving me any errors -- for whatever reason the stylesheets/css just weren't loading

julianguyen:
@CSumm The screen resolution is 1080x2160 and it's on the latest version of Android.

CSumm:
I'll see if I can recreate it on my own first with chrome and if not I'll definitely ask for the Browserstack info.

PR #1277, opened 2018-11-22
Modal no longer hidden behind url searchbar on Google Chrome mobile browser

CSumm:
# Description 

Fix the issue of pop up modal appearing behind url searchbar on Google Chrome mobile browser by reducing the size of the viewport height from 100 to 80.

## Corresponding Issue

#991 

# Screenshots
screenmodal

# Test Coverage

✅

### DIFF
--- a/client/app/components/Modal/Modal.scss
+++ b/client/app/components/Modal/Modal.scss
@@ -33,8 +33,10 @@
     height: 60vh;
     z-index: $z-index-front;
     @media screen and (max-width: $small) {
+      position: absolute;
       width: 100vw;
-      height: 100vh;
+      height: 80vh;
+      margin: 0 auto;
       border-radius: 0;
     }
     &Header {


client/app/components/Modal/Modal.scss
```diff
@@ -33,8 +33,10 @@
     height: 60vh;
     z-index: $z-index-front;
     @media screen and (max-width: $small) {
+      position: absolute;
       width: 100vw;
-      height: 100vh;
+      height: 80vh;
```
julianguyen: So we still want our modal to take up the full-screen height. Could you try 100% instead and see how that works?

CSumm:
I could definitely try that and see if it'll work.

client/app/components/Modal/Modal.scss
```diff
@@ -33,8 +33,10 @@
     height: 60vh;
     z-index: $z-index-front;
     @media screen and (max-width: $small) {
+      position: absolute;
```
julianguyen: Actually, I just saw this... Is this necessary?

CSumm: I think it was there to begin with. I can check to see if it's actually necessary or not right now.

julianguyen: Looks like it's been added. Yeah, let's check if that and `margin: 0 auto` is necessary! If not let's remove it!

client/app/components/Modal/Modal.scss
```diff
@@ -34,7 +34,8 @@
     z-index: $z-index-front;
     @media screen and (max-width: $small) {
       width: 100vw;
-      height: 100vh;
+      height: 100%;
+      margin: 0 auto;
```
julianguyen: Is this line necessary?

client/app/components/Modal/Modal.scss
CSumm: I removed both margin: 0 auto and position: absolute. It works without those. I must have slid them in thinking they would be useful when they aren't.

client/app/components/Modal/Modal.scss
CSumm: Margin: 0 auto has been removed as well.

PR #1280, opened 2018-11-23
1151/ Updated InputSelect component

alexmanzo:
# Description 

I updated the InputSelect component to accept an optional aria prop, which is then inserted into the input as an aria-label.

Then, I updated ToggleLocale widget to include the aria-label.

## Corresponding Issue

[Issue #1151 - Missing label for language dropdown on homepage](https://github.com/ifmeorg/ifme/issues/1151)

# Screenshots
 
UI unchanged

# Test Coverage
✅ - Added aria as prop in InputSelect test. All tests passed locally.


client/app/components/Input/InputSelect.jsx
```diff
@@ -9,6 +9,7 @@ import type { Option } from './utils';
 export type Props = {
   id: string,
   name?: string,
+  aria?: any,
```
julianguyen: Let's rename this to ariaLabel since aria has lots of properties.

client/app/components/Input/index.jsx
```diff
@@ -159,13 +159,14 @@ export class Input extends React.Component<Props, State> {
 
   displaySelect = () => {
     const {
-      options, name, id, value, onChange, type,
+      options, name, id, aria, value, onChange, type,
```
julianguyen: You'll need to add this new prop to the Props type in this file.

client/app/components/Input/InputSelect.jsx
julianguyen: Let's also add ariaLabel to `inputSelectProps` in `client/app/mocks/InputMocks.jsx` so that we're testing for it as well.

alexmanzo:
Hi Julia! Thanks for the notes. I've mode those changes and resolved the linting error.

client/app/components/Input/utils.js
```diff
@@ -63,6 +63,7 @@ export type Props = {
   small?: boolean,
   large?: boolean,
   value?: any,
+  ariaLabel?: any,
```
julianguyen: Let's make this a string

client/app/components/Input/InputSelect.jsx
```diff
@@ -33,14 +34,22 @@ export class InputSelect extends React.Component<Props, State> {
...
     return (
       <div className={css.select}>
         <div className={css.selectIcon}>
           <FontAwesomeIcon icon={faCaretDown} />
         </div>
-        <select id={id} name={name} value={value} onChange={this.toggleValue}>
+        <select
+          id={id}
+          name={name}
+          aria-label={ariaLabel}
```
julianguyen: Let's do `aria-label={label || ariaLabel}` and assume a select input field with a label can also be used as the aria-label.
And then add label to line 38.

client/app/components/Input/InputSelect.jsx
```diff
@@ -9,6 +9,7 @@ import type { Option } from './utils';
 export type Props = {
   id: string,
   name?: string,
+  ariaLabel?: any,
```
julianguyen: Let's make this a string

alexmanzo:
Thanks! I've made those updates. Let me know if you noticed anything else.

PR #1274, opened 2018-11-20
Adds Error label to inputs

Mera-Gangapersaud:
# Description 

The Moments Page now displays an error message below the input if it's left empty

## More Details

I added an Error Label Component for Inputs so the error message shows up below the textInputArea.

## Corresponding Issue

#1085 

# Screenshots

image

# Test Coverage

✅

Issue #1085, opened 2018-10-03
Add error message to Input components with validation errors


julianguyen:
# Description 

Currently, in for our forms that use React components (Input, Form), we only display validation errors with an exclamation mark icon. That is not the best experience and it would be better if we also had accompanying error text. For folks who use screen readers, this is a particularly bad experience.

Let's add that! Please put the error message in i18n and use react-intl from the client to pull out that text.

# Screenshots

## Currently
![image](https://user-images.githubusercontent.com/3010728/46429188-afe5d980-c6fa-11e8-8d96-4957a8b2674f.png)

## Desired
![image](https://user-images.githubusercontent.com/3010728/46429110-7b721d80-c6fa-11e8-8446-280c569693dc.png)

# Do you want to be the assignee to work on this?

🚫

Janice0218:
Can i take over it?

DamonClark:
Yea go ahead! I can find something else to work on tomorrow :)

julianguyen:
@Janice0218 Thanks for expressing your interest. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps. After that, assign yourself to the issue.

Mera-Gangapersaud:
Hey, @Janice0218  are you still working on this?

julianguyen:
Technically it's unassigned and our [rule](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) is that you have to assign yourself to the issue to be working on it @Mera-Gangapersaud

conkytom:
Hey @Mera-Gangapersaud, I don't mean to step on any toes, but I gave this a shot.  Found an error message already in the i18n files.  I could only get it added to the top of the field next to the exclamation, I couldn't figure out how to add it underneath. https://github.com/ifmeorg/ifme/pull/1250

Mera-Gangapersaud:
I looked at where the error message is added in the InputLabel component and its contained in a separate div from the text box. I'm not too sure, but I think the text box is the InputTextarea component. How would I go about adding a label below that without having to change the dynamic form?

PR #1282, opened 2018-11-24
Adds Error label to inputs

Mera-Gangapersaud:
# Description 

#1274 was closed by accident. This is the same branch. The Moments Page now displays an error message below the input if it's left empty.

## More Details

I added an Error Label Component for Inputs so the error message shows up below the textInputArea.

## Corresponding Issue

#1085 

# Screenshots

screenshot from 2018-11-22 09-45-40

# Test Coverage
✅

### DIFF
--- a/client/app/components/Input/InputError.jsx
+++ b/client/app/components/Input/InputError.jsx
@@ -1,39 +1,26 @@
 // @flow
 import React from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import {
-  faAsterisk,
-  faQuestion,
-  faExclamation,
-} from '@fortawesome/free-solid-svg-icons';
-import { Tooltip } from '../Tooltip';
+import { faExclamation } from '@fortawesome/free-solid-svg-icons';
 import css from './Input.scss';
 import globalCss from '../../styles/_global.scss';
 import { I18n } from '../../libs/i18n';
 
 export type Props = {
-  label: string,
-  required?: boolean,
-  info?: string,
   error?: boolean,
 };
 
-
 export const InputError = (props: Props) => {
-  const {
-    error, label, required, info,
-  } = props;
+  const { error } = props;
   return (
-    <div className={`${globalCss.gridRowSpaceBetween} ${css.label}`}>
     <div className={css.labelInfo}>
-    {error ? (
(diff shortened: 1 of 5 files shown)


client/app/components/Input/InputError.jsx
```diff
@@ -0,0 +1,26 @@
...
+import { I18n } from '../../libs/i18n';
+
+export type Props = {
+  error?: boolean,
+};
+
+export const InputError = (props: Props) => {
+  const { error } = props;
+  return (
+    <div className={css.labelInfo}>
```
julianguyen: Let's do this instead:
```
if (!error) return null;
return (
<div className={css.labelInfo}>
	<div className={`labelError ${globalCss.smallMarginTop} ${css.label}`}>
		<FontAwesomeIcon icon={faExclamation} />
		&emsp;
		{I18n.t('common.form.error_explanation')}
	</div>
</div>
);
```

client/app/components/Input/InputError.jsx
```diff
@@ -0,0 +1,26 @@
+// @flow
```
julianguyen: Let's write a unit test for this, `client/app/components/Input/__tests__/InputLabel.spec.jsx` is a good example to follow.

client/app/components/Input/index.jsx
```diff
@@ -236,6 +237,15 @@ export class Input extends React.Component<Props, State> {
     return null;
   };
 
+  displayError = () => {
+    const { type } = this.props;
+    const { error } = this.state;
+    if (REQUIRES_LABEL.includes(type)) {
```
julianguyen: I think we always want to display an error even if the component may not have a label.

julianguyen:
Any updates on this? :)

geppy:
@julianguyen Could we make a specific error message for each input?  `common.form.error_explanation` applies to the form as a whole.

julianguyen:
We could @geppy, but I think there's only one type of error i.e. empty value. Are we missing anything else though?

julianguyen:
@geppy That's a good point, and they should come from the server! Would you be interested in investigating this further? Thanks for bringing it up!

julianguyen:
Hey @Mera-Gangapersaud any updates on this?

Issue #1084, opened 2018-10-03
Can't tab out out of InputTextarea into the next input field


julianguyen:
# Description 

Can't tab out out of InputTextarea into the next input field

# Reproduction Steps

From /moments/new or /strategies/new, try to tab into the input fields instead of clicking into them. You'll notice that you can tab into the WYSIWYG editor, but you can't tab out of it.

# Screenshots

![capture](https://user-images.githubusercontent.com/3010728/46428762-adcf4b00-c6f9-11e8-8873-fdc672fab284.gif)

# Do you want to be the assignee to work on this?

🚫

shmesa22:
I can work on this one

julianguyen:
@shmesa22 Thanks for expressing your interest. Please check out https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack for next steps. After that, assign yourself to the issue.

arun1595:
I am going to assign myself to this issue as there hasn't been any response from @shelleyvadams

arun1595:
_Update_: I haven't been able to figure out the issue yet. Will try to do some reading on how `tabindex` works on non-interactive elements and see if it helps resolving this issue.

julianguyen:
@arun1595 Thnks for posting an update! You might have to set up a listener to watch for a tab key press in that element. Let us know on Slack if you need any more help :)

geppy:
@arun1595 This is caused by [a bug in Pell](https://github.com/jaredreich/pell/issues/66 "A11y Issue: Keyboard/Tab Trap").  It's [canceling 'Tab' `keydown` events](https://github.com/jaredreich/pell/blob/master/src/pell.js#L129-L130), which it shouldn't be doing.

We could fork Pell and fix that bug, or use another editor.

geppy:
I sent [a PR fixing this bug in pell](https://github.com/jaredreich/pell/pull/155).

julianguyen:
Woo awesome catch/fix @geppy!

arun1595:
Nice catch @geppy :tada:

LMulvey:
I've reached out to the developer of Pell to see if he'd allow other contributors to the project (I offered a bit of my time) so at the very least the outstanding PRs can be reviewed/merged. He seems to be based quite close to me so I'll meet up with him if need be, especially in light of the `event-stream` shenanigans last week.

jaredreich:
Check out v1.0.5 of pell, just published. Sorry for the delay everyone.

[...]

PR #1303, opened 2018-12-08
upgrade Pell (resolves ifmeorg/ifme#1084)

geppy:
# Description 

Upgrade [pell](https://github.com/jaredreich/pell/) to `^1.0.6`, which resolves the bug causing #1084.

## More Details

Behavior in pell 1.0.4 was causing #1084, and [my PR changing that behavior](https://github.com/jaredreich/pell/pull/155) has been merged.  I'm bumping the pell requirement to `^1.0.6` so we get a version with that change.

## Corresponding Issue

[#1084 <cite>Can't tab out out of InputTextarea into the next input field</cite>](https://github.com/ifmeorg/ifme/issues/1084)

# Screenshots

This wouldn't fit in a single-frame screenshot.

# Test Coverage

No tests are provided, since the change is in an upstream component.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -37,7 +37,7 @@
     "js-cookie": "^2.2.0",
     "jstimezonedetect": "^1.0.6",
     "location-autocomplete": "^1.2.4",
-    "pell": "^1.0.4",
+    "pell": "^1.0.6",
     "react": "^16.4.0",
     "react-autocomplete": "^1.8.1",
     "react-chartkick": "^0.1.3",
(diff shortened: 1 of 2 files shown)


geppy:
You can tab out of the editor in the forward direction (with <kbd>Tab</kbd>), but you can't escape the editor in the reverse direction (with <kbd>Shift</kbd>+<kbd>Tab</kbd>).

I think that issue's in the ifme codebase, but I'm not sure because React.

Issue #1302, opened 2018-12-07
Make locale tests dynamically generated


matcha440:
# Description 
Currently tests are statically created based off a known list of files that contain translations.  Instead, the tests could look at `config.i18n.available_locales` and use that to generate what test cases it needs to look at for each of the various files that are present.

matcha440:
Trying this out on my local, it _seems_ to work if I just take a quick glance on the page where I made the change.  But when I run the rspec tests things explode, so I would probably not recommend proceeding with this until the reason for those test failures is determined.

matcha440:
Actually I tried again just a moment ago and now the tests seem fine, so I'm not sure what I did differently.

matcha440:
I've split off some of the ideas I came up with into their own issues and reduced down the scope of this issue so that it is just about the work I did with the tests.

I've also updated the [translations](https://github.com/ifmeorg/ifme/wiki/Translations) section of the Wiki to reflect changes that were made in #1305.

FYI @julianguyen

PR #1305, opened 2018-12-09
Rewrite compare_locales_support_spec to be dynamic

matcha440:
# Description 
Derives tests for compare_locales_support based dynamically off of what English files are present.  Then for each locale that is defined:
1. Test for the existence of the locale file
2. Assert that the local file contains the same keys as what is in the English file.

## Corresponding Issue
Relates to #1302 

# Test Coverage

✅


spec/support/compare_locales_support.rb
```diff
@@ -1,48 +1,8 @@
...
-    kaminari_swedish: "#{LOCALES_DIR}/kaminari.sv.yml",
-    kaminari_norwegian: "#{LOCALES_DIR}/kaminari.nb.yml",
-    kaminari_dutch: "#{LOCALES_DIR}/kaminari.nl.yml",
-    kaminari_italian: "#{LOCALES_DIR}/kaminari.it.yml",
-    kaminari_vietnamese: "#{LOCALES_DIR}/kaminari.vi.yml",
-    kaminari_german: "#{LOCALES_DIR}/kaminari.de.yml",
-    kaminari_french: "#{LOCALES_DIR}/kaminari.fr.yml"
-  }.freeze
+  ENGLISH_FILES = Dir.glob("#{LOCALES_DIR}/**/*en.yml")
+  NON_ENGLISH_LOCALES = ['es', 'de', 'it', 'nb', 'nl', 'pt-BR', 'sv', 'vi', 'fr']
```
matcha440: I'm wondering if there is a way to reference the list of locales that is defined in in [config/application.rb](https://github.com/ifmeorg/ifme/blob/003785f3d72587f5ae6925d69e94a155b46f0b03/config/application.rb#L53), as I think that would be better but I've not yet figured out how to do that.

config/locale.rb
```diff
@@ -0,0 +1,9 @@
+module Locale
+  def Locale.available_locales()
+    # The available locales to be used by the app.  Unit tests are dynamically
+    # generated from this array.  If adding a new locale, be sure to refer to
+    # any failures when running rspec spec/compare_locales_support_spec.rb
+    # to see what keys and files are missing that need to be added.
+    ['en', 'es', 'de', 'it', 'nb', 'nl', 'pt-BR', 'sv', 'vi', 'fr']
```
julianguyen: Nice! We would also have to update our wiki [doc](https://github.com/ifmeorg/ifme/wiki/Translations#adding-new-locale-to-i18n) once merged!

spec/compare_locales_support_spec.rb
```diff
@@ -1,211 +1,25 @@
 # frozen_string_literal: true
 describe CompareLocalesSupport do
-  describe 'check non english locales include all keys' do
```
julianguyen: This is a great refactor 🙌

Issue #1183, opened 2018-10-17
Adding missing tests for MomentsHelper


julianguyen:
# Description 

Add missing tests! Use as an opportunity to refactor code (including getting rid of Rubocop disable comments like `# rubocop:disable MethodLength`)

# Do you want to be the assignee to work on this?

🚫

rhiroyuki:
Can I do this one @julianguyen ? :)

julianguyen:
@rhiroyuki Go ahead and assign yourself to the issue :D

julianguyen:
I recommend moving out new_moment_props and edit_moment_props to a new helper i.e. MomentFormHelper

gr1d99:
hi @julianguyen, can I pick this issue?

bmwachajr:
Hi @julianguyen, I have reached out to @gr1d99 and I will be taking on this issue.

PR #1366, opened 2019-02-15
Add missing tests and refactor MomentsHelper

bmwachajr:
# Description 
- Refactored the moments-helper method `moments_stats`
- Added new tests for the helper methods `new_moment_props` and `edit_moment_props`
- Refactor `moment_form_inputs`

## Corresponding Issue

fixes: #1183 

# Screenshots

# Test Coverage

✅


app/helpers/moments_helper.rb
```diff
@@ -146,24 +141,33 @@ def data_for(item)
...
+    {
+      total: current_user.moments.all.count,
+      monthly: current_user.moments.where(
+        created_at: Time.current.beginning_of_month..Time.current
+      ).count
+    }
+  end
+
+  def total_moment
+    unless moment_count[:total] == 1
```
hmasila: I think we can flip the `unless` here to be an `if`

app/helpers/moments_helper.rb
```diff
@@ -146,24 +141,33 @@ def data_for(item)
...
     end
 
-    t('stats.total_moment', count: total_count)
+    t('stats.total_moment', count: moment_count[:total])
   end
 
-  def monthly_moment(monthly_count)
-    unless monthly_count == 1
+  def monthly_moment
+    unless moment_count[:monthly] == 1
```
hmasila: I think we can flip the `unless` here to be an `if`

app/helpers/moments_helper.rb
```diff
@@ -21,22 +21,17 @@ def edit_moment_props(moment, viewers)
...
   end
 
-  # rubocop:disable MethodLength
   def moments_stats
-    total_count = current_user.moments.all.count
-    monthly_count = current_user.moments.where(
-      created_at: Time.current.beginning_of_month..Time.current
-    ).count
-    return '' if total_count <= 1
+    return '' if moment_count[:total] <= 1
```
bmwachajr: @hmasila I noticed this line returns `""` when a user has <= 1 moments. which makes Line_170 redundant. Cc @julianguyen

hmasila: @bmwachajr https://github.com/ifmeorg/ifme/pull/1366/files#diff-5f72a77ce1e4b1289dbf5eb071fb44f1R160 uses `moment_count[:total]` which is totally fine because it doesn't need this method

app/helpers/moments_helper.rb
```diff
@@ -146,24 +166,29 @@ def data_for(item)
...
+      t('stats.total_moments', count: moment_count[:total])
     end
+  end
 
-    t('stats.monthly_moment', count: monthly_count)
+  def monthly_moment
+    if moment_count[:monthly] == 1
+      t('stats.monthly_moment', count: moment_count[:monthly])
+    else
+      t('stats.monthly_moments', count: moment_count[:monthly])
```
julianguyen: You could do something like this:
```
moment_key = moment_count[:monthly] == 1 ? 'moment' : 'moments'
t("stats.monthly_#{moment_key}", count: moment_count[:monthly])
```

app/helpers/moments_helper.rb
```diff
@@ -146,24 +166,29 @@ def data_for(item)
...
+    }
   end
 
-  def monthly_moment(monthly_count)
-    unless monthly_count == 1
-      return t(
-        'stats.monthly_moments', count: monthly_count
-      )
+  def total_moment
+    if moment_count[:total] == 1
```
julianguyen: You could do something like this:
```
moment_key = moment_count[:total] == 1 ? 'moment' : 'moments'
t("stats.total_#{moment_key}", count: moment_count[:monthly])
```

app/helpers/moments_helper.rb
bmwachajr: This is so neat @julianguyen . I am definitely doing it 😄

bmwachajr:
@julianguyen this PR is good to go. Just waiting for your (re)approval 🙂

Issue #1024, opened 2018-10-01
Improve performance on production by reducing JS payloads


julianguyen:
# Description 

We've already done some significant work to improve performance (please note on production) including
- Code splitting in Webpack 4
- Inline rendering of CSS and Rails JS assets
- Deferring JS assets from React
- Using async on third party JS scripts
- Gzipping
- Minifying CSS in both Rails and React
- Caching

The Audit tool in Google developer console is super rad (you should check it out if you haven't already). Our performance score went from ~3 to ~40 on mobile and ~0 to ~36 on desktop on the signed in home page (Simulated Fast 3G, 4x CPU slowdown).

There is still significant work we can do with cutting down on reducing JS payloads. See screenshots below for more details.

# Screenshots

## Mobile
![screen shot 2018-09-26 at 6 46 10 pm](https://user-images.githubusercontent.com/3010728/46288015-4c607e00-c539-11e8-8e40-b5e1c48aac7a.png)

## Desktop
![screen shot 2018-09-26 at 6 56 30 pm](https://user-images.githubusercontent.com/3010728/46288016-4c607e00-c539-11e8-81fa-be817b413cc4.png)

## Payload Stats

![image](https://user-images.githubusercontent.com/3010728/46287976-30f57300-c539-11e8-96dc-e2ffe18f8d26.png)

![image](https://user-images.githubusercontent.com/3010728/46287986-36eb5400-c539-11e8-9a4e-4914669a2128.png)

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

julianguyen:
@geppy Just checking in, how is this going? :)

geppy:
I worked out where our JavaScript bytes are coming from, then planned out some straightforward changes that'd make a huge difference.

My day job's been too pointlessly dramatic for me to want to work on projects when I get home, but it should start calming down temporarily.  And I took off the rest of this week.

I'll try to work on this in PR-sized chunks, in this order:
- [ ] rewrite `Header` component
	- only user-visible change should be improved performance
	- taking out [`react-render-html`](https://www.npmjs.com/package/react-html-parser) might save 200+ kB (10%+ of the app's entire JavaScript payload)
[...]

Issue #726, opened 2017-10-17
Split off language-based assets into their own JavaScript files


baohouse:
# Description

Right now our main JavaScript file contains all language bundles. This isn't a huge performance problem right now, but we'll want to start thinking of splitting off the language bundles to load only what we need.

May have to think about how JavaScript is loaded so we don't run into dependency and load-order issues.

---
Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

baohouse:
This improvement could probably be done as part of #719

julianguyen:
We should definitely do this! Another neat optimization is only loading the translations needed for a page.

son-vyas:
@baohouse @julianguyen , I would like to give this a try. We you make me understand what it is to be done, since I am a new beginner

anushkajj:
Hi! I am a beginner and would like to make my first contribution. 
Is this issue still available?

PR #1368, opened 2019-02-18
Decrease chart.js bytes

geppy:
# Description 

Excluded data for unsupported locales, and removed an unused dependency.

## More Details
- Configured Webpack to exclude moment.js locale data for locales ifme doesn't support.
- Removed unused [`chartjs`](https://www.npmjs.com/package/chartjs) dependency (typo of [`chart.js`](https://www.npmjs.com/package/chart.js))

## Corresponding Issue
- part of #1024

### unlocalized charts
We aren't actually configuring the locale for charts: time axis labels are always in English (at least on `/moments`: tested in prod, and tested in a container with my changes).

chart.js pulls in all the moment.js locales, but only ever uses the builtin English locale data (unless you reconfigure moment.js before using chart.js).  They're abandoning time axis labeling altogether, so we'll need to handle labels ourselves:
- [issue with discussion on removing moment.js](https://github.com/chartjs/Chart.js/issues/5235)
- [PR updating docs to use chart.js without moment.js](https://github.com/chartjs/Chart.js/pull/5958)

@julianguyen Has anyone worked on #726?

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -30,7 +30,6 @@
     "axios": "^0.18.0",
     "babel-core": "7.0.0-bridge.0",
     "chart.js": "^2.6.0",
-    "chartjs": "^0.3.24",
     "chartkick": "^2.2.4",
     "es5-shim": "^4.5.9",
     "font-awesome": "^4.7.0",
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -10,6 +10,7 @@ const ManifestPlugin = require('webpack-manifest-plugin');
 const OptimizeCssAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 const webpackConfigLoader = require('react-on-rails/webpackConfigLoader');
 const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
+const webpack = require('webpack');
 const baseConfig = require('./webpack.config.base');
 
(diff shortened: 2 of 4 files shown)


client/package.json
```diff
@@ -30,7 +30,6 @@
     "axios": "^0.18.0",
     "babel-core": "7.0.0-bridge.0",
     "chart.js": "^2.6.0",
-    "chartjs": "^0.3.24",
```
julianguyen: Great catch! Oh wow, we had that and `chart.js` 🤔

client/webpack.config.js
```diff
@@ -99,6 +100,8 @@ const config = Object.assign(baseConfig, {
       hot: !!devOrTestMode,
     }),
     new ManifestPlugin({ publicPath: output.publicPath, writeToFileEmit: true }),
+    // only load moment.js data for locales we support (see config/locale.rb)
+    new webpack.ContextReplacementPlugin(/moment[\/\\]locale$/, /en|es|de|it|nb|nl|pt-BR|sv|vi|fr/),
```
julianguyen: According to Codeclimate, it looks like you don't need that escape character!

geppy: It's gone now.  Thanks!

PR #1381, opened 2019-02-28
Add filters on resources page

rhiroyuki:
# Description 
This PR enables the user to add filtering on the resources page.
- Add filter to resources page
- Add history package 
- Add dynamic url query param on resources page 

## More Details
It enables it by allowing to type on the URL the `filter[]` query param like in the example below:
```https//:localhost:3000/resources?filter[]=english```

If you wish to add multiple filters you can do it by repeating the query params:
```https//:localhost:3000/resources?filter[]=chatbot&filter[]=english```

It also adds on the React part to add/remove these filters dynamically as you click on the corresponding checkboxes (as showed on the gif below). The [history package](https://www.npmjs.com/package/history) was used in order to alter the page URL.

## Corresponding Issue
Solves #1372 

# Screenshots

ifme1

# Test Coverage

✅

### CODE REVIEW

**Comment by rhiroyuki** [comment]
One note to add, I think it would be a good call to extract `fetch_resources` and `filter_keywords` methods to a ruby service class to handle this logic outside `pages_controller`. I preferred to not do it in this PR because it would get too much out of the issue's scope.

File: client/app/widgets/Resources/index.jsx [hunk 1]
```diff
@@ -38,8 +40,21 @@ export class Resources extends React.Component<Props, State> {
     this.state = { checkboxes: this.createCheckboxes() };
   }
 
+  componentDidUpdate() {
```
julianguyen: Are you planning on adding tests for this in another PR? We should definitely be testing this!

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.3.4
```
julianguyen: Why was this committed out of curiosity? Wondering if it's necessary!

rhiroyuki: Oh sorry, this is from [asdf](https://github.com/asdf-vm/asdf), a version manager for some programming languages. I use this tool instead of rbenv/rvm and had to set it in order to use that ruby version.

Will delete it. :+1:

client/app/widgets/Resources/index.jsx
rhiroyuki: TBH, I didn't know how to test it :P
But I can give it a try in this same PR I will just need more time (probably this weekend) in order to do some research about how to test it.

.tool-versions
julianguyen: We could commit it if you think it will be helpful for others!

client/app/widgets/Resources/index.jsx
julianguyen: No worries! Let me know if you need any help, I'm happy to pair on it!

rhiroyuki: Hi @julianguyen, I asked a frontend friend for some guidance and he ended up doing some pairing with me. Because of that, we wrote the specs and also changed a little bit so `history` could be "mockable".

.tool-versions
rhiroyuki: I was looking at the asdf repo, it seems that I forgot to set it up properly (it should check `.ruby-version` and `.node-version` in case `.tools-version` is absent. :bowing_man:

PR #1384, opened 2019-02-28
Refactor strategies controller

bmwachajr:
# Description 
Refactor strategies controller removing the `# rubocop:disable ClassLength` and `# rubocop:disable MethodLength` comments that temporarily make Robocop happy.

## More Details
- Refactored the StrategiesController#premade action.
- Created a new StrategiesConrcern to harbor excess private methods

## Corresponding Issue
A partial solves the issue #840  

# Screenshots

# Test Coverage

✅


bmwachajr:
Hi @julianguyen looking forward to getting your feedback on this PR or an LGTM 🙂

app/controllers/concerns/strategies_concern.rb
```diff
@@ -0,0 +1,39 @@
...
+                  :render_errors, :empty_array_for,
+                  :premade_strategy
+  end
+
+  def publishing?
+    params[:publishing] == '1'
+  end
+
+  def saving_as_draft?
+    params[:publishing] != '1'
```
julianguyen: You could also do `!publishing?`.

spec/controllers/strategies_controller_spec.rb
```diff
@@ -82,6 +82,11 @@
           post :premade
           expect(response).to redirect_to(strategies_path)
         end
+
+        it 'returns status code 204 for json format' do
+          post :premade, as: :json
+          expect(response).to have_http_status(204)
```
julianguyen: Nice catch!

app/controllers/concerns/strategies_concern.rb
bmwachajr: I think we could rework the idea of `params[:publishing]` to determine the state(draft/published) of the `strategy` and rework the two methods `publishing?` and `saving_as_draft?` into a single check. 

I reckon that would be a less cumbersome refactor. I will look into that and raise an extended PR for this.

@julianguyen how does that sound?

julianguyen: That's a great point! That works for me :D

julianguyen:
Go ahead and merge whenever you're ready!

Issue #1160, opened 2018-10-14
Investigating the Bullet gem to help improve performance for DB queries


julianguyen:
# Description 

@jm96441n made mention of this [gem](https://github.com/flyerhzm/bullet) in a recent [pull request](https://github.com/ifmeorg/ifme/pull/1143) I made refactoring model and controller code related to Comments (thank you!)

It would be great to investigate this gem further and use it to improve performance for DB queries.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

gabrielpedepera:
Hi @julianguyen, I would like to work on this, may I?

muratbsts:
@julianguyen @gabrielpedepera if you'd not like to work on this, I'd really like.

marcellalcantara:
Just to keep everyone updated, @gabrielpedepera  and I installed the [Bullet](https://github.com/flyerhzm/bullet) gem and executed in the application.

Our first strategy was to run the specs, so we already have a `bullet.log` with some warnings to improve.

We will send the PR with the first improvements soon.

subwindow:
If no one is still working on this I'd like to take it.

julianguyen:
@subwindow Not sure if @marcelkalveram and @gabrielpedepera are still working on it. But feel free to take a stab at it! :)

marcellalcantara:
Hello. 
Sorry for the delay in answering. We have already installed the tool but have not yet had time to work on it. As soon as we start working, we’ll let you know. Hugs

cartothemax:
Looks like this issue might be good to be closed? 😄 Nice work @gabrielpedepera and @marcellalcantara

julianguyen:
It would appear so @cartothemax! I guess the only thing left to do is actually use the tool to improve query performance. Thoughts @gabrielpedepera and @marcellalcantara? :)

andrew-schutt:
I will spend some time investigating the output from `bullet.log` locally and see if we can't get some actionable issues spun up similar to the `controller` -> `request` spec refactoring that is going on right now. Similar to this collection of issues https://github.com/ifmeorg/ifme/issues/1815

julianguyen:
Awesome thanks!!

[...]

PR #1395, opened 2019-03-07
Adding and configuring bullet gem

gabrielpedepera:
Co-authored-by: Marcella Alcantara <marcellalcantara@gmail.com>

# Description 

This PR just adds and configures the [Bullet](https://github.com/flyerhzm/bullet).

This GEM provides many configs to enable the Bullet notifications, but for while we just enabled three configs to help the process of development:

`Bullet.bullet_logger`: log to the Bullet log file (Rails.root/log/bullet.log)
`Bullet.console`: log warnings to your browser's console.log (Safari/Webkit browsers or Firefox w/Firebug installed)
`Bullet.rails_logger`: add warnings directly to the Rails log

We started working on these points where they were being reported in the logs, but our lack of knowledge of the features started to get a bit difficult to map the collateral effects of our changes. 

So, we follow a different approach for while, just adding and configuring the gem to identify queries performance issues during the development of some feature. 

And after that, maybe create separate issues for specific resources. e.g.:

- comments;
- groups;
- moments;
- etc...

## More Details

The Bullet gem is designed to help you increase your application's performance by reducing the number of queries it makes. It will watch your queries while you develop your application and notify you when you should add eager loading (N+1 queries), when you're using eager loading that isn't necessary and when you should use a counter cache.

## Corresponding Issue

#1160

# Test Coverage

🚫

This PR just adds and configure a GEM to help the process of development on environments `test` and `development`. It's disabled for the production environment.

### DIFF
--- a/config/initializers/bullet.rb
+++ b/config/initializers/bullet.rb
@@ -1,0 +1,10 @@
+unless Rails.env.production?
+  Rails.application.configure do
+    config.after_initialize do
+      Bullet.enable = true
+      Bullet.bullet_logger = true
+      Bullet.console = true
+      Bullet.rails_logger = true
+    end
+  end
+end
--- a/config/initializers/bullet.rb
+++ b/config/initializers/bullet.rb
@@ -3,8 +3,6 @@ unless Rails.env.production?
     config.after_initialize do
       Bullet.enable = true
       Bullet.bullet_logger = true
-      Bullet.console = true
-      Bullet.rails_logger = true
     end
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -75,6 +75,8 @@ group :development, :test do
   gem 'rubocop-rspec'
 
   gem 'foreman'
+
+  gem 'bullet'
 end
 
 group :test do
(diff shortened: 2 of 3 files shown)


config/initializers/bullet.rb
```diff
@@ -0,0 +1,10 @@
+unless Rails.env.production?
+  Rails.application.configure do
+    config.after_initialize do
+      Bullet.enable = true
```
julianguyen: Does setting true to all these mean the same logs will be available in all three places?

If that's the case, should we just choose one place to set to true?

gabrielpedepera: Yep, that's it. We need to have the config `Bullet.enable` as true and choose another config to display the logs. 

https://github.com/flyerhzm/bullet#configuration

When we were testing, we used more the config `Bullet.bullet_logger`. Maybe, we can just leave the configs `Bullet.enable` and `Bullet.bullet_logger` and, if desirable during the process of development, who want, can activate other configs.

What do you think?

julianguyen: Yeah, let's start off with both enable and bullet_logger then!

gabrielpedepera: Done! 😄

Issue #1410, opened 2019-03-16
Add missing tests to PusherController


julianguyen:
# Description 

Looks like we're missing tests here! This can also be an opportunity to refactor the code itself! 🎉 

# Do you want to be the assignee to work on this?

🚫

DamonClark:
I wouldn't mind working on this!

PR #1414, opened 2019-03-22
Missing tests pusher controller

DamonClark:
This is an attempt at an rspec test for the pushers controller. I am still new to rspec and learning. So I am sure it will need some changes. Submitting a draft pull request. 

## More Details

https://github.com/ifmeorg/ifme/issues/1410

# Screenshots

# Test Coverage

✅

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,5 +13,5 @@
     "node_modules",
     "client/node_modules"
   ],
-  "version": "11.10.1"
+  "version": "1.9.4"
 }
--- a/package.json
+++ b/package.json
@@ -12,5 +12,6 @@
   "cacheDirectories": [
     "node_modules",
     "client/node_modules"
-  ]
+  ],
+  "version": "11.10.1"
 }
--- a/spec/controllers/pusher_controller_spec.rb
+++ b/spec/controllers/pusher_controller_spec.rb
@@ -1,0 +1,16 @@
+describe PusherController do
+
+  let(:user) { create(:user) }
+  let(:pusher) { create(:pusher) }
+
+  describe 'auth' do
+    context 'when the user is logged in' do
+      before { sign_in user }
+
+    it 'renders pusher response notification' do
+        get :auth
+        expect(response).to render_template("auth")
+      end
+    end
+  end
+end


spec/controllers/pusher_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:user) { create(:user) }
+  let(:pusher) { create(:pusher) }
+
+  describe 'auth' do
+    context 'when the user is logged in' do
+      before { sign_in user }
+
+    it 'renders pusher response notification' do
+        get :auth
+        expect(response).to render_template("auth")
```
julianguyen: What's the error message you're getting when you run this test?

DamonClark: `1) PusherController auth when the user is logged in renders pusher response notification
     Failure/Error: response = Pusher[params[:channel_name]].authenticate(params[:socket_id])

     NoMethodError:
       undefined method `length' for nil:NilClass`
`# ./spec/controllers/pusher_controller_spec.rb:11:in `block (4 levels) in <top (required)>'`

package.json
```diff
@@ -12,5 +12,6 @@
   "cacheDirectories": [
     "node_modules",
     "client/node_modules"
-  ]
+  ],
+  "version": "1.9.4"
```
julianguyen: Can you remove this from the commit? Do you know why this was added?

spec/controllers/pusher_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
+describe PusherController do
```
julianguyen: Could you use two space indentation instead of four?

spec/controllers/pusher_controller_spec.rb
julianguyen: So you'll want to test two cases:
1) When the user is not signed in
2) When the user is signed in

For case 1) you have an expect that does something like this: `expect(response).to render_template(text: I18n.t('pusher.not_authorized'))`.

For case 2), notice that the response is constructed like so: `response = Pusher[params[:channel_name]].authenticate(params[:socket_id])`

This means you'll need to pass in params for the `channel_name` and `socket_id`.

Let me know if you have any further questions!

bracken:
Since this hadn't made progress in a few months @janessatran and I picked up where this was left off. We made good progress and will probably push an updated PR next week when we work together.

If you end up finishing this before then no problem, we're just trying to find something to get us going in the project and this seemed like a good one for that.

DamonClark:
@bracken You can go ahead and move forward with this. I was focusing on it for a while but then got side tracked.

janessatran:
Hi @julianguyen, 

I have an update to this pull request, but I don't have permissions to push to this branch or project. Do you want to add that or should I make a fork and send a pull request to that? 

Thanks in advance!!

janessatran:
Hey, 

We (me and @bracken) went ahead and pushed from my own repository, as I wanted practice doing that anyway :)  So, we recommend closing this pull request and reviewing ours. Thanks!!

Issue #1390, opened 2019-03-04
Deploy storybook site to production


julianguyen:
# Description 

Currently, our storybook site is only accessible through the local development environment i.e. running `yarn storybook` in the client directory.

We want to deploy it to production as well and have it viewable at `https://www.if-me.org/design`.

We can do that by [exporting Storybook as a static app](https://storybook.js.org/basics/exporting-storybook/) and pointing the output to the new route.

Ideally, we want to this to happen in the build step in our CircleCI config so that we don't have to commit the output to master. 

# Do you want to be the assignee to work on this?

🚫 This would be a great issue to work on for folks interested in deploys! I'm happy to help if needed :)

paschalidi:
@julianguyen @KatieOfCode please let me know if I could help with anything here.

julianguyen:
@paschalidi maybe you could help review this PR: https://github.com/ifmeorg/ifme/pull/1391 and potentially pair with @KatieOfCode? :D

paschalidi:
@julianguyen I will have a look tomorrow on the PR. Thanks for letting me know. 
Can you maybe give me an idea of what the goal of https://github.com/ifmeorg/ifme/pull/1391 is? Sorry but the description is not very clear to me?

julianguyen:
@paschalidi Yeah the goal of this issue is to deploy a static version of the Storybook site (where we house our React components and CSS best practices) on production (at `https://www.if-me.org/design`).

PR #1439, opened 2019-04-07
Deploy storybook site to production

paschalidi:
👋 👋 hello fellow contributors 🌸 
# Description 

The goal is to have a page with static storybook deployed.
I aborted `storybook-deployer` as it was complaining quite a lot for no good reason (one that I could understand at least 😅). 
We use `surge` instead to deploy the static storybook website on https://storybook-ifme.surge.sh 
The logic behind is the same, no static code committed on `master`. Storybook is getting build and deployed via `surge` during the CI time.

The new job that has been added on this PR `deploy-storybook` will be triggered only when a PR is merged on `master` as we dont want storybook to be deployed on every PR one makes.

## Corresponding Issue

Closes https://github.com/ifmeorg/ifme/issues/1390

# Test Coverage

This PR contains only infrastructural work thus no tests have been written.


paschalidi:
@julianguyen this is now ready for review !! :D

.circleci/config.yml
```diff
@@ -130,6 +130,42 @@ jobs:
...
+      - run: rm -rf ~/.yarn
+      - run: npm i -g yarn
+      - run: yarn install
+      - run:
+          name: Deploy Storybook
+          command: |
+            cd client
+            yarn build:storybook
+            mv .out .public
+            yarn run surge --project .public --domain storybook-ifme.surge.sh
```
julianguyen: Let's rename this to ifme.surge.sh and I can point if-me.org/design to this url!

PR #1444, opened 2019-04-11
Upload Images in Moments

### LINKED ISSUE #1480: Back-end: uploading Images in Moments and Strategies
# Description 

We want to have a backend API endpoint that uploads images to Cloudinary.

reference feature: #798 
# Reproduction Steps

# Screenshots

✅

bmwachajr:
# Description 
When creating a moment. it is valuable to upload an image to capture the moment/mood of the user.
## More Details

More Details:
- Upload images to cloudinary.

To do:
- [x] Create a cloudinary service
- [x] create an picture upload endpoint
- [x] add tests
- [ ] Ship it.

## Corresponding Issue
Fixes #1480 

# Screenshots

# Test Coverage

✅ 
🚫


julianguyen:
Is this ready for review? :)

bmwachajr:
@julianguyen yes it is ready for review. A couple of things to note.
1. Do we have the cloudinary `env` values set on CircleCi?
2. Integration Test for the API endpoint has not yet been committed, I keep getting a RoutingError.

julianguyen:
We set the Cloudinary env variables in Heroku!

app/controllers/moments_controller.rb
```diff
@@ -70,6 +70,18 @@ def destroy
     redirect_to_path(moments_path)
   end
 
+  # POST /moments/1/picture
+  # POST /moments/1/picture.json
+  def picture
+    moment = set_moment(params[:moment_id])
```
julianguyen: Can we set the value of moment after line 79?

spec/services/cloudinary_service_spec.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+FILENAME = "moment.jpeg"
+FILE_PATH = "/public/uploads/tmp/moment"
```
julianguyen: Can we mock this file path instead of using a real one?

bmwachajr:
@julianguyen yeah, all tests pass locally. It would be a good idea to set the environment variables for Cloudinary in CircleCI! Otherwise, once this PR is merged we shall always have these tests failing.

spec/services/cloudinary_service_spec.rb
bmwachajr: Apparently, to successfully upload to cloudinary, I need to attach a real file. I haven't been successful in mocking an image.

You are right about not using a real file path. As a workaround(fairly embarrassed) I created an uploads folder in the specs module. The real image will reside there and we can attach it using the `spec/uploads/` file path.

Pro of this is that this image is reusable in the future. and doesn't reside in a sensitive place.

Not my best work. I am open to jamming on this for a better workaround.

app/controllers/moments_controller.rb
bmwachajr: Nice catch :)

spec/services/cloudinary_service_spec.rb
julianguyen: No worries and I think what you have is fine! :)

PR #1451, opened 2019-04-19
Bug fix: accepted ally invite doesn't create allies

julianguyen:
# Description 

* Fixed things so that accepted ally invite creates an allyship with status `pending_from_ally` between the invitee and inviter.
* Create migration to rename name columns in database to email if empty

Related to the following issue on production: https://sentry.io/share/issue/7ee8ad53c48646a4b7cf517eb0484990/

# Test Coverage

✅ 

I also did manual testing!


app/controllers/users/invitations_controller.rb
```diff
@@ -16,6 +16,32 @@ def create
       redirect_to new_user_invitation_path
     end
 
+    # PUT /resource/invitation
```
julianguyen: Comes from https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L46

Issue #1463, opened 2019-04-26
Delete and Edit tooltips can be misconstrued as a button


toshitapandey:
## Description 

When I go to `/categories` route after I have added some of the categories, I can edit or delete the categories that I have. As of now, the tooltip can be misconstrued as a button. It should look like info rather than a button. 

## Reproduction Steps

- Sign up (_if you haven't already_)
- Go to categories (_by clicking Categories on the left-hand side panel_)
- Click on "Add all" (_this will add all the listed categories to your categories_)
- Hover on "edit" or "delete" icon at the right-hand side of each category

### Current implementation: 
![image](https://user-images.githubusercontent.com/3788470/56784768-db075600-680f-11e9-8860-0f5ed548fda0.png)
### Expected implementation: 
![image](https://user-images.githubusercontent.com/3788470/56784900-7f899800-6810-11e9-9afa-0d979ec3cdc4.png)

For doing this, title attribute can be added to the anchor tag. 
e.g.
```
<a title="Some text that will show up when I hover over this link">
  My link
</a>
```
There can be better ways of handling this.

256hz:
I'd like to give this one a shot please!

julianguyen:
We would love your help @256hz! Please assign yourself to the issue - thanks ^^

256hz:
I'm still working on getting the site running locally, I'll claim it as soon as I do, don't want to sit on it in case it takes me a while.  Thanks!

256hz:
I solved this in the tooltip CSS and made a pull request.   Hope this helps!

PR #1471, opened 2019-05-03
Delete and Edit tooltips can be misconstrued as a button #1463

256hz:
# Description 

The old tooltip styling looks like a clickable button, so I modified it to look more like the example given by @toshitapandey.  Now the background is dark gray, the text is light gray and smaller, and the corner radius is reduced.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1463

# Screenshots

abe-tooltip-1463

# Test Coverage

🚫 

Visual only


256hz:
Looks like I joined at the same time as someone else and our blurbs are conflicting, I would resolve this but don't want to commit to master!

app/assets/stylesheets/base/_values.scss
```diff
@@ -1 +1 @@
-$z-index-front: 9000;
+$z-index-front: 9001;
```
julianguyen: Could you explain why you increased this value?

client/app/components/Tooltip/Tooltip.scss
```diff
@@ -12,16 +12,16 @@
...
     text-transform: none;
     letter-spacing: normal;
-    color: $black;
+    color: $light-grey;
     box-sizing: border-box;
     box-shadow: $size-0 $size-2 $size-10 $black-10;
-    border-radius: $size-4;
-    background: $white;
+    border-radius: $size-1;
+    background: $black-70;
```
julianguyen: I think it would be worth checking whether the colour contrast is good accessibility.

You can use this [tool](https://webaim.org/resources/contrastchecker/) :D

I think it would be great to have @nshki review the design side of this!

doc/pages/blurbs.json
```diff
@@ -438,5 +438,14 @@
...
     "social": "github"
+  },
+  {
+    "name": "Abe Dolinger",
+    "image": "/assets/contributors/abe_dolinger.jpg",
+    "profile": "I'm a product designer, startup vet, sound recordist, musician, & software engineer.  I'm most proficient in Ruby on Rails, JS, and Python.  I'm a strong believer in increasing resources for, and awareness of, mental health, so I was very happy to find if-me!",
+    "location": "Seattle, WA",
+    "link_name": "abedolinger",
+    "link": "https://github.com/256hz",
+    "social": "github"
```
julianguyen: This is great! I think to separate the logic of this PR, it would be worth making a separate one just for your contributor blurb :)

client/app/components/Tooltip/Tooltip.scss
```diff
@@ -12,16 +12,16 @@
   }
 
   > div:last-of-type {
-    @include setFontSize($size-18);
-    @include setPadding($size-10, $size-10, $size-10, $size-10);
+    @include setFontSize($size-14);
+    @include setPadding($size-6, $size-6, $size-6, $size-6);
```
nshki: Let's actually leave this as it was before at `$size-10`. Making the tooltip too slim makes it look a little bit out of place in the application.

client/app/components/Tooltip/Tooltip.scss
```diff
@@ -12,16 +12,16 @@
...
     letter-spacing: normal;
-    color: $black;
+    color: $light-grey;
     box-sizing: border-box;
     box-shadow: $size-0 $size-2 $size-10 $black-10;
-    border-radius: $size-4;
-    background: $white;
+    border-radius: $size-1;
+    background: $black-70;
     width: 200px;
```
nshki: Let's tweak this so it's 150px.

client/app/components/Tooltip/Tooltip.scss
```diff
@@ -12,16 +12,16 @@
...
     font-weight: $font-weight-400;
     text-transform: none;
     letter-spacing: normal;
-    color: $black;
+    color: $light-grey;
     box-sizing: border-box;
     box-shadow: $size-0 $size-2 $size-10 $black-10;
-    border-radius: $size-4;
-    background: $white;
+    border-radius: $size-1;
```
nshki: Let's actually also leave this as it was before at `$size-4`. Since many components use a larger radius, having something less makes it look out of place.

app/assets/stylesheets/base/_values.scss
nshki: Seconding.

256hz: Not a serious suggestion! I thought the original value of 9000 was cheeky, so I +1'ed it to be over 9000.  😅 
```suggestion
$z-index-front: 9000;
```
is good.

PR #1472, opened 2019-05-05
Replace infinite scroll with load more button

julianguyen:
# Description 

Replace infinite scroll with load more button because it's better user experience. When there were many stories, infinite scroll made it nearly impossible to get to the footer of the site.

# Screenshots

capture

# Test Coverage

✅

### DIFF
--- a/client/app/components/BaseContainer/index.jsx
+++ b/client/app/components/BaseContainer/index.jsx
@@ -1,18 +1,19 @@
 // @flow
-import React from 'react';
+import React, { Fragment } from 'react';
 import axios from 'axios';
 import { StoryContainer } from './StoryContainer';
+import css from '../../styles/_global.scss';
+import { I18n } from '../../libs/i18n';
 
 export type Props = {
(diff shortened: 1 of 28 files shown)


client/app/components/BaseContainer/index.jsx
```diff
@@ -1,99 +1,76 @@
...
-        .get(url)
-        .then((response: any) => {
-          this.setState({
-            loading: false,
-            page: page + 1,
-            data: data.concat(response.data.data),
-            isLastPage: response.data.lastPage,
-          });
-        })
-        .catch((error) => {
```
julianguyen: Removing the catch block because we want to throw the error from the server. The value of loading will already be false if an error is thrown.

Jacek-202:
Looks good for me :). Nice you added index_spec.jsx :clap: :clap:

bmwachajr:
I hadn't seen this PR 🙈, It looks awesome. Great Work on this.

PR #1525, opened 2019-06-17
Adding error handling and tests to PusherController

janessatran:
# Description 
If the two required parameters, `channel_name` and `socket_id`, are not passed to the Pusher endpoint, an error message is returned in the same Rails ActiveModel error format as the other controllers.

## Corresponding Issue
#1410 

# Test Coverage
✅

### DIFF
--- a/app/controllers/pusher_controller.rb
+++ b/app/controllers/pusher_controller.rb
@@ -8,10 +8,21 @@ class PusherController < ApplicationController
   #
   # Returns an auth token from Pusher that looks like this
   # {:auth=>":7944d5b3dc2915f6a9c55dbedcbe91037799bb3627fc868890c2a7fd6f5fde4b"}
+  # 
+  # If there is an error, the response will look like this
+  # {"channel_name":["can't be blank"],"socket_id":["can't be blank"]}
   def auth
     if current_user
-      response = Pusher[params[:channel_name]].authenticate(params[:socket_id])
-      render json: response
+      errors = {}
+      errors[:channel_name] = ["can't be blank"] if params[:channel_name].blank?
+      errors[:socket_id] = ["can't be blank"] if params[:socket_id].blank?
+      
+      if errors.empty?
+        response = Pusher[params[:channel_name]].authenticate(params[:socket_id])
+        render json: response
+      else
+        render json: errors, status: :bad_request
+      end
     else
       render text: t('pusher.not_authorized'), status: '403'
     end
--- a/app/controllers/pusher_controller.rb
+++ b/app/controllers/pusher_controller.rb
@@ -4,6 +4,10 @@ class PusherController < ApplicationController
   # stop rails CSRF protection for this action
   protect_from_forgery except: :auth
 
+  # POST /pusher/auth
+  #
+  # Returns an auth token from Pusher that looks like this
+  # {:auth=>":7944d5b3dc2915f6a9c55dbedcbe91037799bb3627fc868890c2a7fd6f5fde4b"}
   def auth
     if current_user
       response = Pusher[params[:channel_name]].authenticate(params[:socket_id])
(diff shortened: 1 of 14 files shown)


app/controllers/pusher_controller.rb
```diff
@@ -4,10 +4,27 @@ class PusherController < ApplicationController
   # stop rails CSRF protection for this action
   protect_from_forgery except: :auth
 
+  # POST /pusher/auth
+  # @argument channel_name [String] The name of the pusher channel.
+  # @argument socket_id [String] The id of the pusher socket.
+  #
+  # Returns an auth token from Pusher that looks like this
+  # {"auth": ":7944d5b3dc2915f6a9c55dbedcbe91037799bb3627fc868890c2a7fd6f5fde4b"}
```
julianguyen: Is this an actual Pusher auth key that's in use? If so, we should remove it from this comment and generate a new one. You could replace the string with `"some-auth-key"`.

app/controllers/pusher_controller.rb
```diff
@@ -4,10 +4,27 @@ class PusherController < ApplicationController
...
+  # 
+  # If there is an error, the response will look like this
+  # {"channel_name":["can't be blank"],"socket_id":["can't be blank"]}
```
julianguyen: I think we should consider internationalization (i18n) here!

app/controllers/pusher_controller.rb
janessatran: Hey Julia! No, this isn't a real one. We just put it up as an example, but I'll swap it out for "some-auth-key" instead so others don't worry too :)

Thanks!

app/controllers/pusher_controller.rb
janessatran: Thanks for that suggestion!

janessatran:
@julianguyen - thanks for the feedback! We addressed your comments, let us know if you would like us to update anything else :D

config/locales/en.yml
```diff
@@ -154,6 +154,7 @@ en:
       header_message: 'This page cannot be displayed'
     not_found:
       header_message: 'This page does not exist'
+    empty_params: 'Can''t be blank'
```
julianguyen: So you'll need to translate this to the other locales! You can use Google translate for now!

app/controllers/pusher_controller.rb
```diff
@@ -4,12 +4,26 @@ class PusherController < ApplicationController
...
+  # @argument channel_name [String] The name of the pusher channel.
+  # @argument socket_id [String] The id of the pusher socket.
+  #
+  # Returns an auth token from Pusher that looks like this
+  # {"auth": "some-auth-key"}
+  #
+  # If there is an error, the response will look like this
+  # {"channel_name":["can't be blank"],"socket_id":["can't be blank"]}
   def auth
-    if current_user
```
janessatran: Because PusherController is a child of ApplicationController, it is protected by the `before_action :if_not_signed_in`, so there will always be a current user and we don't need to check it in this endpoint.

janessatran:
Hey @julianguyen- we added the locales and addressed the CodeClimate concerns! :)

julianguyen:
Awesome work! It looks like there's one failing test in CircleCI. Let me know if you need any help addressing it! https://circleci.com/gh/ifmeorg/ifme/8675?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

julianguyen:
Looks like there's still a [failing test](https://circleci.com/gh/ifmeorg/ifme/8685). Let me know if you need any help resolving it!

janessatran:
Hey @julianguyen, can you rerun the CircleCI tests? We don't have permissions to rerun it and we determined the issue is that the failing test is flaky: #1528 . 

We're happy to look into that but we want to finish our google profile picture feature first. We're hoping that if you retrigger this we'll be able to resolve this PR :) 

Thanks!!

PR #1526, opened 2019-06-25
Add missing group controller tests

janessatran:
# Description 

Added missing tests for update and create and destroy. 

## More Details

We ran SimpleCov to see what tests were missing in the controllers, and noticed GroupsController was lacking tests.

# Test Coverage

✅


spec/controllers/groups_controller_spec.rb
```diff
@@ -77,7 +77,7 @@
     end
   end
 
-  describe 'GET #update' do
+  describe 'PUT #update' do
```
julianguyen: Good catch!

Issue #1373, opened 2019-02-20
Inconsistent padding on forms with format controls


Gibbo3771:
# Description 

When creating a new moment/category/mood/strategy, any form field that hosts the format controls has some inconsistent padding going on. I suggest that the padding be adjusted to prevent odd resizing to occur after the first character is typed. 

# Reproduction Steps

I would say it's a bug. Go to Moods, then new and in any form with format controls, click and you can see that the caret is placed. It's _barely_ inside the box, enter any character and it will fix. If you can backspace the character away, and see that the padding change remains consistent, until you backspace again and remove a phantom character.

# Screenshots
![padding-issue](https://user-images.githubusercontent.com/10871784/53096779-cf4d9980-3517-11e9-92d6-49707ead52d0.gif)

# Do you want to be the assignee to work on this?

🚫

Gibbo3771:
Just an update on this, I have tested the current code across different browsers and this seems to be a problem in Firefox. Chrome is working fine, so not sure how to tackle it.

ajquintero:
Currently trying to figure out the problem for Firefox.

Gibbo3771:
> Currently trying to figure out the problem for Firefox.

Are you able to reproduce? I am running the following OS and Firefox version. I can't confirm the bug on Windows, as I don't have a windows machine at the moment.

- 4.19.24-1-MANJARO
- Mozilla Firefox 65.0.1
- i3 version 4.16.1

ajquintero:
@Gibbo3771 I'm running on a Mac (OS 10.14.3) and Firefox 65.0.1. The issue is reproducing.

vildevev:
Is anyone working on this? If not I'm interested.

ajquintero:
@vildevev I'm looking into it again, but feel free to work on it as well!

neon98:
@vildevev are you working on this? If not I'd want to work on this.

PR #1558, opened 2019-09-24
Eliminate inconsistant padding in InputTextarea

neon98:
# Description 

Inconsistent padding was going on in any form field that hosts the format controls in firefox browser. Added min-height to solve this issue.

## Corresponding Issue
#1373 

# Screenshots
Screenshot (186)

# Test Coverage

✅

### DIFF
--- a/app/assets/stylesheets/base/_sizes.scss
+++ b/app/assets/stylesheets/base/_sizes.scss
@@ -27,4 +27,4 @@ $size-46: 46px;
 $size-48: 48px;
 $size-50: 50px;
 $size-60: 60px;
-$size-73: 73px;
+$size-70: 70px;
--- a/app/assets/stylesheets/base/_sizes.scss
+++ b/app/assets/stylesheets/base/_sizes.scss
@@ -27,3 +27,4 @@ $size-46: 46px;
 $size-48: 48px;
 $size-50: 50px;
 $size-60: 60px;
+$size-73: 73px;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/base/_sizes.scss
```diff
@@ -27,3 +27,4 @@ $size-46: 46px;
 $size-48: 48px;
 $size-50: 50px;
 $size-60: 60px;
+$size-73: 73px;
```
julianguyen: Could you make this value 70px instead - just to be consistent with the even numbers we use?

client/app/components/Input/InputTextarea.scss
```diff
@@ -59,3 +59,7 @@
     }
   }
 }
+
+.minHeight{
```
julianguyen: Could you rename this to `.editor`?

neon98: should I replace editor with css.editor then? or keep both editor and css.editor?

julianguyen: I think you'll want to keep both actually since we use editor as a reference point in our acceptance tests.

You'll want to do this: 
```
<div className={`editor ${css.editor}`} />
```

neon98: did the changes. Thank you :D

Issue #1159, opened 2018-10-14
Add missing test coverage for Notifications widget


julianguyen:
# Description 

Oh noes! Missing test coverage, let's fix that! It's also an opportunity to do any refactoring needed in the widget.

For an idea on how to spy and stub out Axios, see the Comments widget.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to join our [organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Gibbo3771:
@gitburd still working on this?

Gibbo3771:
Gonna have to drop my progress on this, not had any self care time all week!

gitburd:
This is my first contribution and I can't really figure it out. I need to
find someone to pair with me. I'm gonna post on slack.

julianguyen:
Hey @gitburd I'm happy to pair with you, let's chat more on Slack!

ashwanijha04:
Any updates on this? Can I help too?

PR #1562, opened 2019-09-30
Add test coverage for Notifications widget

briceyokoyama:
# Description 

- Added tests for Notifications widget.
- Modified fetchNotifications function in Notifications widget to return promise (used for testing only).

## More Details

Test checks:
- Notification renders correctly.
- fetchNotifications updates state.notifications with data received from API.
- Clear button closes modal.
## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1159

# Test Coverage

✅

### DIFF
--- a/client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
+++ b/client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
@@ -2,7 +2,7 @@
 import React from 'react';
 import axios from 'axios';
 import { mount } from 'enzyme';
-import { Notifications } from '../index'
+import { Notifications } from '../index';
 
 let axiosGetSpy;
 let axiosDeleteSpy;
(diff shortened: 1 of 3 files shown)


client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+const data = {
+  signed_in: 1,
+  fetch_notifications: ['Notification 1'],
+};
+
+const component = (
+  <Notifications
+    element={<button type="button">Notifications</button>}
+    plural="Notifications"
```
julianguyen: Ooh, this is an opportunity to refactor this component so we don't need to pass in prop. We were originally doing this because i18n wasn't working in the React side. 

You can pull in the translation key directly in the component by importing the i18n service from the libs directory and using `I18n.t('notifications.plural')` in the component.

client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
```diff
@@ -0,0 +1,76 @@
...
+  beforeEach(() => {
+    axiosGetSpy = jest.spyOn(axios, 'get').mockImplementation(() => Promise.resolve({
+      data,
+    }));
+    axiosDeleteSpy = jest.spyOn(axios, 'delete').mockImplementation(() => Promise.resolve({
+      data: { ok: true },
+    }));
+    componentDidMountSpy = jest
+      .spyOn(Notifications.prototype, 'componentDidMount')
+      .mockImplementation(() => {});
```
julianguyen: Could you explain why you decided to mock out `componentDidMount()` and not have it call `fetchNotifications()`? I think it should be enough to use your Axios spies above. It's a general best practice to use mocks only when necessary.

PR #1572, opened 2019-10-04
Update fr.yml

nicovak:
# Description

I Tried to improve It as much as I can. Some recurring mistakes are missing space before (and after)  `?` and `!`. Others are just wrong word usage or misspelling.

## More Details

Also note if a sentence is using `'` I would suggest a `"` to declare It

---

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -318,7 +318,7 @@ fr:
       premade4_description: 'Ne pas accepter certain faits, se sentir en colère, vindicatif(ve), et/ou jaloux(se)'
       premade5_name: Désespoir
       premade5_description: 'Sentiment de tristesse et déséspoir - incapable de trouver une solution'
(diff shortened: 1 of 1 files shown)


config/locales/fr.yml
```diff
@@ -695,21 +695,21 @@ fr:
...
         subtitle: 'Commencez à raconter vos histoires et maintenez votre santé dans un espace privé et sécurisé'
         label: 'Quelques détails'
     sessions:
       subtitle: 'Bon retour parmi nous, nous nous réjouissons de pouvoir vous aider à attaquer cette journée'
     shared:
-      sign_in_google: 'S''identifier avec Google'
-      sign_up_google: 'S''enregistrer avec Google'
-      confirmation: 'Vous n''avez pas reçu vous informations de confirmation?'
-      unlock: 'Vous n''avez pas reçu vos informations pour débloquer votre compte?'
+      sign_in_google: "S'identifier avec Google"
```
julianguyen: I think we should be consistent about whether we want to use double-quotes or single quotes. Your changes above got rid of the double quotes. But this change introduces them. I would pick single quotes.

PR #1574, opened 2019-10-04
gitignore update and clean

nicovak:
# Description

`vscode` config should not be present.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@
 
 # Ignore all IDE, TextEditor configuration files
 .vs/
+.vscode/
 .idea/
 .tags
 *.swp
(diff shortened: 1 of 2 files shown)


julianguyen:
Could you explain why it shouldn't be present? We include other editors in our `.gitignore`. VS Code config is typically in either`.vs` or `.vscode`.

nicovak:
Editor's config is specific to user and should not be present on the remote repository to avoid unwanted configuration or junk. As you said, `.gitignore` includes already `.idea` , `.vs` so why not `.vscode` ?
`# Ignore all IDE, TextEditor configuration files` is specified in the `.gitignore`.
If you want to keep it locally but not on remote repository you can do a `git rm -r --cached .vscode`

It's my preferrence and peoples have different opinion about it ;)

PR #1577, opened 2019-10-05
Adding / refactoring some specs

guilhermekbral:
Adding [Shoulda Matchers gem](https://github.com/thoughtbot/shoulda-matchers), adding more models specs, and refactoring some specs

### DIFF
--- a/spec/models/category_spec.rb
+++ b/spec/models/category_spec.rb
@@ -18,6 +18,11 @@ describe Category do
   context 'with validations' do
     it { is_expected.to validate_presence_of :user_id }
     it { is_expected.to validate_presence_of :name }
+
+    it 'is invalid without a name' do
+      category = build(:category, name: nil)
+      expect(category).to have(1).error_on(:name)
+    end
   end
 
   context 'with relations' do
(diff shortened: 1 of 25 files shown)


spec/models/category_spec.rb
```diff
@@ -13,28 +13,14 @@
...
 
-    it 'is invalid without a user_id' do
-      category = build(:category, user_id: nil)
-      expect(category).to be_invalid
-    end
-
-    it 'is invalid without a name' do
-      category = build(:category, name: nil)
-      expect(category).to have(1).error_on(:name)
-    end
```
julianguyen: Could you explain why you're removing these error relayed tests?

guilhermekbral: Hi Julia! Since I'm validating presence with the `validate_presence_of` matcher, for me this spec is redundant. In a matter of fact, we'll have more redundant specs in other models, but I'm not sure what should we do in this case.

If we keep them, our suite will grow bigger, but we'll have redundant specs. To remove them, we'd need to be sure that we're still testing each case.

For now, I'll keep this test under the `validations` context, and will start a discussion in the dev channel to get some insights, what do you think?

julianguyen: Thanks for explaining! Sounds good to me to remove!

Issue #1548, opened 2019-08-31
Fix flakey toggle_locale_spec.rb test


julianguyen:
# Description 

```
Failure/Error: expect(find(check_for_element, wait: 10)).to condition

Capybara::ElementNotFound:
  Unable to find css ".pageTitle"
./spec/support/page_transition_support.rb:15:in `change_page'
./spec/features/toggle_locale_spec.rb:33:in `block (4 levels) in <top (required)>'
```

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

guilhermekbral:
Good afternon @julianguyen. I was taking a look at this, and the spec is green here. Ran it against `master` branch.

Is there a specific branch in which the spec is broken? Is it fixed already?

julianguyen:
Good question @guilhermekbral! One example is `44331`! It's not fixed yet!

Please [join](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack) our contributor community so you can assign yourself to issues!

guilhermekbral:
Nice! Just sent the email! I'll take a look with that seed! Thanks for the info!

tiffanyisapr0:
This appears to be fixed! @julianguyen can you confirm that? :)

julianguyen:
@tiffanyisapr0 I don't think it's fixed! I had to re-build some PRs that were failing cause of it this week.  It's flaky so sometimes it passes.

guilhermekbral:
I'm still working on this flaky spec. It's trickier than I expected!

sbpipb:
True, I can attest that this one is indeed still happening

leeacto:
I couldn't make the test "flakey" (intermittently failing), but I did see if consistently fail. An error was raised because it couldn't find the translation key. This key is not used anymore and was replaced with a hardcoded `Linebreak`.

Removing the translation attempt let the page load successfully and pass the test.

julianguyen:
I think this error is still happening @leeacto :( It existed before we introduce the Chinese translations you fixed up the bug in :/

leeacto:
Oh bummer. Thanks I'll look some more.

julianguyen:
[...]

PR #1584, opened 2019-10-08
updated code to resolve warnings

dhuang612:
refactored code and ran yarn lint and yarn test
both completed without errors.

### DIFF
--- a/client/app/components/Header/HeaderProfile.jsx
+++ b/client/app/components/Header/HeaderProfile.jsx
@@ -2,11 +2,10 @@
 import React from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faBell } from '@fortawesome/free-solid-svg-icons';
-import { Avatar } from '../Avatar';
-import type { Profile } from './index';
+import type { Profile } from './types';
 import { Notifications } from '../../widgets/Notifications';
-// eslint-disable-next-line import/no-cycle
-import css from './Header.scss';
+import { Avatar } from '../Avatar';
+import css from './HeaderProfile.scss';
 import globalCSS from '../../styles/_global.scss';
 
 export type Props = {
--- a/client/app/components/Header/HeaderProfile.scss
+++ b/client/app/components/Header/HeaderProfile.scss
@@ -1,0 +1,45 @@
+@import "../../styles/_global.scss";
+
+  .header {
+    z-index: $z-index-front;
+  
+    > div {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+    }
+    &Profile {
+        width: 100%;
+        box-sizing: border-box;
+        padding: $size-22 $size-26;
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        justify-content: space-between;
+        color: $white;
+        &Info {
+          @include setFontSize($size-18);
+          color: $white;
+          &Links {
+            @include setFontSize($size-0);
+            margin-top: $size-4;
+            line-height: 3.5;
+            display: flex;
+            flex-direction: row;
+            flex-wrap: wrap;
+            align-items: center;
+            justify-content: space-evenly;
+            > div {
+              margin: $size-8 $size-8 $size-0 $size-0;
+              &:last-of-type {
+                margin: $size-8 $size-0 $size-0 $size-0;
+              }
+            }
+          }
+        }
+        @media screen and (max-width: $medium) {
+          text-align: center;
+          justify-content: center;
+        }
+      } 
+    }    
--- a/client/app/components/Header/index.jsx
+++ b/client/app/components/Header/index.jsx
@@ -4,30 +4,10 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faBars, faTimes } from '@fortawesome/free-solid-svg-icons';
 import ReactHtmlParser from 'react-html-parser';
 import { Logo } from '../Logo';
-// eslint-disable-next-line import/no-cycle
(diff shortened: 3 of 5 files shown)


dhuang612:
The error I'm receiving is related to ruby and not related to any of the files I touched?
rspec ./spec/features/toggle_locale_spec.rb:28 # ToggleLocale Toggling the locale dropdown to change the language When on pages that do not require sign in persists locale selection on a different page

julianguyen:
@dhuang612 The test that's failing is sadly a flaky test we're working on [fixing](https://github.com/ifmeorg/ifme/issues/1548). 

You can re-run the test:

image

You'll have to sign into CircleCI with your GitHub credentials to access our org!

client/app/components/Header/index.jsx
```diff
@@ -112,7 +111,7 @@ export class Header extends React.Component<Props, State> {
     return (
       <div id="headerMobile" className={css.headerMobileNav}>
         <div>
-          {profile ? <HeaderProfile profile={profile} /> : null}
+          {profile ? <Avatar profile={profile} /> : null}
```
julianguyen: Unfortunately, you can't swap this component out for `<Avatar />` since `<HeaderProfile />` has is more than just `<Avatar />`. The issue with the circular dependency comes from `import type { Profile } from './index';` in the HeaderProfile component.

dhuang612: thank you for the help!
I'm going to continue reviewing this tomorrow.

julianguyen: Oh you're right! The issue is with `import css from './Header.scss';` cause it's imported in both files!

dhuang612: cool, just double checking
thanks!

dhuang612:
So I've been reviewing this for a while now.
Under HeaderProfile.jsx 
the comment can just be removed and nothing changes.

However under index.js it looks like this circular dependence can't be removed because
under Header it makes a call to Avatar which can't be removed and then under Index it's used once under mobile links which is creating the circular logic.

dhuang612:
never mind I understand my error.

dhuang612:
second update, it's not the css files that are causing the warning. 
I completely separated the scss files and am still getting the warning under index.jsx!

client/app/components/Header/HeaderProfile.scss
```diff
@@ -0,0 +1,45 @@
+
+  .header {
```
julianguyen: This is still a good refactor! You'll want to import `@import "../../styles/_global.scss";` at the top of this file. 

I think you just want to keep what's at line 10 and beyond.

client/app/components/Header/HeaderProfile.jsx
```diff
@@ -2,11 +2,10 @@
 import React from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faBell } from '@fortawesome/free-solid-svg-icons';
-import { Avatar } from '../Avatar';
 import type { Profile } from './index';
```
julianguyen: I think still might be the issue. `Profile` comes from [Header/index.jsx](https://github.com/ifmeorg/ifme/pull/1584/files#diff-92470f5ab85c5de61d20bac88b5732deR19) `Header/index.jsx` pulls in `<HeaderProfile />` so that's a circular dependency.

dhuang612: Hi Julian,

That's what I am getting at, I'm not sure this can be refactored to resolve the circular import error.
**HeaderProfile**
```
import type { Profile } from './index';

export type Props = {
  profile: Profile,
};
```

**index.jsx**
```
import { HeaderProfile } from './HeaderProfile';

export type Profile = {
  avatar?: string,
  name: string,
  profile: Link,
  account: Link,
  notifications: {
    plural: string,
    none: string,
    clear: string,
  },
};
```
Profile is being imported and exported between the two files and I don't know if the circular logic can be broken?
I can finish up the css refactoring

dhuang612:
https://stackoverflow.com/questions/58292227/working-on-refactoring-a-react-component-to-remove-import-no-cycle

client/app/components/Header/HeaderProfile.jsx
dhuang612: Hi Julia,

Thanks for the response, sorry about misspelling your name!

dhuang612: @julianguyen 
Hey, I separated the profile into it's own file profile.js
I then imported it into the two files and ran into the same  import/no-cycle error.
I'm update my PR with this new code for you to review.

Thanks!

dhuang612:
Hi Julia,
I think with your help I was able to get this refactor completed.
Thanks a lot for your assistance and helpful hints!
I've gone ahead and squashed the old commits to bring it back down to 1.
@julianguyen

client/app/components/Header/profile.js
```diff
@@ -0,0 +1,21 @@
+// @flow
```
julianguyen: Let's rename this `types.js`.

client/app/components/Header/HeaderProfile.scss
```diff
@@ -0,0 +1,46 @@
...
+              }
+            }
+          }
+        }
+        @media screen and (max-width: $medium) {
+          text-align: center;
+          justify-content: center;
+        }
+      } 
+
```
julianguyen: Let's remove this extra line.

dhuang612: done

client/app/components/Header/profile.js
dhuang612: and done
Thanks

Issue #1545, opened 2019-08-31
Replace react-autocomplete with react-autosuggest


julianguyen:
# Description 

[react-autocomplete](https://github.com/reactjs/react-autocomplete) is no longer being maintained. 

After updating our client to the latest version of React, we receive these warnings:
```
      Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

      * Move code with side effects to componentDidMount, and set initial state in the constructor.
      * Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

      Please update the following components: Autocomplete
    console.warn node_modules/react-dom/cjs/react-dom.development.js:11494
      Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

      * Move data fetching code or side effects to componentDidUpdate.
      * If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
      * Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

      Please update the following components: Autocomplete
```

We could hypothetically fork `react-autocomplete` and update React ourselves and use it in our client, but that creates an maintenance overhead for us. The best course of action is to use another framework. 

Let's use [react-autosuggest](https://github.com/moroshko/react-autosuggest) instead!

You'll want to update `client/app/components/Input/InputTag.jsx`.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

CSumm:
Is anyone assigned to this issue? I'd like to take a shot at it if it's still available

neon98:
hey @CSumm are you working on this issue? I'd love to help!

[...]

PR #1591, opened 2019-10-15
Replace react-autocomplete with react-autosuggest issue #1545

SonuToor:
# Description

Replaced the react-autocomplete library with the react-autosuggest library

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1545


client/app/components/Input/InputTag.jsx
```diff
@@ -65,9 +65,8 @@ export class InputTag extends React.Component<Props, State> {
     const inputValue = autocompleteLabel.trim().toLowerCase();
     const inputLength = inputValue.length;
```
SonuToor: I had tried fixing 
`Error -------------------------------------------------------------------------- app/components/Input/InputTag.jsx:68:37

Missing type annotation for `T`. `T` is a type parameter declared in array type [1] and was implicitly instantiated at
call of method `filter` [2].
`
This current solution obviously didn't work.

I was thinking maybe if I add the following code at line 67 it might pass this error.

`    const { checkboxes } = this.state;
`

client/app/components/Input/InputTag.jsx
```diff
@@ -103,13 +102,13 @@ export class InputTag extends React.Component<Props, State> {
...
 
-  onAutocompleteChange = (e: SyntheticEvent<HTMLInputElement>, { newValue }) => {
+  onAutocompleteChange = (e: SyntheticEvent<HTMLInputElement>, { newValue }: {newValue: string}) => {
     this.setState({
       autocompleteLabel: newValue,
     });
   };
 
-  onSelect = (event: SyntheticEvent<HTMLInputElement>, { suggestion, suggestionValue } ) => {
+  onSelect = (event: SyntheticEvent<HTMLInputElement>, { suggestion, suggestionValue }: {suggestion: string, suggestionValue: string} ) => {
```
SonuToor: Over here the suggestionValue isn't needed so I will go ahead and remove that. But even still the issue will persist I believe. Don't really know what to do here.

julianguyen: It looks like this error is happening because suggestion is a string therefore it can't have the label key. Should you be doing this on a `checkbox` instead?

client/app/components/Input/InputTag.jsx
julianguyen: Why exactly are you calling `.slice(0, inputLength)` on `checkbox.label.toLowerCase()`? Shouldn't using `checkbox.label.toLowerCase()` be enough?

julianguyen: So the issue is that `checkbox.label.toLowerCase().slice(0, inputLength) === inputValue)` returns a value of type `Checkbox[]`.

`this.getSuggestions()` is being used to update the `this.state.suggestions`, which is expecting a value of type `string[]`.
```
onSuggestionsFetchRequested = ({ value }: {value: string}) => {
    this.setState({
      suggestions: this.getSuggestions(value)
    });
  };
```

SonuToor: Yep you're definitely right about changing this.state.suggestions to type Checkbox[]. That solves the other errors. But the following error remains:

Missing type annotation for T. T is a type parameter declared in array type [1] and was implicitly instantiated at
call of method filter [2].

I am reading some through other issues that people have had. Hopefully I'll run into a suitable solution soon :)

SonuToor:
> I don't think we need this new `suggestions` state, we can pass in `checkboxes`. The [documentation](https://github.com/moroshko/react-autosuggest#suggestions-required) for this library says it just has to be an array.

But the checkboxes that we have stored in state are all the options available, the suggestions array in state is a modified version of checkboxes that changes based on user input.

client/app/components/Input/InputTag.jsx
SonuToor: @julianguyen this error has been resolved!

SonuToor:
> It seems I am still failing some tests in CircleCI.
> 
> FAIL app/widgets/Resources/**tests**/Resources.spec.jsx
> FAIL app/components/Input/**tests**/InputTag.spec.jsx
> 
> These are the two places the tests failed.
> 
> I am having a bit of trouble deciphering what's going on here. Looks like I have some reading of Docs to do.
> 
> Any help or pointers would be greatly appreciated!



I've had a look and from what I can tell these tests fail because after the refactoring I did, the class names 'tagMenu' and 'tagLabel' aren't being used. They were assigned in the following two functions that aren't in the code anymore because react-autosuggest uses different functions.

`  renderMenu = (items: any) => (
    <div className={`tagMenu ${items.length ? css.tagMenu : ''}`}>{items}</div>
  );

  renderItem = (checkbox: Checkbox, highlighted: boolean) => (
    <div
      key={checkbox.id}
      className={`tagLabel ${highlighted ? css.tagHighlighted : ''} ${
        css.tagLabel
      }`}
    >
      {checkbox.label}
    </div>
  );
 `

These class names were used for the styling in InputTag.scss, but react-autosuggest uses it's own generated class names to style the Autosuggest component. For example the styling that 'tagMenu' applied previously is now to be applied through the class 'suggestionsContainerOpen'. 

@julianguyen what would be the appropriate next steps here? I could still add those class names to their respective places but that feels like a cheap workaround. Any ideas?

julianguyen:
Sorry for the late reply, I've been traveling. I think you'll want to update the test. It's fine giving those elements the original classNames if that's easier.

SonuToor:
> Sorry for the late reply, I've been traveling. I think you'll want to update the test. It's fine giving those elements the original classNames if that's easier.

No problem at all, hope you're enjoying the vacay!  

I have tried giving them the original classNames, it passes the previous tests but now fails at 2 tests. The two tests are identical.

`    Method “simulate” is meant to be run on 1 node. 0 found instead.
             .find('.tagLabel')
             .at(0)
    >        .simulate('click');
         |        ^
           expect(wrapper.find(id).exists()).toEqual(true);
         });`

So it is when the tests run the method simulate that it fails. I'll try asking for some help in the slack chat.

client/app/components/Input/__tests__/InputTag.spec.jsx
```diff
@@ -8,35 +8,35 @@ describe('InputTag', () => {
...
+    // const value = { value: 'Two' };
     // Hack: Looks like react-autocomplete is looking for target, but we should use currentTarget
-    wrapper.find('.tagAutocomplete').prop('onChange')({
-      target: value,
-      currentTarget: value,
-    });
+    // wrapper.find('.tagAutocomplete').prop('onChange')({
+    //   target: value,
+    //   currentTarget: value,
+    // });
```
julianguyen: Why is this commented out? We should still be testing for this.

client/app/components/Input/__tests__/InputTag.spec.jsx
```diff
@@ -8,35 +8,35 @@ describe('InputTag', () => {
...
     expect(wrapper.find(id).exists()).toEqual(true);
     wrapper.find(id).prop('onChange')({ currentTarget: { checked: false } });
     wrapper.update();
     expect(wrapper.find(id).exists()).toEqual(false);
     // Hack: Looks like react-autocomplete is looking for target, but we should use currentTarget
-    wrapper.find('.tagAutocomplete').prop('onChange')({
-      target: value,
-      currentTarget: value,
-    });
+    // wrapper.find('.tagAutocomplete').prop('onChange')({
```
julianguyen: Similar question as above!

spec/features/user_creates_a_draft_moment_spec.rb
```diff
@@ -35,7 +35,7 @@
       within('#moment_category_accordion') do
         find('.accordion').click
         find('.tagAutocomplete').set('Test Category')
-        page.find('.tagAutocomplete').native.send_keys(:return)
```
julianguyen: Hitting enter should still work. Why was this removed?

spec/features/user_creates_a_draft_strategy_spec.rb
```diff
@@ -37,7 +37,7 @@
       within('#strategy_category_accordion') do
         find('.accordion').click
         find('.tagAutocomplete').set('Test Category')
-        page.find('.tagAutocomplete').native.send_keys(:return)
```
julianguyen: Similar question as above.

spec/features/user_creates_a_published_moment_spec.rb
```diff
@@ -81,11 +81,11 @@
       within('#moment_viewers_accordion') do
         find('.accordion').click
         find('.tagAutocomplete').set('Ally 0')
-        page.find('.tagAutocomplete').native.send_keys(:return)
```
julianguyen: Similar question as above.

spec/features/user_creates_a_published_strategy_spec.rb
```diff
@@ -49,11 +49,11 @@
       within('#strategy_viewers_accordion') do
         find('.accordion').click
         find('.tagAutocomplete').set('Ally 0')
-        page.find('.tagAutocomplete').native.send_keys(:return)
```
julianguyen: Similar question as above.

client/app/components/Input/__tests__/InputTag.spec.jsx
SonuToor: Hey Julia, I think I've figured what I need to do here. I am still working on it but home to put up for this test soon enough.

spec/features/user_creates_a_published_strategy_spec.rb
SonuToor: This relates to the issue I'm having with the form submit within the 'viewers_accordion'

client/app/components/Input/__tests__/InputTag.spec.jsx
SonuToor: So I reckon I have the answer to it, the test I have instead of the commented out test is:

`describe('InputTag', () => {
  it('type in value and select it', () => {
    const wrapper = mount(component);
    const id = `#${InputMocks.inputTagProps.checkboxes[1].id}`;
    const value = { value: 'Two' };
    expect(wrapper.find('.tagMenu').exists()).toEqual(true);
    wrapper.find('.tagAutocomplete').simulate('change', { target: value });
    wrapper.find('.tagAutocomplete').simulate('focus');
    expect(wrapper.find(id).exists()).toEqual(false);
    wrapper
      .find('.tagLabel')
      .at(0)
      .simulate('click');
    expect(wrapper.find(id).exists()).toEqual(true);
  });

  it('un-check existing value and retype and select it', () => {
    const wrapper = mount(component);
    const id = `input#${InputMocks.inputTagProps.checkboxes[0].id}`;
    const value = { value: 'One' };
    expect(wrapper.find(id).exists()).toEqual(true);
    wrapper.find(id).prop('onChange')({ currentTarget: { checked: false } });
    wrapper.update();
    expect(wrapper.find(id).exists()).toEqual(false);
    wrapper.find('.tagAutocomplete').simulate('change', { target: value });
    wrapper.find('.tagAutocomplete').simulate('focus');
    expect(wrapper.find('.tagMenu').exists()).toEqual(true);
    expect(wrapper.find(id).exists()).toEqual(false);
    wrapper
      .find('.tagLabel')
      .at(0)
      .simulate('click');
    expect(wrapper.find(id).exists()).toEqual(true);
  });
});`

client/app/components/Input/__tests__/InputTag.spec.jsx
```diff
@@ -9,13 +9,9 @@ describe('InputTag', () => {
...
     const id = `#${InputMocks.inputTagProps.checkboxes[1].id}`;
     const value = { value: 'Two' };
-    // Hack: Looks like react-autocomplete is looking for target, but we should use currentTarget
-    wrapper.find('.tagAutocomplete').prop('onChange')({
-      target: value,
-      currentTarget: value,
-    });
-    wrapper.find('.tagAutocomplete').simulate('focus');
     expect(wrapper.find('.tagMenu').exists()).toEqual(true);
+    wrapper.find('.tagAutocomplete').simulate('change', { target: value });
```
SonuToor: So this is the main change that has been made to the tests. This provides the text input (.tagAutocomplete') with the value which in this case is 'Two'. This replaces the below code:

```
    wrapper.find('.tagAutocomplete').prop('onChange')({	
      target: value,	
      currentTarget: value,	
    });	
    wrapper.find('.tagAutocomplete').simulate('focus');
```
This code causes the tests to fail because react-autosuggest doesn't render the suggestion based on the value, hence find('.tagLabel') can't find anything because there's no suggestion. 

In essence the changes I made do the same thing as before, once the input changes to 'Two' the onChange method is fired and react-autosuggest provides the appropriate suggestion.

Let me know what you think!

client/app/components/Input/__tests__/InputTag.spec.jsx
julianguyen: I think your proposed change should be fine! Thanks for elaborating and yay for figuring it out.

client/app/components/Input/__tests__/InputTag.spec.jsx
julianguyen: This change works and is an improvement actually! :) Thanks for the update.

spec/features/user_creates_a_draft_moment_spec.rb
```diff
@@ -77,6 +77,10 @@
         find('#submit').click
       end
 
+      within('#moment_strategy_accordion') do
+        find('.accordion').click
```
julianguyen: Could you explain why you added this? We open the accordion but we aren't testing for anything.  I think this can be removed.

SonuToor:
> Looking good so far! Just some remaining feedback before we can merge! Great work! :)

Awesome I'll take a look. Thanks for being so patient with this PR Julia, I've honestly learned LOADS and super excited to my first real contribution to open source.

spec/features/user_creates_a_draft_moment_spec.rb
SonuToor: Hey @julianguyen  so I had a look!

This code snippet closes the accordion that was opened on line 68. 

The test would still pass without this code as well, the accordion could be left open it makes no difference to the behaviour, but if you look at line 47 and line 63 the previously opened accordions are closed in the exact same way. 

So do you want me to remove it or should I just leave it in?

julianguyen: I think it's safe to remove since we're not testing anything related to the accordion being closed!

SonuToor: Done :)

SonuToor:
It seems to be failing the flaky test intermittently @julianguyen

spec/features/user_creates_a_draft_moment_spec.rb
```diff
@@ -71,12 +71,12 @@
         find('.tagAutocomplete').set('Some New Strategy')
         page.find('.tagAutocomplete').native.send_keys(:return)
       end
-
+
```
julianguyen: Can we remove the unwanted whitespace but keep the newline?

spec/features/user_creates_a_draft_moment_spec.rb
```diff
@@ -71,12 +71,12 @@
...
         page.find('.tagAutocomplete').native.send_keys(:return)
       end
-
+      
       within '.modal' do
         fill_in_textarea('A Strategy Description', '#strategy_description')
         find('#submit').click
       end
-
+
```
julianguyen: Same comment as above.

spec/features/user_creates_a_draft_moment_spec.rb
```diff
@@ -98,7 +98,7 @@
       expect(page).to have_content 'What strategies would help?'.upcase
       expect(page).to have_content 'Test Strategy'
       expect(page).to have_content 'Some New Strategy'
-      find('.storyActionsViewers').hover
+      find('.storyActionsViewers').hover
```
julianguyen: Can we remove the unwanted whitespace after `hover`?

SonuToor:
> Excellent work! Thanks so much for working on this! 😎
> 
> There's a few stylistic changes that need to be made. After that, it should be ready to merge!

Done!

PR #1594, opened 2019-10-20
Adds development setup for Docker

sbpipb:
# Description
Adds a development setup for Docker

## More Details
- to improve convenience in development setups
- Docker images are _smaller_
  Postgres is down to 36mb from 200mb+
  - ifme_db_1    postgres     9.6-alpine   081f71293264   36.1 MB
- Usage of docker Volumes for Bundle and Node caches, and DB persistence.

### Caveats on Images
This setup is opinionated to my convenience.
- opted Ruby Slim


sbpipb:
> Can you replace or combine this with the existing Docker files? We don't need to designate a dev version of these files, though your idea to do so is a good one!

correct me if i'm wrong, we have the other Dockerfile for the CI right?

sbpipb:
The only thing that I could remove/replace is the docker compose override file

julianguyen:
We don't actually use the existing Dockerfile for CI, though we should!

julianguyen:
I think you'll want to keep the original files and make your changes in them!

sbpipb:
done with the adjustments

PR #1609, opened 2019-11-24
updated imports and functions to resolve eslint errors

dhuang612:
worked on form component and quickCreate components to remove cycle dependency.


dhuang612:
reviewed circleci tests that failed:
./spec/features/toggle_locale_spec.rb:28 
./spec/features/user_creates_a_published_moment_spec.rb:19
spec ./spec/features/user_creates_a_published_strategy_spec.rb:17

dhuang612:
So I see the circleci tests failed again

When I run npm run test:circleci on my local machine it passes.
Screen Shot 2019-11-25 at 1 25 01 PM

Please let me know if there is anything I need to resolve on my side.

client/app/components/Form/index.jsx
```diff
@@ -128,14 +126,14 @@ export class Form extends React.Component<Props, State> {
...
-  displayQuickCreate = (input: QuickCreateProps) => {
+  displayQuickCreate = (input: any) => {
     const {
       id, name, label, placeholder, checkboxes, formProps,
     } = input;
     if (!checkboxes || !name || !label) return null;
     return (
       <div key={id}>
-        <QuickCreate
+        <Checkbox
```
julianguyen: It looks like there are legitimate failing tests! I believe it's because you replaced `QuickCreate` with `Checkbox`. This component needs to be `QuickCreate` - curious why you swapped it.

dhuang612: because from my understanding the final refactor involves all three of the remaining files listed under the original issue:
**form (index.jsx)**
**dynamic form**
and **QuickCreate (index.jsx)**

index.jsx form

requires in quickCreate
->

dynamicForm
requires form

quickCreate
requires dynamicForm

julianguyen: So QuickCreate should be there instead of Checkbox, it's a custom component that utilizes Checkbox. I think we'll want to extract this method into another file to resolve your cyclical imports issue. Maybe put it in `ifme/client/app/components/Form/QuickCreate.jsx`?

dhuang612: Hi,

I wanted to confirm for this.
What you want me to do is under *form index.jsx*

move this function
```
displayQuickCreate = (input: QuickCreateProps)
```
into it's own file under Form/QuickCreate.jsx?

dhuang612: okay, updated the pr to most recent updates. 

I've moved displayQuickCreate to it's own page.

This has moved the error to be between
**index.jsx form** & **index.jsx quickCreate**

I'm trying the same idea of moving 
```
displayQuickCreateForm function

```

to a separate page from under **index.jsx quickCreate**

currently getting the following error:

```
Cannot get this.props because property props is missing in global object [1].

      4│ import type { Props as QuickCreateProps } from './index';
      5│
      6│ export const displayQuickCreateForm = (nameValue: string) => {
 [1]  7│     const { formProps } = this.props;
```
I'll continue to investigate this week!

dhuang612:
This is **QuickCreate** under the Form component

```
// @flow
import React from 'react';
import { QuickCreate } from '../../widgets/QuickCreate';
import type { Props as QuickCreateProps } from '../../widgets/QuickCreate';

export const displayQuickCreate = (input: QuickCreateProps) => {
  const {
    id, name, label, placeholder, checkboxes, formProps,
  } = input;
  if (!checkboxes || !name || !label) return null;
  return (
    <div key={id}>
      <QuickCreate
        id={id}
        name={name}
        label={label}
        placeholder={placeholder}
        checkboxes={checkboxes}
        formProps={formProps}
      />
    </div>
  );
};

```
I am now calling this into index.jsx under Form folder. 

I'm now trying to use this displayQuickCreate as a centralized location to break the cycle dependency. 


However, I'm getting errors when trying this under
QuickCreate component

and under the DynamicForm component

dhuang612:
Hi, 

I am working on this issue again. 

The next steps I am trying is to move displayQuickCreateForm function from out of QuickCreate component and into it's own file. I'm running into flow errors while trying to complete this 
image
Doesn't matter if I try to move this into its own file within the Form or QuickCreate folders.

Investigating into this.

julianguyen:
Hey! Sorry, I didn't get a chance to review this during this weekend. I'm going to check out your branch and play around!

dhuang612:
Thank you! no worries 
: )

client/app/widgets/QuickCreate/index.jsx
```diff
@@ -4,10 +4,9 @@ import { Modal } from '../../components/Modal';
 import { Input } from '../../components/Input';
 import type { Checkbox } from '../../components/Input/utils';
 import { Utils } from '../../utils';
-// eslint-disable-next-line import/no-cycle
 import css from './QuickCreate.scss';
 // eslint-disable-next-line import/no-cycle
-import { DynamicForm } from '../../components/Form/DynamicForm';
+import { DynamicForm} from '../../components/Form/DynamicForm';
```
julianguyen: Looks like you'll want to do `import { DynamicForm }` instead of `import { DynamicForm}` since this will raise a linting error i.e.

```
/home/circleci/ifmeorg/ifme/client/app/widgets/QuickCreate/index.jsx
  9:21  error  A space is required before '}'  object-curly-spacing

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

julianguyen: If you run `yarn lint` in the client folder locally, it will fix things for you automatically!

dhuang612:
Hi Julia,

So I am updating this pr with the latest local code. When I run eslint after separating out the quickCreate and removing the no-cycle comments it fails with the dependency error on my local machine

julianguyen:
I'll check out your latest changes!

julianguyen:
Oof yeah you're right about how creating the separate `client/app/components/Form/QuickCreate.jsx` doesn't help.

julianguyen:
So it looks like `client/app/components/Form/DynamicForm.jsx` is the main culprit for the cycle.

So `client/app/components/Form/index.jsx` is the parent component and it's using `client/app/widgets/QuickCreate/index.jsx` when `type="quickCreate"`. `client/app/widgets/QuickCreate/index.jsx` is using `client/app/components/Form/DynamicForm.jsx` which imports `import { Form } from './index';`. To break the cycle we have to find another way to use `DynamicForm`.

There are two nested forms in this component. 

I think we'll want to do an overhaul of `client/app/components/Form/index.jsx` and separate it into two distinct components, one where we want to include the form tag and one where we don't want to (DynamicForm).

Let me know if this makes sense. Sorry for misleading you by telling you to create another QuickCreate component.

julianguyen:
To truly separate `Form` and `DynamicForm`, I think we want to extract `displayInputs`, `displayQuickCreate`, `displayInput`, `isInputError`, `handleError` into a service where we can use it in both `Form` and `DynamicForm`.  `handleNoFormTagSubmit` can then go in `DynamicForm`.

Then you can separate the `render()`:

```
render() {
    const { action, noFormTag, noFormTagRef } = this.props;
    if (noFormTag) {
    // THIS WOULD GO IN DYNAMICFORM
      return (
        <div className={css.form} ref={noFormTagRef}>
          {this.displayInputs()}
        </div>
      );
    }
   // THIS WOULD STAY IN THIS FILE
    if (!action) return null;
    return (
      <form
        onSubmit={this.onSubmit}
        acceptCharset="UTF-8"
        className={css.form}
        method="post"
        action={action}
      >
        {this.displayInputs()}
      </form>
    );
  }
```

dhuang612:
```
To truly separate Form and DynamicForm, I think we want to extract displayInputs, displayQuickCreate, displayInput, isInputError, handleError into a service
```

When you say service, do you mean a component?

julianguyen:
It doesn't need to be a React component. I think it should be a service so you can export the functions and use them in Form and DynamicForm.

dhuang612:
thanks, I'll take a look at separating out the functions mentioned above.

julianguyen:
Thanks so much for taking the time to work on this! Sorry again for the misleading advice earlier. I forgot how complicated things were.

dhuang612:
Here's what I have so far:

I've moved the functions: 
**displayInput** & **displayInputs** into it's own file inside of widgets:
under a new folder called Input -> index.jsx


```
// @flow
import React from 'react';
import { TYPES as INPUT_TYPES } from '../../components/Input/utils';
import { Input } from '../../components/Input';
import type {State} from '../../components/Form/index'
import type { Props as InputProps } from '../../components/Input/utils';
import { displayQuickCreate } from '../../components/Form/quickCreate';
import { any } from 'bluebird';

type KeyProps = { myKey?: any };
type MyInputProps = InputProps & KeyProps;

type DisplayInputType = {
    noFormTag: boolean,
}

type InputState = {
  State: State
}

export type Props ={
  inputs: any[],
  noFormTag?: boolean, // Can't have nested forms i.e. quick create
}

const displayInput = ({input: MyInputProps, noFormTag: DisplayInputType}) => {
    const { noFormTag } = DisplayInputType;
    return (
      <div key={input.id}>
        <Input
          id={input.id}
          key={input.myKey}
          type={input.type}
          name={input.name}
          label={input.label}
          placeholder={input.placeholder}
          error={input.error}
          value={input.value}
          readOnly={input.readOnly}
          copyOnClick={input.copyOnClick}
          disabled={input.disabled}
          required={input.required}
          info={input.info}
          minLength={input.minLength}
          maxLength={input.maxLength}
          dark={input.dark}
          large={input.large}
          checked={input.checked}
          uncheckedValue={input.uncheckedValue}
          options={input.options}
          checkboxes={input.checkboxes}
          accordion={input.accordion}
          onClick={
            input.type === 'submit' && noFormTag
              ? this.handleNoFormTagSubmit
              : undefined
          }
          onError={input.type !== 'submit' ? this.handleError : undefined}
          myRef={(element) => {
            this.myRefs[input.id] = element;
          }}
          formNoValidate={input.type === 'submit'}
        />
      </div>
    );
  };

export const displayInputs = ({State: InputState}) => {
    const { inputs } = InputState;
    return inputs.map((input: any) => {
      if (INPUT_TYPES.includes(input.type)) {
        return displayInput(input);
      }
      if (input.type === 'quickCreate') {
        return displayQuickCreate(input);
      }
      return null;
    });
  };
```
I'm having a problem with flow for the second function displayInputs
in where I can't get the state for inputs imported into this file.

I can see under root Form index.jsx there's an import for state:
```
export type State = {
  inputs: any[],
  errors: Errors,
};
```

Methods I've tried
directly importing in the export from Form -> index.jsx
ontop of the file inside of widgets -> Inputs -> index.jsx
```
import type {State} from '../../components/Form/index'

type InputState = {
  State: State
}

export const displayInputs = ({inputs: State}) => {
    const { inputs } = State;
```
this gives me an error about State not being defined.


Second method:
restating the state inside of widgets -> input -> index.jsx

```
type State = {
  inputs: any[]
}

export const displayInputs = ({inputs: State}) => {
    const { inputs } = State;

Missing type annotation for destructuring.
Property inputs is missing in undefined [1].
```

I was able to resolve the props declaration under displayInput by importing it
```
type DisplayInputType = {
    noFormTag: boolean,
}
and then using it within the function:
const displayInput = ({input: MyInputProps, noFormTag: DisplayInputType}) => {
    const { noFormTag } = DisplayInputType;

```
But it doesn't seem to be working the same for state.

dhuang612:
@julianguyen 
Hey Julia,

I am having problems trying to get the state set within the function displayInputs, could I have you take a look?
**inputs.jsx under widgets folder** 
```
import type { State as InputState} from '../../components/Form/index';


export const displayInputs = (): any => {
    const { inputs } = MyInputState;

```

Thank you

julianguyen:
I'll take a look tonight! Sorry I didn't get a chance to this weekend - was in transit.

julianguyen:
Looks like `MyInputState` should be `InputState` since that's what you namespaced `State` to be in the import. I recommend debugging by inserting a `console.log(InputState)` to inspect.

dhuang612:
Hey @julianguyen 

So the current error I'm getting is related to flow and importing in the state.
```
Cannot reference type InputState [1] from a value position.
from this line:
const { inputs } = InputState;
```
I think I need to make this into a component to help with the movement of state?

dhuang612:
@julianguyen 
I've gone back and reviewed all the notes up to this point, and I believe I have a bit better understanding of next steps. 

I am currently working on splitting up Form index.jsx into two files
index.jsx and FormwithTag

index.jsx -> dynamicForm
FormwithTag -> default form

still need to finish moving out the functions from under index.jsx
also still seeing the no cycle warning because index.jsx -> dynamicForm -> quickCreate

dhuang612:
Also, won't this just move the dependency into DynamicForm?
@julianguyen 
since the cycle will then be:
(stand alone function that imports in widgets -> quickCreate ) -> quickCreate -> dynamicForm -> widgets quickCreate

dhuang612:
red lines are the cycle dependency:
Screen Shot 2019-12-12 at 5 29 27 PM

dhuang612:
I think this pr might finally be complete?

PR #1618, opened 2019-12-23
locales:add translation for new resource tag - "physical dependence"

andy5995:
# Description

Adding English translation for the "physical dependence" tag.

@julianguyen , I missed this in #1616 Is a test needed to catch this in the future?

Right now there's a message on https://www.if-me.org/resources about the missing translation.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -569,6 +569,7 @@ de:
         stories: geschichten
         teen: jugendlich
         coaching: coaching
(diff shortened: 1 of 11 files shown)


andy5995:
> 
> @julianguyen , I missed this in #1616 Is a test needed to catch this in the future?
> 

I created ticket #1619 to open that up for more discussion if needed.

andy5995:
I think I got all the translations and this is ready for review.

Issue #1345, opened 2019-01-23
Requesting Chinese Translation


isabelleyiu:
# Description 

Translate app to Chinese

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #1634, opened 2020-01-16
chinese translation

Manasa2850:
# Description
< Added files for chinese translation>

## Corresponding Issue

< https://github.com/ifmeorg/ifme/issues/1345 >

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -27,6 +27,7 @@ de:
     de: Deutsch
     fr: Français
     hi: 'हिंदी'
+    ch: 中文
   navigation:
     main_menu: 'Hauptmenü der Site'
     user_menu: 'Benutzermenü'
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -27,7 +27,7 @@ de:
     de: Deutsch
     fr: Français
     hi: 'हिंदी'
-    ch: 中文
+    zh: 中文
   navigation:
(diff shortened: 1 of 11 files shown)


Manasa2850:
@julianguyen Yes I will be continuing it.

config/locales/de.yml
```diff
@@ -27,6 +27,7 @@ de:
     de: Deutsch
     fr: Français
     hi: 'हिंदी'
+    ch: 中文
```
julianguyen: I looked up the language code for Chinese and it's actually `zh`.

Also what do you think of specifying whether it's Simplified or Traditional?


Chinese | zh
Chinese (Simplified) | zh-Hans
Chinese (Traditional) | zh-Hant

Manasa2850: @julianguyen sure I'll do that!
I also need separate yml files (like ch.yml) for the three right?

julianguyen: Yeah you'll want to create the following [here](https://github.com/ifmeorg/ifme/tree/master/config/locales):
* zh.yml
* devise.zh.yml
* devise_invitable.zh.yml
* zh.yml
* kaminari.zh.yml

Not sure if you've already seen our [Translations guide](https://github.com/ifmeorg/ifme/wiki/Translations)!

Manasa2850: @julianguyen Thanks! I'll make changes and submit the final PR in a day or two!

Issue #1635, opened 2020-01-16
some dead links should be removed


andy5995:
Some of these dead links should be removed, changed, or double-checked.

![image](https://user-images.githubusercontent.com/16764864/72563495-6da2c600-3873-11ea-9ca1-babfe0151133.png)

PR #1636, opened 2020-01-19
Remove dead links

GayathriVenkatesh:
# Description

Removed broken links. 

## Corresponding Issue

Issue #1635 
# Screenshots
Screenshot from 2020-01-19 10-31-20

---


app/views/pages/about.html.erb
```diff
@@ -40,7 +40,7 @@
 
 <div class="messageBody">
   <%= raw t('pages.about.message_text_three', {
-    linebreak_link: link_to(t('pages.about.linebreak'), 'https://linebreak.co', target: 'blank'),
+    linebreak_link: link_to(t('pages.about.linebreak'), 'http://linebreak.io', target: 'blank'),
```
julianguyen: Looks like this is a different Linebreak! Maybe they shut down? In that case, I would update this translation key (`pages.about.linebreak`) so that you no longer have to pass in a link!

app/views/layouts/application.html.erb
```diff
@@ -4,7 +4,7 @@
     <title><%= render 'layouts/title' %></title>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <link rel="manifest" href="manifest.json">
+    <link rel="manifest" href="/manifest.json">
```
julianguyen: Did you double-check that this links correctly?

GayathriVenkatesh: I was actually not sure about this one.
When I ran the command "blc http://localhost:3000 -ro", it didn't show any broken link (like the above screenshot).
How do I correct it?

app/views/pages/about.html.erb
GayathriVenkatesh: Is it ok if I change the statement to the following:-
linebreak_link: t('pages.about.linebreak')
So now there is no link, just the "Linebreak" text.

app/views/layouts/application.html.erb
julianguyen: We don't need to change this link! I checked and what's already there works in production.

app/views/pages/about.html.erb
julianguyen: Exactly - we just want the Linebreak text!

app/views/pages/about.html.erb
```diff
@@ -40,7 +40,7 @@
 
 <div class="messageBody">
   <%= raw t('pages.about.message_text_three', {
-    linebreak_link: link_to(t('pages.about.linebreak'), 'https://linebreak.co', target: 'blank'),
+    linebreak_link: t('pages.about.linebreak'),
```
julianguyen: Nice! In addition to this, let's also update the translation key itself. 

We'll want to replace `%{linebreak_link}` in this [file and other equivalent translation files](https://github.com/ifmeorg/ifme/blob/master/config/locales/en.yml#L394) with just Linebreak. Then we can delete this [line](https://github.com/ifmeorg/ifme/blob/master/config/locales/en.yml#L395).

GayathriVenkatesh: I have updated the .yml files accordingly :)

Issue #1638, opened 2020-01-19
Automatically generate Refill date for new medication


GayathriVenkatesh:
# Description
While adding a new medication, the refill field is mandatory. 
However, given the total no. of tablets and dosage per day, the refill date can be generated in the backend (rather than having the user calculate it). This date can be the default value for the field, and can be manually changed if necessary.

# Screenshots

![Screenshot from 2020-01-19 20-38-09](https://user-images.githubusercontent.com/32172764/72683428-4eb76480-3afd-11ea-94da-7f0e3d567226.png)

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

GayathriVenkatesh:
@julianguyen Do share your thoughts on this :)

julianguyen:
Hey, @GayathriVenkatesh this is a really good observation and I think we should implement this feature! :D Thank you for playing around with it and suggesting this!

GayathriVenkatesh:
Hi @julianguyen, could you tell me the command I need to use to access the input values of Total and Dosage? I tried a couple of things but it doesn't seem to work.
Here, when I try to print medication_field('total'), this is what gets displayed:

![Screenshot from 2020-01-21 09-38-25](https://user-images.githubusercontent.com/32172764/72774871-3e54d600-3c32-11ea-847f-a0d0ce1be308.png)

The value field seems to be nil even after entering a value.
What am I doing wrong?

PR #1654, opened 2020-02-03
Automatically Generate Refill Date

GayathriVenkatesh:
# Description
Added feature to automatically fill in medication refill date, if not specified.

## Corresponding Issue
#1638 

# Screenshots

Screenshot from 2020-02-02 11-34-43

In this example, refill date is not filled in by the user. So it defaults to the second-last day of the dosage.

Screenshot from 2020-02-02 11-35-39

<br><br><br>
Screenshot from 2020-02-02 11-38-33

Here, the date is explicitly specified by the user, so it remains as it is.

Screenshot from 2020-02-02 11-39-03

---

### DIFF
--- a/app/controllers/medications_controller.rb
+++ b/app/controllers/medications_controller.rb
@@ -46,15 +46,16 @@ class MedicationsController < ApplicationController
   end
 
   def add_refill(medication)
-    return medication[:refill] unless medication[:refill].empty?
+    return medication[:refill] if medication[:refill].present?
 
-    time = Time.zone.now.strftime('%d/%m/%Y')
-    weekly_dosage = medication[:weekly_dosage] || [0, 1, 2, 3, 4, 5, 6]
-    no_of_days = medication[:total].to_i / medication[:dosage].to_i
+    weekly_dosage = @medication&.weekly_dosage ||
+                    medication[:weekly_dosage] || [0, 1, 2, 3, 4, 5, 6]
+    total = @medication&.total || medication[:total]
+    dosage = @medication&.dosage || medication[:dosage]
+    no_of_days = total.to_i / dosage.to_i
     no_of_weeks = no_of_days.to_i / weekly_dosage.length
-    rem = no_of_days % weekly_dosage.length
-    days = 7 * no_of_weeks + rem
-    time.to_date + days - 1
+    days = 7 * no_of_weeks + (no_of_days % weekly_dosage.length)
+    Time.zone.now.strftime('%d/%m/%Y').to_date + days - 1
   end
 
   # POST /medications
@@ -78,7 +79,9 @@ class MedicationsController < ApplicationController
   def update
     return unless save_refill_to_google_calendar(@medication)
 
-    if @medication.update(medication_params)
+    if @medication.update(medication_params.merge(
+                            refill: add_refill(medication_params)
+                          ))
       redirect_to_medication(@medication)
     else
       render_unprocessable_medication
@@ -107,8 +110,7 @@ class MedicationsController < ApplicationController
     respond_to do |format|
       format.html { render :new }
       format.json do
-        render(json: @medication.errors,
-               status: :unprocessable_entity)
+        render(json: @medication.errors, status: :unprocessable_entity)
       end
     end
   end
@@ -122,8 +124,7 @@ class MedicationsController < ApplicationController
 
   def medication_params
     params.require(:medication).permit(
-      :name, :dosage, :refill,
-      :total, :strength,
+      :name, :dosage, :refill, :total, :strength,
       :dosage_unit, :total_unit, :strength_unit,
       :comments, :add_to_google_cal,
       take_medication_reminder_attributes: %i[active id],
--- a/app/controllers/medications_controller.rb
+++ b/app/controllers/medications_controller.rb
@@ -45,24 +45,25 @@ class MedicationsController < ApplicationController
     redirect_to_medication(@medication)
   end
 
-  def add_refill(medicine)
-    if medicine.refill.nil?
-      time = Time.zone.now.strftime('%d/%m/%Y')
-      no_of_days = medicine.total / medicine.dosage
-      no_of_weeks = no_of_days / medicine.weekly_dosage.length
-      rem = no_of_days % medicine.weekly_dosage.length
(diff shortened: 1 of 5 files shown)


GayathriVenkatesh:
Hi @julianguyen, I worked on issue #1638 and have added screenshots for the same. However, there are a few codeclimate errors in the code. Is there a workaround to this? I tried a few things out but was unable to condense the code any further.

app/controllers/medications_controller.rb
```diff
@@ -50,6 +50,9 @@ def edit
   def create
     @medication =
       Medication.new(medication_params.merge(user_id: current_user.id))
+
+    @medication.refill = @medication.refill.nil? ? Time.zone.now.strftime('%d/%m/%Y').to_date +
+      7 * ((@medication.total / @medication.dosage) / @medication.weekly_dosage.length) + ((@medication.total / @medication.dosage) % @medication.weekly_dosage.length) - 1 : @medication.refill
```
julianguyen: I would would break apart this logic into a helper function. That should reduce the complexity of this create function and better separate the logic you're adding.

julianguyen: I would use `if` and `unless` like Code Climate suggests instead of having multi-line ternary operators. That should also improve readability of the method.

You can run `rubocop` locally and it should throw these warnings.

GayathriVenkatesh: Thanks! I have fixed those issues, now there is no error in codeclimate. However, the total-coverage and diff-coverage checks have failed. I did research a bit but I couldn't pinpoint the error. Would be great if you could look into it!

app/controllers/medications_controller.rb
```diff
@@ -45,11 +45,24 @@ def edit
     redirect_to_medication(@medication)
   end
 
+  def add_refill(medicine)
```
julianguyen: We should also write a unit test for this function!

julianguyen:
Looks like we should have test coverage for this new feature! Let me know if you need any help writing tests :)

app/controllers/medications_controller.rb
GayathriVenkatesh: Hi @julianguyen, I am not quite familiar with writing tests for such a function. Would be great if someone could guide me through it!

julianguyen: What I would do this update this part of the [test](https://github.com/ifmeorg/ifme/blob/master/spec/controllers/medications_controller_spec.rb#L126) to ensure that `@medication.refill` is the expected value.

GayathriVenkatesh:
Hi @julianguyen, I updated the spec.rb file to ensure that the refill date is always before the last date of dosage. Could you please have a look and let me know if there's something I should change?

spec/controllers/medications_controller_spec.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-
+Rails::logger.debug "Interesting stuff"
```
julianguyen: We should remove this!

spec/controllers/medications_controller_spec.rb
```diff
@@ -143,6 +143,9 @@ def post_create(medication_params)
         it 'has no validation errors' do
           post_create valid_medication_params
           expect(assigns(:medication).errors).to be_empty
+  
+          expect(assigns(:medication)[:refill]).to be_between("01-01-0001".to_date,
+            ((7 * assigns(:medication)[:total]) / (assigns(:medication)[:dosage] * assigns(:medication)[:weekly_dosage].count)).days.from_now)
```
julianguyen: Is checking between 01-01-0001 and the number of days from when the days are calculated the best way of verifying this works? Why not set the created_at date to a value upon create and then check for the specific date in the test?

app/controllers/medications_controller.rb
```diff
@@ -45,11 +45,24 @@ def edit
...
+    if medicine.refill.nil?
```
julianguyen: We should also test when `medicine.refill` is not nill!

What also happens when we update a medication - can we change the refill values?

julianguyen:
The Codeclimate coverage failures mean that we should add more tests! I mentioned above that we should test the nil case.

spec/controllers/medications_controller_spec.rb
GayathriVenkatesh: Screenshot from 2020-03-06 13-26-03

Hi @julianguyen, I tried changing this part so that it checks whether the refill date is between date of creation of medication and the maximum possible date (as calculated). However this throws the above error. Could you please guide me further?

app/controllers/medications_controller.rb
GayathriVenkatesh: Updating medication works fine, no issues on that front :)

spec/controllers/medications_controller_spec.rb
julianguyen: Ah so this error is happening because the medication's refill is being set to `01/01/2020` in `spec/factories/factories.rb`. We use this factory to generate a medication in this test.

I tested this feature locally and it doesn't work because the front-end prevents us from submitting the form if the refill field is left blank.

julianguyen: To fix this, we'll just need to set this field as non-required (just in the front-end not back-end). I can push a commit to this PR to update this.

julianguyen: As for getting the test to pass, we'll need to call `add_refill` when we call `Medication.new` instead of doing it after the new medication entry is created. I'll also update your PR for you.

app/controllers/medications_controller.rb
julianguyen: So another important question to consider is whether we want add_refill to be called when updating a medication.

julianguyen: Actually, we'll have to call add_refill for updating. It's possible that the user would empty out the refill field in the form. I can update this too!

julianguyen:
Can you get rid of the two recent commits you added - it undoes some changes I made that we need.

julianguyen:
I am going to merge this if the tests pass and the security audit fails. Our master has the latest version of the vulnerable gems. Outside of this PR, you should rebase from our master in your forked master.

Issue #1665, opened 2020-02-19
Create moments_moods join table


leeacto:
# Description
Moments currently track moods as a serialized array of Mood IDs on the moments table. Per Slack, this was a legacy decision. A more efficient way to store the relationship would be through a join table. This implementation would eliminate N+1 queries when iterating over moments to display mood names.

Acceptance Criteria:
* Implement join table to relate Moments to Moods
* Create and execute rake task to backfill existing relationships

In order to seamlessly implement this feature, the current implementation should not be removed. Instead, we can create relations in both manners and remove the old method in a subsequent deployment.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #1669, opened 2020-02-20
Adds moment_moods join table

raulfernando:
# Description
Adds `moment_moods` join table for a more efficient way to store the relationship between `Moment` and `Mood` models.

## Corresponding Issue
Solves [#1665](https://github.com/ifmeorg/ifme/issues/1665)

### DIFF
--- a/app/models/moment_mood.rb
+++ b/app/models/moment_mood.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+# == Schema Information
+#
+# Table name: moment_moods
+#
+#  id                :bigint(8)        not null, primary key
+#  moment_id         :bigint
+#  mood_id           :bigint
+#
+
+class MomentMood < ActiveRecord::Base
+  belongs_to :moment
+  belongs_to :mood
+end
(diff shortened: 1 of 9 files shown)


leeacto:
I had already started working on this since I assigned the issue to myself (https://github.com/ifmeorg/ifme/pull/1670).

That being said, the PR looks good. I think the table name (according to Rails) would use plural nouns for both sides (moments_moods), though. See [here](https://stackoverflow.com/questions/11590469/rails-naming-convention-for-join-table)

app/models/moment_mood.rb
```diff
@@ -0,0 +1,14 @@
...
+# Table name: moment_moods
+#
+#  id                :bigint(8)        not null, primary key
+#  moment_id         :bigint
+#  mood_id           :bigint
+#
+
+class MomentMood < ApplicationRecord
+  belongs_to :moment
+  belongs_to :mood
```
leeacto: You'll also want some uniqueness validation to make sure that there is no duplication of records with the same moment and mood ID.

raulfernando:
I'm closing this PR because @leeacto is working on an equal implementation

PR #1670, opened 2020-02-20
Moment moods

leeacto:
# Description

Creates relation between Moments and moods.
Current implementation is still available so that information won't be lost.
Created a rake task to backfill existing relationship.
After backfill is done, a follow up PR will be required to remove implementation of old method.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/1665

---

### DIFF
--- a/app/models/moment.rb
+++ b/app/models/moment.rb
@@ -60,18 +60,10 @@ class Moment < ApplicationRecord
   end
 
   def self.populate_moments_moods
-    sql = 'INSERT INTO moments_moods (moment_id, mood_id) VALUES '
-    insertions = []
     Moment.all.find_each do |moment|
-      moment.mood.each do |mood_id|
-        insertions << "(#{moment.id}, #{mood_id})"
-      end
+      moment.mood = Mood.where(id: moment.mood).pluck(:id)
+      moment.save
     end
-
-    sql += insertions.join(', ')
-    sql += ' ON CONFLICT DO NOTHING;'
-
-    Moment.connection.execute(sql)
   end
 
   def category_array_data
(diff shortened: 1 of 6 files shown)


app/models/moment.rb
```diff
@@ -57,6 +59,13 @@ def self.find_secret_share!(identifier)
     )
   end
 
+  def self.populate_moments_moods
+    Moment.all.find_each do |moment|
+      moment.mood = Mood.where(id: moment.mood).pluck(:id)
```
julianguyen: Can the `where` be replaced with `find`?

leeacto: no because there may be many moods per moment. `find` will only get one.
This is a case that will look like `moment.mood = Mood.where(id: [1,2,3]).pluck(:id)` and create many Moods for one Moment.

julianguyen: Ah yes! LGTM then!

leeacto:
Merging now. Upon deployment the new migration must be run to create the join table. Also, you can run `rake populate_moments_moods` to backfill existing moment moods.

julianguyen:
@leeacto Deployed, ran both `rake db:migrate` and `rake populate_moments_moods`. On the latter I got the following error:
 ```
NameError: uninitialized constant Moment
/app/Rakefile:49:in `block in <top (required)>'
/app/vendor/bundle/ruby/2.6.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/app/vendor/bundle/ruby/2.6.0/gems/bundler-1.17.3/exe/bundle:30:in `block in <top (required)>'
/app/vendor/bundle/ruby/2.6.0/gems/bundler-1.17.3/exe/bundle:22:in `<top (required)>'
/app/bin/bundle:3:in `load'
/app/bin/bundle:3:in `<main>'
Tasks: TOP => populate_moments_moods
(See full trace by running task with --trace)
```

Issue #1671, opened 2020-02-20
QuickCreate dropdowns not expanded when default checkboxes are checked


leeacto:
# Description
I noticed when editing a Moment that if categories/moods/strategies had been selected, they aren't visible until you click the caret to expand. From a UX perspective, I should be able to see what's been selected in order to add/update. Otherwise, I may not know or remember.

Updating the accordionOpen default state property to be conditional should fix this.

# Reproduction Steps
1. Create a Moment with a Mood assigned to it
2. Edit the Moment
3. Observe that the Mood dropdown showing the previously selected mood is not expanded

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #1672, opened 2020-02-20
[#1671] expand when pre checked

leeacto:
# Description
QuickCreate component's accordionOpen state property is closed by default. In the case that a user is editing an object, they may want to see all of the pre-selected items. For instance, a Moment with categories will not show pre-selected categories upon page load.

This change conditionally opens the accordion if any options have already been checked.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/1671

---

### DIFF
--- a/spec/features/user_creates_a_published_moment_spec.rb
+++ b/spec/features/user_creates_a_published_moment_spec.rb
@@ -116,21 +116,18 @@ describe 'UserCreatesAPublishedMoment', js: true do
       expect(find('.pageTitle')).to have_content 'Edit My New Moment'
 
       within('#moment_category_accordion') do
-        find('.accordion').click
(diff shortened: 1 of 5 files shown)


spec/features/user_creates_a_published_moment_spec.rb
```diff
@@ -91,7 +91,7 @@
 
       find('#moment_comment_switch').click
       find('#moment_publishing_switch').click
-      find('#submit').click
+      find('#submit', wait: 10).click
```
julianguyen: Is the `wait` necessary? I'd like to avoid using it unless absolutely needed.

leeacto: I'll take it out. It seemed like I got some failures before. let's see

PR #1674, opened 2020-02-21
Issue 1655 moment moods cleanup

leeacto:
# Description

Removes Moment#mood column and works with the join table relation instead. 
There's a lot of shared code and metaprogramming related to strategies and categories. Since the old field was named `mood` (singular) and the proper relation is `moods`, there is some code based around `mood` being examined. When other models get properly related, the conditional statements will go away.

---

### DIFF
--- a/app/helpers/most_focus_helper.rb
+++ b/app/helpers/most_focus_helper.rb
@@ -26,10 +26,10 @@ module MostFocusHelper
   def get_data_type(model_object, data_type)
     data = []
     model_object.select do |item|
-      related_objects = data_type == 'moods' ? item.moods.pluck(:id) : item[data_type]
-      next unless related_objects.any? && viewable?(item)
+      objs = data_type == 'moods' ? item.moods.pluck(:id) : item[data_type]
+      next unless objs.any? && viewable?(item)
 
-      data.concat(related_objects)
+      data.concat(objs)
     end
     data
   end
--- a/app/helpers/most_focus_helper.rb
+++ b/app/helpers/most_focus_helper.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module MostFocusHelper
   def most_focus(data_type, user)
-    return unless user && %w[mood strategy category].include?(data_type)
+    return unless user && %w[moods strategy category].include?(data_type)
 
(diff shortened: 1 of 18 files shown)


julianguyen:
Looks like you'll want to rebase and resolve some conflicts before we merge.

app/models/concerns/common_methods.rb
```diff
@@ -3,10 +3,10 @@ module CommonMethods
   extend ActiveSupport::Concern
 
   def mood_names_and_slugs
-    return unless attribute(:mood)
+    return unless self.class.reflect_on_association(:moods)
```
julianguyen: Nice!

app/helpers/most_focus_helper.rb
```diff
@@ -26,9 +26,10 @@ def viewable?(item)
   def get_data_type(model_object, data_type)
     data = []
     model_object.select do |item|
-      next unless item[data_type].any? && viewable?(item)
+      objs = data_type == 'moods' ? item.moods.pluck(:id) : item[data_type]
```
julianguyen: We should eventually extract this into helper and generalize his for categories and strategies. This can be handled in a subsequent PR.

db/migrate/20200221153634_drop_moments_mood_column.rb
```diff
@@ -0,0 +1,5 @@
+class DropMomentsMoodColumn < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :moments, :mood
```
nshki: Likely won't be a big deal, but [dropping a column can cause exceptions during deployment](https://github.com/ankane/strong_migrations#removing-a-column) until the app is rebooted. If we want to avoid that, it's recommended that we queue up two migrations and deploy them separately.

julianguyen: Ooh great point!

julianguyen:
I think this should be safe to merge!

db/migrate/20200221153634_drop_moments_mood_column.rb
leeacto: Ahh I see. Thanks for the link.

Issue #1656, opened 2020-02-08
Re-write existing React components with new Hooks API


julianguyen:
# Description

We're currently on React 16.9.0, which means we can leverage the Hooks API. Let's re-write our components to use it where suitable.

I'm going to go ahead and divide the work so you can assign yourselves to it:

- [x]  client/app/components/Accordion/index.jsx [@RomanMaru]
- [x] client/app/components/BaseContainer/index.jsx [@thesid01]
- [x] client/app/components/Chart/ChartControl.jsx [@udayanshevade]
- [x] client/app/components/Form/DynamicForm.jsx AND client/app/components/Form/index.jsx [@alekpentchev]
- [x] client/app/components/Header/index.jsx [@alekpentchev]
- [x] client/app/components/Input/index.jsx AND client/app/components/Input/InputCheckboxGroup.jsx AND client/app/components/Input/InputLocation.jsx AND client/app/components/Input/InputSelect.jsx AND client/app/components/Input/InputSwitch.jsx AND client/app/components/Input/InputTag.jsx [@Orelongz]
- [x] client/app/components/Modal/index.jsx [@alekpentchev]
- [x] client/app/widgets/Comments/index.jsx [@gdemu13]
- [x] client/app/widgets/Notifications/index.jsx [@thesid01] 
- [x] client/app/widgets/QuickCreate/index.jsx [@MicahRamirez] 
- [x] client/app/widgets/Resources/index.jsx [@gdemu13] 

If there are any missing tests, please add them 🙏 

Please comment below and let me know which files you want to tackle. I'll update this comment accordingly!

Thanks everyone 🙌 

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

wstan2:
Can I help with this?

julianguyen:
@wstan2 Of course! Feel free to assign yourself to this issue :D Thanks!

Loriick:
hello, I assign myself to this issue , I hope it's okay ?

julianguyen:
Go head! Actually I think @Loriick and @wstan2 could split up the work! I recommend deciding that in this issue :)

RomanMaru:
Hello there, I'd like to help with this

[...]

PR #1686, opened 2020-03-06
Issue 1656 forms

alekpentchev:
# Description

Components - DynamicForm and Form together with their tests - were rewritten to support React Hooks API.

## Corresponding Issue

# Screenshots

image

---


alekpentchev:
One lint error exists because of eslint issue not resolved yet - https://github.com/yannickcr/eslint-plugin-react/issues/1280

This is also the reason why circleci fails.

julianguyen:
Looks like there's legitimate rspec failures related to form submission! Let me know if you need any help debugging.

julianguyen:
I did some refactors just now for duplicate code between Form and DynamicForm. Here's the merged [PR](https://github.com/ifmeorg/ifme/pull/1688). You'll need to rebase, let me know if you need any help!

alekpentchev:
@julianguyen It's super strange - these failing tests occured after adding ```// eslint-disable-next-line ``` before extracting ```action``` from ```props```.
After removing this comment tests fail with the old error that ```action is missing in props validation```.

alekpentchev:
For testing purposes I've added ```prop-types``` package to client's dependencies and defined propTypes for the Form component like this:
```
Form.propTypes = {
  action: PropTypes.string,
};
```
Eslint / Flow error about missing props in validation does not appear anymore. But I've mixed feelings about adding another dependency just to fix such strange and should-not-happen-at-all issue. What is your opinion on that topic?

julianguyen:
I think we could away from Flow in the long-term since PropTypes from React offers a lot of what Flow does. You could try doing this for this PR and then we can do the same elsewhere in another PR.

alekpentchev:
> I think we could move away from Flow in the long-term since PropTypes from React offers a lot of what Flow does. You could try doing this for this PR and then we can do the same elsewhere in another PR.

I see your point but I'm not sure whether moving away from Flow completely would be a good approach. In my opinion we'd lose important information about what types do state variables have, what type of data do functions return etc.

On the other hand I've asked a question on that matter (the issue with propTypes) directly in the eslint-react-plugin repo and I've received a such reply on the interactions between Flow types and propTypes:

> Both are required; neither flow nor typescript is capable of catching everything PropTypes can.

To sum up - I'll commit my changes containing added propTypes for the "problematic"component. We'll see then how circleci tests behave, too.

julianguyen:
Yeah, the trade-off is losing state type information :/. But I think it should fine. We should rely on native React features as much as possible.

Sounds like a good plan!

julianguyen:
I'm actually going to check out your branch and see if I can fix the linting issues without having to move away from Flow.

I think we can revisit the move to prop-types another time, outside of this pull request/task.

alekpentchev:
@julianguyen Thank you very much for your support. I was leaning towards your direction and your implementation looks fine. I will review for myself to use it with the other components assigned to me.

+ I agree that we should discuss Flow in another thread.

PR #1689, opened 2020-03-10
Issue 1656 header

alekpentchev:
# Description

Rewrite Header component to support React Hooks API.

## More Details

Solved common React on Rails issue with using multiple components using React Hooks on the same page. Used mount instead of shallow to wrap Header component for testing due to issues with updating state.

## Corresponding Issue

# Screenshots

image

---

### DIFF
--- a/client/app/components/Header/__tests__/Header.spec.jsx
+++ b/client/app/components/Header/__tests__/Header.spec.jsx
@@ -7,7 +7,7 @@ const component = (
   <Header
     home={{ name: 'Home', url: '/some-path' }}
     links={[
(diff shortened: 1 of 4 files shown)


alekpentchev:
@julianguyen It's relatively hard for me to reduce the cognitive complexity of props destructuring. Do you know how to tackle this issue?

julianguyen:
Hmm good question. We could move some functions outside of the Header function, but we would need to pass a lot of arguments. This would worsen readability. I think what you have good. We can just ignore this Code Climate warnings and mark it is "wontfix".

PR #1694, opened 2020-03-15
Rewrite Inputs Components To Use React Hooks API

Orelongz:
## Description

Rewrite Inputs Components To Use React Hooks API

## Corresponding Issue

### DIFF
--- a/client/app/components/Input/InputLocation.jsx
+++ b/client/app/components/Input/InputLocation.jsx
@@ -1,5 +1,5 @@
 // @flow
-import React from 'react';
+import React, { useState } from 'react';
 import LocationAutocomplete from 'location-autocomplete';
 import { I18n } from '../../libs/i18n';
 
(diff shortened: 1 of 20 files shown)


client/app/components/Input/InputLocation.jsx
```diff
@@ -10,44 +10,37 @@ export type Props = {
...
+  return (
+    <LocationAutocomplete
+      name="user[location]"
+      placeholder={placeholder}
+      googleAPIKey={apiKey}
+      onChange={handleChange}
+      onDropdownSelect={handleDropdownSelect}
+      className="smallerMarginBottom"
+      id={id}
+      aria-label={ariaLabel}
```
julianguyen: Looks like we don't need line 31 and can just use its value directly here!

client/app/components/Input/InputTag.jsx
```diff
@@ -15,185 +16,158 @@ export type Props = {
...
-    return checkbox.label === autocompleteLabel ? checkbox.label : '';
-  };
+  const getSuggestionValue = (checkbox: Checkbox) => (checkbox.label === autocompleteLabel ? checkbox.label : '');
 
-  onSuggestionsFetchRequested = ({ value }: { value: string }) => {
-    const newSuggestions = this.getSuggestions(value);
-    this.setState({
-      suggestions: newSuggestions,
-    });
+  // eslint-disable-next-line react/no-unused-prop-types
```
julianguyen: Can we remove this and fix the issue?

client/app/components/Input/InputTag.jsx
```diff
@@ -15,185 +16,158 @@ export type Props = {
...
-  };
+  const getSuggestionValue = (checkbox: Checkbox) => (checkbox.label === autocompleteLabel ? checkbox.label : '');
 
-  onSuggestionsFetchRequested = ({ value }: { value: string }) => {
-    const newSuggestions = this.getSuggestions(value);
-    this.setState({
-      suggestions: newSuggestions,
-    });
+  const onSuggestionsFetchRequested = (valueProp: { value: string }) => {
+    const { value } = valueProp;
```
Orelongz: This had to be done this way because of https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md#false-positives-sfc.

Orelongz: And @julianguyen, thanks for the pointer on the `react_component` error

julianguyen: Ah thanks for explaining! Let's leave it in there for now then!

Issue #1315, opened 2018-12-17
Long prose in yaml files is hard to work with


matcha440:
# Description 
The yaml file has in some places entire paragraphs of text, which in order to be able to read while making modifications you have to scroll horizontally.  As another consequence, it can be hard to measure progress on a translation since not all lines are equal.  Yaml files do support multi-line text, so I think it is possible to break these up so they are easier to read.

For example, if we look at [this line](https://github.com/ifmeorg/ifme/blob/v6.3.0/config/locales/en.yml#L391), it could be done like this and it would still work I think
```yaml
      data_access_answer: |
        Data is not given or made accessible to any third party.
        Deleting your account will remove all of your data.<br><br>
        User data is analyzed within the app to provide customized recommendations.
```

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

harshlele:
Hey, I'm new to open source, and this issue looks like it's pretty easy to solve. Can i work on this?

matcha440:
Yeah it seems pretty easy, but this has somewhat been a take it or leave it type thing for me.  Since word wrapping gets around this for the most part, maybe this isn't too important to have.  I'm still waiting for what others may have to say on the matter, though.

julianguyen:
@anushkajj This issue is still available! Feel free to assign yourself to it and start working on it!

anushkajj:
Hey! Will keeping the character limit to around 120 per line work?
I think this will prevent the problem of horizontal scrolling to read the files.

GayathriVenkatesh:
@julianguyen @anushkajj I would love to help out with this issue! Let me know if this is something I can still work on :)

julianguyen:
@anushkajj @GayathriVenkatesh Sorry for the late reply! I think 80 characters work! Feel free to work on this together

GayathriVenkatesh:
Hi @anushkajj, could you please fill me in on your progress? I would like to help out with whatever's left.

julianguyen:
@GayathriVenkatesh Great job doing this for en.yml! We'll want to do this for the other locales.

PR #1697, opened 2020-03-18
Remove long prose in en.yml

GayathriVenkatesh:
## Corresponding Issue
#1315 

---

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -127,8 +127,8 @@ en:
       invitation_expired: 'Invitation expired'
       invite: 'Invite your loved ones!'
       none: >-
-         'You currently have no allies! You can search for them or invite
-         loved ones to join!'
+         You currently have no allies! You can search for them or invite
+         loved ones to join!
       view_all_allies: 'View all allies'
       cancel: Cancel
     form:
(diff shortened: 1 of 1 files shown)


GayathriVenkatesh:
@julianguyen I have made all the lines in en.yml less than 80 characters long (approx). If this looks good to you, we shall proceed to do the same for the other yaml files :)

config/locales/en.yml
```diff
@@ -126,7 +126,8 @@ en:
       invited_on: 'Invited on: %{date}'
       invitation_expired: 'Invitation expired'
       invite: 'Invite your loved ones!'
-      none: 'You currently have no allies! You can search for them or invite loved ones to join!'
+      none: 'You currently have no allies! You can search for them or invite
```
julianguyen: So it looks like you're encountering failed tests because of how you created a new line.

Something like this:

```
Key: >
  This is a very long sentence
  that spans several lines in the YAML
  but which will be rendered as a string
  with only a single carriage return appended to the end.
```

Taken from this [StackOverflow post](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-over-multiple-lines).

GayathriVenkatesh: Fixed most of it, thanks :)
There seems to be an issue with line 436-437 when I split it into multiple lines. https://github.com/GayathriVenkatesh/ifme/blob/b2737c523d53a9fb14cca77c7b6bcd83009d1682/config/locales/en.yml#L436
Not sure why that's happening. Everything else works fine.

GayathriVenkatesh: Fixed most of it, thanks! 
There seems to be an issue with line 436 when I split it into multiple lines. Not sure why.
Everything else works fine.

PR #1705, opened 2020-03-24
Facebook Authentication

Manasa2850:
Issue #795 
I've just made a few changes. The issue is not fixed yet


Manasa2850:
@julianguyen in `app/models/user.rb` I just need to replicate the functions we have for Google authentication for Facebook also right?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -14,6 +14,17 @@ def google_oauth2
     end
   end
 
+  def facebook
```
julianguyen: We should test this new method!

app/controllers/registrations_controller.rb
```diff
@@ -14,4 +14,8 @@ def update_resource(resource, params)
       resource.update_with_password(params)
     end
   end
+
+  def oauth_provider?(provider)
+    provider == 'google_oauth2' || provider == 'facebook'
```
julianguyen: I think we can simplify this logic by doing the following:

```
 %w[google_oauth2 facebook].include? provider
```

Manasa2850:
@julianguyen I'm not getting the desired result. Could you please let me know if I need to change something in `user.rb`

config/initializers/devise.rb
```diff
@@ -317,4 +317,19 @@
...
+
+  require 'omniauth-facebook'
+  config.omniauth(
+    :facebook,
+    ENV['FACEBOOK_CLIENT_ID'],
+    ENV['FACEBOOK_CLIENT_SECRET'],
+    scope: 'public_profile,email',
+    info_fields: 'email,first_name,last_name',
+    client_options: {
+    site: 'https://graph.facebook.com/v2.11',
```
bmwachajr: If this site changes based on the Version number for graph, please make it an ENV value instead.

app/models/user.rb
```diff
@@ -55,7 +55,7 @@ class User < ApplicationRecord
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :invitable, :database_authenticatable, :registerable, :uid,
          :recoverable, :rememberable, :trackable, :validatable, :omniauthable,
-         omniauth_providers: [:google_oauth2]
+         omniauth_providers: %i[google_oauth2 facebook]
```
bmwachajr: A suggestion for increased maintainability, could be a good idea to rename the callback `facebook` to `facebook_oauth`.

app/models/user.rb
```diff
@@ -81,20 +81,27 @@ class User < ApplicationRecord
   validates :locale, inclusion: {
     in: Rails.application.config.i18n.available_locales.map(&:to_s).push(nil)
   }
-  validate :password_complexity
+  validate :password_complexity, unless: :oauth_provided?
```
bmwachajr: Love this

app/models/user.rb
```diff
@@ -81,20 +81,27 @@ class User < ApplicationRecord
...
 
   def active_for_authentication?
     super && !banned
   end
 
-  def self.find_for_google_oauth2(access_token)
-    user = find_or_initialize_by(email: access_token.info.email)
-    user.name ||= access_token.info.name
+  def self.find_for_google_oauth2(auth)
```
bmwachajr: Is `auth` an object or a value(token?)?

Manasa2850: It's a token

app/models/user.rb
Manasa2850: Sure I'll do this

config/initializers/devise.rb
Manasa2850: @bmwachajr sorry I didn't get you. What should I change to ENV value?
I currently have this:
`ENV['FACEBOOK_APP_ID'],
 ENV['FACEBOOK_APP_SECRET']
`

app/models/user.rb
julianguyen: This is a good question Ben asked!

It seems like it's an object, but line 94 treats these entire object as a token. That seems off to me. I would double-check what the value of auth actually is.

app/controllers/omniauth_callbacks_controller.rb
julianguyen: You'll want to add the test in `spec/controllers/omniauth_callbacks_controller_spec.rb`. The `GET #google_oauth2` describe block should give you hints on how to test the Facebook auth.

app/views/devise/shared/_links.erb
```diff
@@ -2,9 +2,19 @@
...
+          provider), method: :post, class: 'buttonGhostM' %>
       <% elsif provider.to_s == "google_oauth2" %>
-        <%= button_to t('devise.shared.sign_up_google'), omniauth_authorize_path(resource_name, provider), method: :post, class: 'buttonGhostM' %>
+        <%= button_to t('devise.shared.sign_up_google'), omniauth_authorize_path(resource_name,
+          provider), method: :post, class: 'buttonGhostM' %>
+      <% end -%>
+
+      <% if provider.to_s == "facebook" && (current_page?(new_user_session_path) || current_page?(new_user_password_path)) %>
+        <%= button_to t('devise.shared.sign_in_facebook'), omniauth_authorize_path(resource_name, provider),
+        method: :post, class: 'buttonGhostM' %>
```
julianguyen: We'll want to add some space margin between this button and the Sign Up with Google one above. Let's also add `smallMarginTop` to the class here.

app/views/devise/shared/_links.erb
```diff
@@ -2,9 +2,19 @@
...
+      <% elsif provider.to_s == "facebook" %>
+        <%= button_to t('devise.shared.sign_up_facebook'), omniauth_authorize_path(resource_name, provider),
+        method: :post, class: 'buttonGhostM' %>
```
julianguyen: Same comment as above.

app/controllers/omniauth_callbacks_controller.rb
julianguyen: This test will be helpful for uncovering potential bugs @bmwachajr has brought up below :)

bmwachajr:
Hey @Manasa2850 in case you need any help on this PR, let me know. Cheers.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -14,16 +14,31 @@ def google_oauth2
...
+    else
+      redirect_to new_user_session_path, notice: t('omniauth.access_denied')
+    end
+  end
+
   private
 
   def user
-    @user ||= User.find_for_google_oauth2(request.env['omniauth.auth'])
+    @user ||= User.find_for_facebook(request.env['omniauth.auth'])
```
bmwachajr, line 31:
```suggestion
    @user ||= User.find_for_oauth(request.env['omniauth.auth'])
```
I would consider refactring this to cater for both google and facebooth oauth in one fell swoop.

app/models/user.rb
```diff
@@ -81,28 +81,51 @@ class User < ApplicationRecord
...
+    user = find_or_initialize_by(email: auth.info.email)
+    user.name ||= auth.info.name
     user.password ||= Devise.friendly_token[0, 20]
-    update_access_token_fields(user: user, access_token: access_token)
+    update_access_token_fields(user: user, access_token: auth)
     user
   end
 
+  def self.find_for_facebook(auth)
```
bmwachajr, line 98: consider merging this method with the `def self.find_for_google_oauth2(auth)` method into
```suggestion
  def self.find_for_oauth(auth)
```
as hinted https://github.com/ifmeorg/ifme/pull/1705/files?file-filters%5B%5D=.rb#r411277154

app/models/user.rb
```diff
@@ -81,28 +81,43 @@ class User < ApplicationRecord
...
+          uid: auth.provider + auth.uid).first_or_create do |user|
+      UserBuilder::Builder.build(user: user, auth: auth)
+    end
+  end
+
   def google_access_token
     google_access_token_expired? ? update_access_token : token
   end
 
+  def facebook_access_token
```
bmwachajr, line 109: A suggestion to shrink `def google_access_token` and `def facebook_access_token` since it technically does the samething. Abstarct them into one method

app/models/user.rb
```diff
@@ -140,7 +155,15 @@ def update_access_token
...
 
+  def oauth_provided?
+    provider.present? || token.present?
+  end
+
   def google_access_token_expired?
     !access_expires_at || Time.zone.now > access_expires_at
   end
+
+  def facebook_access_token_expired?
```
bmwachajr, line 166: Same comment here abstract `google_access_token_expired?` and `facebook_access_token_expired?` into one method

Manasa2850:
@bmwachajr we've made the changes you suggested!

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -14,16 +14,31 @@ def google_oauth2
...
   def user
-    @user ||= User.find_for_google_oauth2(request.env['omniauth.auth'])
+    @user ||= User.find_for_oauth(request.env['omniauth.auth'])
   end
 
   def google_avatar
     request.env['omniauth.auth']&.[]('info')&.[]('image')
   end
 
+  def facebook_avatar
```
bmwachajr, line 38: It's looking good. a minor refactor here. Could we combine `google_avatar` and `facebook_avatar` into one `omniauth_avatar` method?

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -115,5 +119,89 @@
...
+    end
+
+    context 'user avatar image uploads' do
+      before { stub_env_for_omniauth_fb }
+      let(:oauth_email) { request.env['omniauth.auth']['info']['email'] }
+      let(:oauth_user) { User.find_by(email: oauth_email) }
+      let(:user) { oauth_user }
+      before { get :facebook }
+
+      context 'when third party avatar is nil' do
```
julianguyen, line 199: What happens when the third party avatar is not nil? I think we should test for that case as well :)

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -195,6 +195,26 @@
       let(:oauth_user) { User.find_by(email: oauth_email) }
       let(:user) { oauth_user }
       before { get :facebook }
+      
+      context 'when third party avatar is not nil' do
+        it 'uploads avatar when third_party_avatar has changed' do
+          new_avatar = 'http://example.com/images/different_profile.jpeg'
+          request.env['omniauth.auth']['info']['image'] = new_avatar
+          get :facebook
```
bmwachajr, line 203: For this failing test. you could debug using `byebug` in this action. Specifically, I would look to check that `request.env['omniauth.auth']['info']['image']` is really set to `new_avatar`.

Might need to stub `request.env['omniauth.auth']['info']['image'] = new_avatar` correctly

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -14,32 +14,43 @@ def google_oauth2
     end
   end
 
+  def facebook
+    if user.present?
+      user.accept_invitation! if invitation_token
```
bmwachajr, line 19: The `context 'when third party avatar is not nil'` is failing because we are not uploading the `third_party_avata` in the `facebook` OAuth action.

Right below this line, we shall have to add `upload_avatar(omniauth_avatar)`. Reference the `google_oauth2` action at [line 7](https://github.com/ifmeorg/ifme/pull/1705/files#diff-aed97266b0fe72994225a9cac3e4230bR7)

bmwachajr:
@julianguyen , just a little sidebar here.

It looks like both the `facebook` and `google_oauth2` actions are duplicates. Their specs are duplicates too. I suggest that once this PR solves the issue. we merge it as is and I will create a followup issue to refactor it.

It's a good idea to have short ship cycles for PRs. I could also use the refactor issue to guide @Manasa2850 and @GayathriVenkatesh on some really good practices.

What do you think?

Cc @Manasa2850 @GayathriVenkatesh would you be open to this?

julianguyen:
@bmwachajr This is really great feedback! Thanks for much for suggesting and offering to refactor. Support this 100%!

config/locales/de.yml
```diff
@@ -514,6 +514,7 @@ de:
   omniauth:
     google: Google
     access_denied: 'Zutritt verweigert'
+    facebook: Facebook
```
julianguyen, line 517: We don't need to add this since it already exists: https://github.com/ifmeorg/ifme/blob/master/config/locales/en.yml#L46

spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -115,5 +119,109 @@
...
+    end
+
+    context 'user avatar image uploads' do
+      before { stub_env_for_omniauth_fb }
+      let(:oauth_email) { request.env['omniauth.auth']['info']['email'] }
+      let(:oauth_user) { User.find_by(email: oauth_email) }
+      let(:user) { oauth_user }
+      before { get :facebook }
+      
+      context 'when third party avatar is not nil' do
```
julianguyen, line 199: I think we can rephrase this context as 'when third party avatar exists'.

PR #1717, opened 2020-04-09
Issue 1656 Notifications

Manasa2850:
Solves #1656

### DIFF
--- a/client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
+++ b/client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
@@ -2,7 +2,7 @@
 import React from 'react';
 import axios from 'axios';
 import { mount } from 'enzyme';
-import { Notifications } from '../index';
+import Notifications from '../index';
 
 let axiosGetSpy;
 let axiosDeleteSpy;
@@ -12,7 +12,7 @@ const data = {
   fetch_notifications: ['Notification 1'],
 };
 
-const component = (
+const getComponent = () => (
   <Notifications element={<button type="button">Notifications</button>} />
 );
 
--- a/client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
+++ b/client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
@@ -12,7 +12,7 @@ const data = {
   fetch_notifications: ['Notification 1'],
 };
 
-const getComponent = () => (
+const component = (
   <Notifications element={<button type="button">Notifications</button>} />
 );
 
--- a/client/app/widgets/Notifications/index.jsx
+++ b/client/app/widgets/Notifications/index.jsx
@@ -27,19 +27,12 @@ const Notifications = ({
   const [modalKey, setModalKey] = useState(undefined);
   const [signedInKey, setSignedInKey] = useState(undefined);
 
-  const setBody = (notifications) => {
+  const setBody = (notifications1: string[]) => { //I have changed notifications to notifications1 in order to avoid no-shadow error.
     let updatedNotifications = '';
-    notifications.forEach((item: string) => {
+    notifications1.forEach((item: string) => {
       updatedNotifications += `<div>${item}</div>`;
     });
     setNotifications(updatedNotifications);
-  };
-
-  const changeTitle = (count: number) => {
-    let { title } = window.document;
-    const eliminate = `${title.substr(0, title.indexOf(') '))})`;
-    title = title.replace(eliminate, '');
-    window.document.title = count === 0 ? title : `(${count}) ${title}`;
   };
 
   const fetchNotifications = () => {
@@ -71,14 +64,14 @@ const Notifications = ({
 
   useEffect(() => fetchNotifications(), []);
 
-  const getPusherKey = (signedInKey) => {
-    setSignedInKey(signedInKey);
+  const getPusherKey = (signedInKey1: number) => {
+    setSignedInKey(signedInKey1);
     const metaPusherKey = Array.from(
       window.document.getElementsByTagName('meta'),
     ).filter((item) => item.getAttribute('name') === 'pusher-key')[0];
     const pusherKey = metaPusherKey.getAttribute('content');
     const pusher = new window.Pusher(pusherKey, { encrypted: true });
-    const channel = pusher.subscribe(`private-${signedInKey}`);
+    const channel = pusher.subscribe(`private-${signedInKey1}`);
     channel.bind('new_notification', (response: any) => {
       if (response && response.data) {
         fetchNotifications();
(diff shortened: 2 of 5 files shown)


client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
```diff
@@ -12,7 +12,7 @@ const data = {
   fetch_notifications: ['Notification 1'],
 };
 
-const component = (
+const getComponent = () => (
```
julianguyen, line 15: I'm confused about why you made this into a function and renamed this. This new function isn't even being used in this file. I think we can keep what was originally here.

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,25 +18,26 @@ export type State = {
...
-    super(props);
-    this.state = { notifications: '', alreadyMounted: false, open: false };
-  }
+const Notifications = ({
+  element,
+}: Props) => {
+  const [notifications, setNotifications] = useState('');
+  const [alreadyMounted, setAlreadyMounted] = useState(false);
+  const [open, setOpen] = useState(false);
+  const [modalKey, setModalKey] = useState('');
```
julianguyen, line 27: Since this prop is not required, I don't think we need to set the initial state to a string. This should be `undefined` instead. I think this may be responsible for some issues you're having with linting and likely the tests.

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,25 +18,26 @@ export type State = {
...
+  const [signedInKey, setSignedInKey] = useState(0);
```
julianguyen, line 28: Same comment as above.

client/app/widgets/Notifications/index.jsx
```diff
@@ -45,90 +46,76 @@ export class Notifications extends React.Component<Props, State> {
...
-          onClick={this.clearNotifications}
+          onClick={clearNotifications}
         >
           {I18n.t('notifications.clear')}
         </button>
       </div>
     );
   };
-
-  render() {
```
julianguyen, line 115: We shouldn't be deleting the return value of this. Instead of using the render function, we should just return.

client/app/widgets/Notifications/__tests__/Notifications.spec.jsx
Manasa2850: Sure :+1:

client/app/widgets/Notifications/index.jsx
Manasa2850: Ok I'll make the required changes. Thank you!

Manasa2850:
@julianguyen could you please help me resolve the `no-use-before-define` error I'm getting in circleci.
Thanks!

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,117 +18,121 @@ export type State = {
...
+    ).filter((item) => item.getAttribute('name') === 'pusher-key')[0];
+    const pusherKey = metaPusherKey.getAttribute('content');
+    const pusher = new window.Pusher(pusherKey, { encrypted: true });
+    const channel = pusher.subscribe(`private-${signedInKey1}`);
+    channel.bind('new_notification', (response: any) => {
+      if (response && response.data) {
+        fetchNotifications();
+      }
+    });
+  };
```
whyderrick, line 88: Hey @Manasa2850, I'm brand new, but I think you can fix the `no-use-before-defined` warning you're seeing by moving this declaration (lines 75 - 88) above line 46. 

`no-use-before-defined` is complaining because you're referring to `getPusherKey` on line 46, but defining `getPusherKey` on 75. Does that make sense?

Manasa2850: That won't help because `getPusherKey()` and `fetchNotifications()` use each other

whyderrick: Oh. Thanks. Let me look a little more closely then. I'll post a new comment when I have a slightly better understanding of what's happending

whyderrick: Could we have `getPusherKey` return the key from line 80 or the pusher instance on line 81 and pull out a separate function `subscribeToPusher`  with the code on line 82 - 87? Right now it's doing 2 things: getting the key *and* subscribing to the channel with that key. Splitting that into two functions might let us define `getPusherKey` above `fetchNotifications` and add some clarity about the responsibility of each function.

(I don't have the project running locally, or much context for the project, so this might be out of step with  practices in the repo, misinformed, or just bad. I'm sorry if any of those are true).

Manasa2850: That didn't work too :(
You can check my latest commits. I've tried what you suggested

Jacek-202: Hi @Manasa2850 
Check this issue https://github.com/eslint/eslint/issues/12473
It should be ok to disable checking function declaring for this particular file.
/*eslint no-use-before-define: ["error", { "functions": false }]*/

whyderrick: Thanks for finding the eslint issue, Jacek!

I was planning to ask in the channel for help refactoring these fix this error, because I don't have the JS chops to do it myself. But if there aren't good suggestions on that today, I'd support skipping the line that raises on it.

Manasa2850: Thanks @Jacek-202 !!

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,117 +18,122 @@ export type State = {
...
+  const fetchNotifications = () => {
+    setAlreadyMounted(alreadyMounted);
+    setSignedInKey(signedInKey);
     return axios
       .get('/notifications/signed_in')
       .then((response: any) => {
         if (response && response.data && response.data.signed_in !== -1) {
           if (response.data.signed_in !== signedInKey) {
-            this.getPusherKey(response.data.signed_in);
+            getPusherKey(response.data.signed_in);
```
whyderrick, line 54: If you place `// eslint-disable-next-line no-use-before-define` above this line, you'll silence the linter error that's breaking on CircleCI. Then you shouldn't need the comment on line 75.

whyderrick: @Manasa2850 I asked in [Slack](https://ifme.slack.com/archives/C725HPXFF/p1586841508019100) and Nishiki had a better solution. We can define change our function definitions to use the function keyword instead of the assigning them to variables. 

`const fetchNotifications = () => {` becomes ``

and `const getPusherKey = (signedInKey1: number) => {` becomes `function getPusherKey(signedInKey1: number)`

After I gave you non-working suggestions, I set up the repo so I could try things out. I ran the linter locally by running `eslint 'app/**/*.js' 'app/**/*.jsx' app/widgets/Notifications/index.jsx` so I'm pretty sure it'll work. (You can use that command to see what lint errors you have without actually pushing up a new commit).

Issue #1687, opened 2020-03-07
Track progress on strategies


GayathriVenkatesh:
# Description

For every strategy created, it would be nice to have a feature that allows the user to enter whether he has followed it that day.

For example, in the screenshot below, I have created 3 strategies. I feel it would be convenient to have an option to keep track of how many days I have followed them. 

![Screenshot from 2020-03-08 00-23-48](https://user-images.githubusercontent.com/32172764/76151184-23ea9500-60d8-11ea-89ec-1c5aed3b42f5.png)

When the user clicks on a date, it can show the list of all strategies with checkboxes. The user can then check or uncheck each of them.
There could also be a weekly report on how many days the user has achieved the targets.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

julianguyen:
This is an interesting idea - integrating these Strategies as part of a TODO list. We could generate this TODO list from Strategies that are tagged in Moments (as these types of posts are supposed to reflect a user's current state). 

We would have to think about what kind of statuses we want on a TODO item, e.g.
* In-progress
* Completed
* Cancelled

alekpentchev:
I agree, it's an interesting idea. Maybe we could introduce another tab to the sidemenu would be a nice place for that?
And each strategy could be automatically translated to a TODO where, as stated by you, could check / uncheck them as done.
The strategy list could show on each tab how many days the user has followed it.

julianguyen:
I like that idea @alekpentchev! That way this TODO feature can be easily accessed.

julianguyen:
So something we need to think about is how this progress tracking feature could potentially leverage the "daily reminder email" checkbox we have set up:

![image](https://user-images.githubusercontent.com/3010728/77861277-c7efc980-71c8-11ea-8195-dc6467df40b8.png)

If someone is "finished" a Strategy, do we disable this checkbox?

In order to work on this, @GayathriVenkatesh or @alekpentchev, we should write out a plan for what we're doing with the data.

GayathriVenkatesh:
[...]

PR #1719, opened 2020-04-14
Track Progress on Strategies

GayathriVenkatesh:
## Description 

This PR creates a Task page that renders all your strategies and displays your progress in each one of them. I have added a column 'Mark as Finished' to the strategies table. I have also created a task _strategies:update_ which automatically runs every 24 hours and updates <b>no_of_days_followed</b> for each strategy (based on its 'finished' value) and the **total_no_of_days** (date.today - date of creation).

The page looks like this so far :

aa

Screenshot from 2020-04-19 09-42-10

My initial thought was to also leverage this data to render progress charts and help in easy visualisation of data. Hence, this PR is not complete. If you have any suggestions on what else the page could render and make it more usable, please leave your thoughts in the comments below!

## Corresponding Issue
#1687 

---


nshki:
@GayathriVenkatesh thanks for the PR! Would you mind providing a PR description of the general strategies you used? This is a pretty large PR, so contextual help from you as the author would be very appreciated. :bow:

.rake_tasks~
```diff
@@ -0,0 +1,114 @@
+about
```
julianguyen, line 1: This file should not be checked in 👀

Gemfile
```diff
@@ -51,6 +51,9 @@ gem 'selenium-webdriver', '~> 3.142.3'
 
 gem 'rubyzip', '~> 1.3.0'
 
+gem 'rake', '12.3.3'
```
julianguyen, line 54: We shouldn't need to put `rake` our Gemfile.

app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,84 @@
+body {
```
julianguyen, line 1: We shouldn't need to introduce new CSS for this feature. We should be able to leverage existing CSS and components.

app/controllers/strategies_controller.rb
```diff
@@ -82,6 +82,13 @@ def premade
     end
   end
 
+  def create_task(strategy)
+    task = Task.find(strategy.id)
+    task.finished = strategy.finished
```
julianguyen, line 87: We should wrap lines 87-89 in a condition to make sure task exists.

app/controllers/strategies_controller.rb
```diff
@@ -94,12 +101,15 @@ def update
...
     shared_update(@strategy, strategy_params)
+    create_task(@strategy)
   end
 
   # DELETE /strategies/1
   # DELETE /strategies/1.json
   def destroy
     shared_destroy(@strategy)
+    task = Task.find(@strategy.id)
+    task.delete
```
julianguyen, line 112: Will this delete all entries? Would `destroy_all` work better?

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class TasksController < ApplicationController
+  before_action :set_task, only: %i[show]
+
+  def index
+    @tasks = Task.all
+  end
+
+  def show; end
```
julianguyen, line 10: We can get rid of this and exclude this route in `routes.rb`.

app/helpers/todo_items_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module TodoItemsHelper
```
julianguyen, line 3: This file can be deleted since it's empty.

app/models/strategy.rb
```diff
@@ -69,4 +68,19 @@ def published?
...
+
+  def update_tasks_table
+    @task = Task.exists?(id) ? Task.find(id) : Task.new(id: id)
+    @task.title = name
+    @task.description = description
+    @task.finished = finished
+    @task.date = updated_at
+    @task.no_of_days_followed = finished ? 1 : 0
+    @task.total_no_of_days = (created_at.to_date - Time.zone.today).to_i + 1
+    @task.save
```
julianguyen, line 84: We should handle errors in the event that save does not work.

app/views/pages/home.html.erb
```diff
@@ -1,4 +1,6 @@
 <% if user_signed_in? && @stories.any? %>
+  <% @page_new = "New Moment" %>
```
julianguyen, line 2: We should be using a translation key here.

app/views/todo_items/_todo_item.html.erb
```diff
@@ -0,0 +1,2 @@
+<p><%= todo_item.content %></p>
+<%= link_to task_todo_item_path(@task, todo_item.id), method: :delete, data: { confirm: "Are you sure?" } %>
```
julianguyen, line 2: We should be using a translation key here.

config/env/development.example.env
```diff
@@ -31,5 +31,5 @@ RAISE_DELIVERY_ERRORS="false"
 
 # Database
 PSQL_HOST=""
-PSQL_USERNAME=""
-PSQL_PASSWORD=""
\ No newline at end of file
+PSQL_USERNAME="pguser"
+PSQL_PASSWORD="1234"
```
julianguyen, line 35: We should not be committing this change.

config/env/test.example.env
```diff
@@ -31,5 +31,5 @@ RAISE_DELIVERY_ERRORS="false"
 
 # Database
 PSQL_HOST=""
-PSQL_USERNAME=""
-PSQL_PASSWORD=""
+PSQL_USERNAME="pguser"
```
julianguyen, line 34: We should not be committing this change.

db/migrate/20200201005117_add_visible_to_strategies.rb
```diff
@@ -1,5 +0,0 @@
-class AddVisibleToStrategies < ActiveRecord::Migration[5.2]
-  def change
-    add_column :strategies, :visible, :boolean, default: true
```
julianguyen, line 3: Why is this being deleted?

db/migrate/20200224201131_create_moment_categories_join_table.rb
```diff
@@ -1,12 +1,12 @@
...
-      t.integer :category_id
+      t.integer :strategy_id
     end
 
-    add_index :moments_categories, [:moment_id, :category_id], unique: true
-    add_foreign_key :moments_categories, :moments
-    add_foreign_key :moments_categories, :categories
+    add_index :moments_strategies, [:moment_id, :strategy_id], unique: true
+    add_foreign_key :moments_strategies, :moments
+    add_foreign_key :moments_strategies, :strategies
```
julianguyen, line 10: You will need to create an entirely new migration to make these changes since we do not rollback migrations in production.

julianguyen: You're getting migration errors in the CircleCI build because of what I pointed out. I recommend reading this [article](https://guides.rubyonrails.org/active_record_migrations.html) about migrations. You'll want to create a new migration for your changes using `rails g migration [etc..]`.

julianguyen: So it looks like actually already have a `moments_strategies` table. I'm not sure why we need to override the migration for `moments_categories` with a table that already exists. Looks like we just want to unstaged and remove this change entirely.

GayathriVenkatesh: That was a totally unintentional commit, sorry about that!

db/migrate/20200201005117_add_visible_to_strategies.rb
julianguyen: Let's not modify this file. I would remove changes to this file from your PR. We shouldn't be modifying previous migration files in any way. This may be responsible for the test migration error in CircleCI.

julianguyen: So you have to remove this change entirely from this pull request. This is still causing the test migration error in CircleCI. There are two ways you can do this: 

1) You should do this by squashing all your commits into one. Doing a soft reset and remove this file and re-commit everything else.

2) Revert the specific commit that made this change

julianguyen:
Hmm try running `bin/rails db:migrate RAILS_ENV=test` in your dev environoment and committing the changes?

julianguyen:
Like a previous comment, I think you should try squashing all these commits into one and deleting the unnecessary file changes and running migrations again.

app/controllers/strategies_controller.rb
```diff
@@ -94,12 +103,14 @@ def update
     end
     empty_array_for :viewers, :category
     shared_update(@strategy, strategy_params)
+    create_task(@strategy)
```
julianguyen, line 106: We should be testing for this new task functionality in the test for update.

app/controllers/strategies_controller.rb
```diff
@@ -94,12 +103,14 @@ def update
...
   end
 
   # DELETE /strategies/1
   # DELETE /strategies/1.json
   def destroy
     shared_destroy(@strategy)
+    Task.where(id: @strategy.id).destroy_all
```
julianguyen, line 113: We should be testing for this new task functionality in the test for destroy.

app/controllers/strategies_controller.rb
```diff
@@ -82,6 +82,15 @@ def premade
     end
   end
 
+  def create_task(strategy)
```
julianguyen, line 85: Let's move this to a helper.

app/controllers/tasks_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class TasksController < ApplicationController
```
julianguyen, line 3: We should be testing this controller.

app/controllers/todo_items_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+#
+class TodoItemsController < ApplicationController
```
julianguyen, line 3: Same as above.

app/helpers/tasks_helper.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module TasksHelper
```
julianguyen, line 3: We should be unit testing this helper.

app/helpers/strategies_helper.rb
```diff
@@ -34,6 +34,7 @@ def strategy_form_inputs(strategy, viewers)
      get_viewers_input(viewers, 'strategy', 'strategies', strategy),
      build_strategy_comment(strategy), build_strategy_publishing(strategy),
      build_strategy_visible(strategy),
+     build_strategy_finished(strategy),
      Rails.env.development? ? build_strategy_bookmarked(strategy) : {},
```
julianguyen, line 38: Looks like we can remove this dev flag from here in this PR.

Gemfile
```diff
@@ -52,6 +52,8 @@ gem 'selenium-webdriver', '~> 3.142.3'
 
 gem 'rubyzip', '~> 1.3.0'
 
+gem 'whenever'
```
julianguyen, line 55: Let's actually move the use of this gem. In production (on Heroku), we have a task scheduler.

config/schedule.rb
```diff
@@ -0,0 +1,26 @@
+# Use this file to easily define all of your cron jobs.
```
julianguyen, line 1: Let's actually move the use of this gem. In production (on Heroku), we have a task scheduler.

db/schema.rb
```diff
@@ -252,6 +253,39 @@
...
+  end
+
+  create_table "todo_lists", force: :cascade do |t|
+    t.date "date"
+    t.boolean "finished"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "todolists", force: :cascade do |t|
```
julianguyen, line 282: I'm noticing there is both `todo_lists` and `todolists`. You might have to rollback your migrations to remove the unwanted `todolists`.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:user) { create(:user) }
+  let(:task) { create(:task) }
+
+  describe '#index' do
+    let(:task) { create(:task, title: 'test') }
+
+    context 'when the user is logged in' do
+      include_context :logged_in_user
+
+      context 'when search params are provided' do
```
julianguyen, line 13: I'm confused why you're testing for search params when there's no search functionality on this page so far.

Manasa2850:
@julianguyen could you help with the circleci build error. It says lines 155 and 171 of `strategies_helper` are causing it. But we haven't changed those lines at all.
Thanks

julianguyen:
@Manasa2850 It looks like you have two tests in the StrategiesHelper spec failing. You'll want to compare the the expected output and actual output to determine what went wrong. Your changes do affect these tests because you are adding `strategy_finished`.

Issue #1713, opened 2020-04-01
Recommend resources for users to use in Moments posts


julianguyen:
# Description

Our [Resources](https://www.if-me.org/resources) has a wealth of tagged resources that we can leverage for basic recommendations in the app.

Based on what a user has inputted in their Moments post (including tagged Categories, Moods, and Strategies), we can recommend some resources.

We can display them at the bottom of a post, like how we do for Strategies. 

![image](https://user-images.githubusercontent.com/3010728/78108869-09d96500-73ad-11ea-8d4f-90f8fc42e78c.png)

Let's chat more in this issue about how we want to build this recommendation engine without being spammy or not useful or potentially triggering.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

S-Warmenhoven:
Aline and I would like to take on this issue to contribute to the project.  We realize that it's not tagged as "beginner-friendly", but would like to take on the challenge together.  Thank you.

julianguyen:
@S-Warmenhoven @AlineRibeiro I think it would be great for you two to pair on this! Let's definitely sync on this feature then :) Thanks for taking it on!

PR #1720, opened 2020-04-16
Recommended resources issue 1713

S-Warmenhoven:
# Description

Based on what a user has inputted in their Moment post (including tagged Categories, Moods, and Strategies), keywords are extracted and matched with Resource tags to display up to 3 recommended resources on the Moment Show Page.  Additionally, there is a "More..." link to redirect the user to the Resources Index Page.
We created two classes called MomentKeywords and ResourceRecommendation that are called in the moments_controller for each Moment.

## More Details

This is Phase #1 of this feature.  There are some other ways to make the recommendations more optimal if necessary for further phases, such as using Text processing tools.

Further suggestions for the Recommendation engine:
- Handle singular/plural
- Handle related words (anxious/anxiety)
- Resources need more tags or another type of tag
- Filter by user location (country-specific resources)
- Searching resources using the database can be more suitable for this feature than parsing and searching though the JSON file

Also, with the "More..." link, certain tags could be pre-selected too when directed to the index page.

## Corresponding Issue

Issue [#1713]

---


app/views/moments/show.html.erb
```diff
@@ -43,6 +43,16 @@
   </div>
 <% end %>
 
+<div class="smallMarginTop">
```
julianguyen, line 46: I think we should only display this new block if `@resources` exists.

app/views/moments/show.html.erb
```diff
@@ -43,6 +43,16 @@
...
+  <div class="label"><%= label_tag "What resources could help?" %> </div>
+  <ul>
+    <% @resources.take(3).each do |item|  %>
+      <li><%= link_to item['name'], item['link'] %></li>
+    <%end %>
+    <li><%= link_to  'More...', resources_path %></li>
```
julianguyen, line 52: For the more link, I think we should also have some tag(s) pre-selected e.g. `https://www.if-me.org/resources?filter[]=hotlines` based on the preprocessing you've done.

app/services/moment_keywords.rb
```diff
@@ -0,0 +1,53 @@
...
+    @moment.name.split
+  end
+
+  def extract_why
+    strip_html(@moment.why).split
+  end
+
+  def extract_fix
+    strip_html(@moment.fix).split
+  end
```
nshki, line 41: The shape of these methods are identical. Could be a good opportunity to DRY this up a bit! Perhaps a general `extract` method that takes in an argument? This new method [could also be used above for this line](https://github.com/ifmeorg/ifme/blob/78adeaa44bc1dd06bd969b0fa7561df3da67b25f/app/services/moment_keywords.rb#L27).

app/services/moment_keywords.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class MomentKeywords
+  def initialize(moment)
+    @moment = moment
+    @moment_keywords = []
+  end
+
+  def extract_moment_keywords
```
nshki, line 9: Not a blocker by any means, but the class name and method name here seem a little redundant. Could be worth exploring some other ways to use the class, such as:

```ruby
MomentKeywords.for(moment)

# Or perhaps something like...

MomentKeywords.new(moment).call
```

app/services/resource_recommendation.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class ResourceRecommendation
+  def initialize(moment)
+    @moment = moment
+    @moment_keywords = []
+  end
+
+  def resources
```
nshki, line 9: Similar consideration for this class—I wonder if it could be worth considering something like:

```ruby
ResourceRecommendations.for(moment)

# or

ResourceRecommendations.new(moment).call
```

app/services/resource_recommendation.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class ResourceRecommendation
```
nshki, line 3: Not a blocker by any means, but what if we pluralized this class name to `ResourceRecommendations` to make it clear that we'll be getting multiple recommendations?

spec/services/moment_keywords_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  it 'removes special characters' do
+    expect(keywords).to include('teachers')
+  end
+
+  it 'returns an array of all the key words' do
+    expect(keywords).to eq(
+                            ['free', 'description','name', 'blog', 'name', 'books',
+                             'addiction', 'self care', 'teachers']
+                        )
+  end
```
nshki, line 33: Great set of tests! Let's add a case for making sure that HTML is properly sanitized.

S-Warmenhoven:
Thank you for the reviews @julianguyen and @nshki.  @AlineRibeiro and I will start working on the modifications tomorrow. Have a great evening!

app/controllers/moments_controller.rb
```diff
@@ -25,6 +25,13 @@ def index
   # GET /moments/1.json
   def show
     show_with_comments(@moment)
+    @resources = ResourceRecommendations.new(@moment).call
+    @matched_tags = ResourceRecommendations.new(@moment).matched_tags
+    @moment_keywords = MomentKeywords.new(@moment).call
+    @filter_tags = @matched_tags.map do |t|
+      "filter[]=#{t}&"
+    end
+    @filter_tags = @filter_tags.join()
```
julianguyen, line 34: Is this line necessary? Nothing is being joined to @filter_tags.

app/services/resource_recommendations.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def matched_tags
+    @moment_keywords = MomentKeywords.new(@moment).call
+     resource_tags = all_resources.flat_map do |resource|
+       resource['tags'].map do |tag|
+         tag.tr('_', '-')
+       end
+     end
+     matched_tags = (@moment_keywords & resource_tags)
```
julianguyen, line 26: Should we be returning `@moment_keywords & resource_tags` instead of setting it to the `matched_tags` variable?

S-Warmenhoven:
Hello @nshki, @julianguyen,  and other reviewers.  We have updated our code according to the initial code reviews, and all local and external tests currently pass.  Please let us know if you have any questions or concerns.  Kind regards, TeamBits :)

app/views/moments/show.html.erb
```diff
@@ -43,6 +43,18 @@
   </div>
 <% end %>
 
+<% if @resources.any? %>
+<div class="smallMarginTop">
+  <div class="label"><%= label_tag "What resources could help?" %> </div>
```
julianguyen, line 48: There's unnecessary whitespace after the `label_tag` we should remove so that it looks like `<div class="label"><%= label_tag "What resources could help?" %></div>`.

app/views/moments/show.html.erb
```diff
@@ -43,6 +43,18 @@
...
+  <ul>
+    <% @resources.take(3).each do |item|  %>
+      <li><%= link_to item['name'], item['link'] %></li>
+    <%end %>
+    <li><%= link_to  'More...', "/resources?#{@resources_tags}" %></li>
```
julianguyen, line 53: "More..." should be in our translation files. Alternatively, you could use the existing `t('load_more')` key.

app/views/moments/show.html.erb
julianguyen: I also realized that this text should be in our translation files. Feel free to use Google Translate for now.

S-Warmenhoven: Hi Julia!  Thank you for your comments.  I do have 2 questions, please:  
1. With regards to the white space, I'm not sure which space you are referring to? To me the code and your example look exactly the same.  Are you referring to the indent space?
<div class="label"><%= label_tag "What resources could help?" %> </div>
<div class="label"><%= label_tag "What resources could help?" %></div>
2. Do we use the Chinese(Simplified) version for Google Translate?

AlineRibeiro:
Thank you @julianguyen, @nshki and @S-Warmenhoven !

Issue #1703, opened 2020-03-21
Skip to main content link 


julianguyen:
# Description

Let's add a "Skip to main content" link so that screen-reader or keyboard navigation users can skip the navigation menus.

We'll want to create a `SkipToContent` React component. We'll want to make sure it skips over to the correct content container for different user states.

We'll want to apply IDs to the elements that use the following classes. For example, we can use `#mainContent`.

* User signed in: `.dashboardSection`
* User not signed in: `.staticContent`

The component should look something like this:

```
<SKipToContent id="mainContent" />
```

The component itself should be an anchor link to `#mainContent`:

```
<a href="#maincontent">Skip to main content</a>
```

The component should only appear when a user tabs into a page. [Deque](https://www.deque.com/) has a good example of what we want to create.

# Screenshots

This is just a rough mockup of what we want this component to look like! Feel free to change the design as you see fit.

![image](https://user-images.githubusercontent.com/3010728/77232890-e7e81300-6b60-11ea-94d6-37247f071b4f.png)

![image](https://user-images.githubusercontent.com/3010728/77232900-f46c6b80-6b60-11ea-9e72-d8e1ef12545c.png)

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

discombobulateme:
I would like very much to work on it :)

julianguyen:
Assigned you! For future reference, you can assign yourself to issues :)

Feel free to DM me on Slack if you have further questions about this issue!

Victoria-Hodder:
As we discussed on Slack, @discombobulateme  and I will pair programme on this :) Looking forward to it!!

discombobulateme:
Dear community, just a quick note to use here as a progression tracker. This is an image of the pre-approved design (at #design slack channel). We are still working on the component... studying a bit about react \o/ 

![bueLinkTop](https://user-images.githubusercontent.com/7980624/78684105-60d0c400-78f0-11ea-80b1-0f2c0192969a.png)
![blueSquare2](https://user-images.githubusercontent.com/7980624/78684139-68906880-78f0-11ea-839b-442cff9ddf4f.png)
[...]

PR #1721, opened 2020-04-17
Skip to content

discombobulateme:
# Description

We added a Skip to Main Content component.

## More Details

This is to ensure readability for screen readers and keyboard users.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1703

# Screenshots

Screen Shot 2020-04-17 at 20 00 02
Screen Shot 2020-04-17 at 20 00 13

---


client/app/components/SkipToContent/SkipToContent.scss
```diff
@@ -0,0 +1,26 @@
...
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+}
+ 
+.skipToContent:focus {
+  position: static;
+  width: auto;
+  height: auto;
+  color: #990019;
```
julianguyen, line 16: Let's use the global variable for this colour i.e. `$carmine` instead.

client/app/components/SkipToContent/SkipToContent.scss
```diff
@@ -0,0 +1,26 @@
...
+  background: #91D7E8;
```
julianguyen, line 17: Let's use the global variable for this colour i.e. `$cornflower` instead.

client/app/components/SkipToContent/index.jsx
```diff
@@ -0,0 +1,15 @@
...
+import React from 'react';
+import css from './SkipToContent.scss';
+
+export type Props = {
+  id?: string,
+};
+
+function SkipToContent(props: Props) {
+
+  return <a className={css.skipToContent} href={`#${props.id}`}>Skip to main content</a>;
```
julianguyen, line 11: Since our app supports internationalization (i18n) we'll want to add "Skip to main content" as a key in `config/locales/en.yml` and translate it in other languages too. You can use Google translate here for now.

Then you can use 

```
import { I18n } from '../../libs/i18n';

// etc.

I18n.t('navigation.skip_to_main_content'); 

```

client/app/components/SkipToContent/index.jsx
```diff
@@ -0,0 +1,15 @@
+// @flow
+import React from 'react';
+import css from './SkipToContent.scss';
+
+export type Props = {
+  id?: string,
```
julianguyen, line 6: I would make this prop required.

julianguyen: So this issue is responsible for the following Flow error that's showing up in the CircleCI build:

```
Cannot coerce `props.id` to string because undefined [1] should not be coerced.
 

 
   app/components/SkipToContent/index.jsx:13:53
 
   13|   return <a className={css.skipToContent} href={`#${props.id}`}>Skip to main content</a>;
 
                                                           ^^^^^^^^
 

 
References:
 
   app/components/SkipToContent/index.jsx:7:8
 
    7|   id?: string,
 
              ^^^^^^ [1]
 
```

discombobulateme:
Dear @julianguyen and @nshki, I think @Victoria-Hodder and I finally got this!!!! 👊🏼 how do we PR this PR? 🧐 to code to code review again?

client/package.json
```diff
@@ -49,7 +49,8 @@
     "react-html-parser": "^2.0.2",
     "react-lazyload": "^2.6.2",
     "react-on-rails": "11.3.0",
-    "react-render-html": "^0.6.0"
```
julianguyen, line 52: We can undo these changes since we're not using these dependencies. Same goes for the `yarn.lock` changes.

yarn.lock
```diff
@@ -2,3 +2,179 @@
 # yarn lockfile v1
 
 
+"@babel/runtime@^7.1.2", "@babel/runtime@^7.4.0":
```
julianguyen, line 5: Let's undo these changes!

client/app/components/SkipToContent/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+export type Props = {
+  id: string,
+};
+
+function SkipToContent(props: Props) {
+  const { id } = props;
+
+  window.addEventListener('Props', () => {
+    window.scrollTo(0, 0)});
```
julianguyen, line 15: I think we're missing a `}` here.

I'm not sure what listening to "Props" actually does. I would assume it's listening to all prop changes? I don't think we want to do that.

Instead, I think we want to put an `onClick` listener on the link element below and put the scroll logic there.

client/package.json
discombobulateme: Dear Julia, I will be very glad to undo it but I guess they came automatically from when I pulled updates from Master 😅. I wouldn't know how to und them 🤨. Any help?

julianguyen: You'll want to do a soft reset on the commit that has those changes and remove them from it. If that commit isn't the latest one. You can just cherry-pick the commits that are on top of that one.

Alternatively, you could just squash everything in one commit and do the soft reset on that.

client/app/components/SkipToContent/index.jsx
Victoria-Hodder: Hey @julianguyen ! Thanks for the feedback. We started working on that this week.

Would you like us to refactor our current SkipToContent function into a class component (in order to add the scroll logic)? Or could we also use react hooks? 

I'm asking because we're not sure if react hooks can be used in a react on rails context. 

Thank you!

julianguyen: You can use React Hooks here! But if you feel more comfortable using React classes, that's fine too.

client/app/startup/registration.js
```diff
@@ -18,7 +18,12 @@ import { Logo } from '../components/Logo';
 import { Modal } from '../components/Modal';
 import { Notifications } from '../widgets/Notifications';
 import { Resource } from '../components/Resource';
+<<<<<<< HEAD
```
julianguyen, line 21: Looks like this merge conflict was not resolved.

client/package.json
Victoria-Hodder: Could we also use `git revert`?
From what I understand `reset` would lose the later commits and could lead to some conflicts with the memory/potential loss of code.

Issue #1724, opened 2020-04-18
Fix warning in DynamicForm.spec.jsx


julianguyen:
# Description

When running our automated React tests we get a warning from `DynamicForm.spec.jsx` related to unique keys. 

# Reproduction Steps

Run `yarn test -- client/app/components/Form/__tests__/DynamicForm.spec.jsx`. The tests pass but we get the following warnings:

```    console.error node_modules/react-dom/cjs/react-dom.development.js:88
      Warning: Encountered two children with the same key, `name-id`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
          in div (created by DynamicForm)
          in DynamicForm (created by _default)
          in _default (created by WrapperComponent)
          in WrapperComponent
```

We're not getting this issue in the implementation so there must a bug with the mocks.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

renjithabby:
Hello there, I'd like to help with this. i have also sent an email to join.ifme@gmail.com. thanks.

julianguyen:
@renjithabby Thanks! It would be great to get this fixed :) Once you accept the team invite, you can assign yourself to this issue!

PR #1726, opened 2020-04-20
 Fix test warning in DynamicForm.spec.jsx

renjithabby:
# Description
The test was giving a warning Warning: Encountered two children with the same key, `name-id`. This was caused by the mutation of mockInputs by unshift operation, if the nameValue property is available. Fixed the issue by adding the new input data to the defaultMockInputs without mutating using spread.
## More Details
Minor refactor , created a getMockInputs() to separate out the logic of retrieving the correct mockInput( based on the nameValue property) and creating the component.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/1724

# Screenshots

Screen Shot 2020-04-21 at 00 25 32

---

### DIFF
--- a/client/app/components/Form/__tests__/DynamicForm.spec.jsx
+++ b/client/app/components/Form/__tests__/DynamicForm.spec.jsx
@@ -7,7 +7,7 @@ import { InputMocks } from '../../../mocks/InputMocks';
 
 // TODO (julianguyen): Include InputTextarea after writing stubs for pell editor
 
-const mockInputs = [
+const defualtMockInputs = [
   { ...InputMocks.inputTextProps, required: true },
   InputMocks.inputSelectProps,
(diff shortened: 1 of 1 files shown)


client/app/components/Form/__tests__/DynamicForm.spec.jsx
```diff
@@ -7,7 +7,7 @@ import { InputMocks } from '../../../mocks/InputMocks';
 
 // TODO (julianguyen): Include InputTextarea after writing stubs for pell editor
 
-const mockInputs = [
+const defualtMockInputs = [
```
julianguyen, line 10: Looks like there's a typo, it should be `defaultMockInputs` instead.

renjithabby: Thanks for the review. fixed in the latest commit.

Issue #1580, opened 2019-10-07
Allow Moments/Strategies to be bookmarked


LMulvey:
**Bookmarks 🔖**

Users should have the ability to bookmark Moments and Strategies. This is identified by a column in the database for these items labeled `bookmark: boolean`

Users can click a bookmark toggle button on the Moment or Strategy they want to bookmark and it will add it to the bookmark list. Clicking the button again will remove the Moment or Strategy from the bookmark list.

[Sub-task for issue #187]

dhuang612:
Hi,

Is this available to work on?

GayathriVenkatesh:
Hi @LMulvey, is anyone presently working on this issue? If not, I'd love to take it up!

LMulvey:
All yours, @GayathriVenkatesh :)

Issue #187, opened 2016-05-30
Care Plan Feature Foundations


julianguyen:
# Description

Recently thought about having a panic button feature, where you can add Strategies and certain words of encouragement (user and their allies). When a user is going through a rough time, they can press this button and view all these nice things.

Also, include links and numbers to hotlines and crisis services.

EDIT (10/2018): Recently gave this feature more thought I think the panic button should be a collection of bookmarked comments, Moments, and Strategies. From a DB standpoint, we would add a column `bookmark: boolean`.

We could have special user input fields for contact information (name/email/phone number) and safety plan (a user can write out their plan of reaching out for direct support, could just be a regular textarea field).

+1 to the idea folks have shared below about determining the correct crisis hotline number based on location.

Maybe calling it the panic button might be triggering or insensitive. So it's also worth exploring a better name e.g. rescue button, help me button, etc.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

starbelly:
@julianguyen Where would the panic button be located in the UI? I assume it would be most beneficial if it was always within reach. That aside, let me make sure I understand the specification for this feature:

* User hits panic button
* Panic button sends request to sever
* Server looks up friends and allies, 
* Server fires off emails to friends and allies with link to the /strategies 

 Is this correct?

julianguyen:
Good question @starbelly! I agree, it should be easy to use - so I'm thinking in the main navigation.

I was imagining it as, user hits the panic button and sees:
* A list of bookmarked Strategies is displayed e.g. user bookmarks a breathing exercise strategy
* A list of bookmarked Allies they can reach out to (I like the idea of that firing off emails to those allies, but what would be in that email? Would it be something like, "You should check up on [name here]?"
* A list of links to numbers to hotlines and crisis services

starbelly:
@julianguyen Yeah, I'll take this on.

julianguyen:
@starbelly Sounds good! If you wanna chat more, please message me on our Slack :)

bshap93:
I'm pretty busy this week but I'd like to have this done before next.

julianguyen:
Not a problem @bshap93! Thanks for keeping us posted :)

[...]

PR #1727, opened 2020-04-21
Add Bookmark switch to Moments and Strategies

GayathriVenkatesh:
# Description
Added column 'bookmarked: boolean' to Moments and Strategies table.
Created a bookmark toggle on the Moments/Strategies page.

This is a subtask for issue #187 

## Screenshots
Screenshot from 2020-04-21 13-42-33

## Corresponding Issue
#1580 

---

### DIFF
--- a/app/helpers/moments_form_helper.rb
+++ b/app/helpers/moments_form_helper.rb
@@ -65,20 +65,15 @@ module MomentsFormHelper
   def moment_comment
     moment_input_props('comment', 'switch', 'comment.allow_comments')
       .merge(value: true,
-             uncheckedValue: false,
-             checked: @moment.comment,
-             info: t('comment.hint'),
-             dark: true)
+             uncheckedValue: false, checked: @moment.comment,
+             info: t('comment.hint'), dark: true)
   end
 
   def moment_publishing
-    { id: 'moment_publishing',
-      type: 'switch',
+    { id: 'moment_publishing', type: 'switch',
       label: t('moments.form.draft_question'),
-      dark: true,
-      name: 'publishing',
-      value: '0',
-      uncheckedValue: '1',
+      dark: true, name: 'publishing',
+      value: '0', uncheckedValue: '1',
       checked: !@moment.published? }
   end
 
--- a/app/helpers/moments_form_helper.rb
+++ b/app/helpers/moments_form_helper.rb
@@ -82,13 +82,24 @@ module MomentsFormHelper
       checked: !@moment.published? }
   end
 
+  def moment_bookmarked
+    moment_input_props('bookmarked', 'switch', 'moments.bookmark')
+      .merge(
+        value: true,
+        uncheckedValue: false,
+        checked: @moment.bookmarked,
+        info: t('comment.hint'),
+        dark: true
+      )
+  end
+
   def moment_form_inputs
     [
       moment_name, moment_why, moment_fix, moment_category, moment_mood,
       moment_strategy, get_viewers_input(
         @viewers, 'moment', 'moments', @moment
       ),
-      moment_comment, moment_publishing
+      moment_comment, moment_publishing, moment_bookmarked
     ]
   end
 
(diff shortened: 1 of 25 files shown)


app/helpers/moments_form_helper.rb
```diff
@@ -82,13 +82,23 @@ def moment_publishing
       checked: !@moment.published? }
   end
 
+  def moment_bookmarked
+    moment_input_props('bookmarked', 'switch', 'moments.bookmark')
```
julianguyen, line 86: I think we'll want to hide this behind a feature flag since it's part of a broader feature. That way it doesn't show up in production. I would just add these prop when `Rails.env.development?` is true.

app/helpers/strategies_form_helper.rb
```diff
@@ -79,6 +79,15 @@ def build_strategy_publishing(strategy)
     )
   end
 
+  def build_strategy_bookmarked(strategy)
+    build_switch_input(true, strategy.bookmarked, false)
```
julianguyen, line 83: Similar comment as above!

app/helpers/moments_form_helper.rb
GayathriVenkatesh: Sure!

app/helpers/moments_form_helper.rb
```diff
@@ -65,30 +65,36 @@ def moment_strategy
...
+
   def moment_form_inputs
     [
       moment_name, moment_why, moment_fix, moment_category, moment_mood,
       moment_strategy, get_viewers_input(
         @viewers, 'moment', 'moments', @moment
       ),
-      moment_comment, moment_publishing
+      moment_comment, moment_publishing,
+      Rails.env.development? ? moment_bookmarked : ''
```
julianguyen, line 97: Instead of returning an empty string if the env if production, I think it should be nil.

app/helpers/strategies_helper.rb
```diff
@@ -34,6 +34,7 @@ def strategy_form_inputs(strategy, viewers)
      get_viewers_input(viewers, 'strategy', 'strategies', strategy),
      build_strategy_comment(strategy), build_strategy_publishing(strategy),
      build_strategy_visible(strategy),
+     Rails.env.development? ? build_strategy_bookmarked(strategy) : '',
```
julianguyen, line 37: Similar comment as above, let's replace the empty string with `nil`.

app/helpers/moments_form_helper.rb
GayathriVenkatesh: Hi @julianguyen, some tests seem to fail when I make this change. I had noticed this previously and hence replaced it with an empty string. Is there another way to solve this?

app/helpers/moments_form_helper.rb
```diff
@@ -65,30 +65,36 @@ def moment_strategy
...
+
   def moment_form_inputs
     [
       moment_name, moment_why, moment_fix, moment_category, moment_mood,
       moment_strategy, get_viewers_input(
         @viewers, 'moment', 'moments', @moment
       ),
-      moment_comment, moment_publishing
+      moment_comment, moment_publishing,
+      Rails.env.development? ? moment_bookmarked : nil
```
julianguyen, line 97: Looks like you're right that `nil` is causing errors! You'll want to replace this with `{}` actually :)

app/helpers/strategies_helper.rb
```diff
@@ -34,6 +34,7 @@ def strategy_form_inputs(strategy, viewers)
      get_viewers_input(viewers, 'strategy', 'strategies', strategy),
      build_strategy_comment(strategy), build_strategy_publishing(strategy),
      build_strategy_visible(strategy),
+     Rails.env.development? ? build_strategy_bookmarked(strategy) : nil,
```
julianguyen, line 37: Same updated comment as above

app/helpers/moments_form_helper.rb
GayathriVenkatesh: Awesome! :)

GayathriVenkatesh:
Sure @julianguyen! I agree, this is a feature that a lot of users would like to have!
@Manasa2850 and I would love to work on this once our existing PRs get merged :)

julianguyen:
That would be great! :) It would be helpful to continue to scope the work based on what @LMulvey had laid things out!

PR #1749, opened 2020-05-13
Show CrisisPrevention modal when necessary in Moments [Content warning: crisis and suicide keywords]

julianguyen:
# Description

Piggybacking off the awesome work @AlineRibeiro and @S-Warmenhoven did for the Resource Recommendations feature, I'm adding a `CrisisPrevention` modal that appears when users post Moments that meet the criteria for this. This modal will only appear to authors of Moments posts.

## More Details

I think we can do more to make this feature more robust and "intelligent", but for now, this modal will show up if certain keywords show up in their posts. It also only shows up for posts that are created or updated on the user's current day.

In the future, it would be great to revisit prior efforts we made in the past to do sentiment analysis.

A pretty easy sub-feature to add to this is could be the ability to dismiss this modal from appearing (on a per Moment basis) using localStorage. I don't think we need to save this to the database. But we could if we decide to follow-up with users who are in a crisis.

# Screenshots

May-12-2020 20-06-14

image

---

GIF of Squidward being Squidward


julianguyen:
@S-Warmenhoven Thanks for the feedback! 

Yeah, I think it would be great to include self-harm keywords as well. There's definitely more to self-harm than cutting, though that's a classic example. I think it would be great to do more research on this and make a follow-up pull request :)

app/services/resource_recommendations.rb
```diff
@@ -1,9 +1,14 @@
 # frozen_string_literal: true
 
+CRISIS_PREVENTION_TAGS = %w[
```
bmwachajr, line 3: Since this is being used across all locale, we might to expand this list of tags to cater for the other locales!

bmwachajr: Maybe we could abstract it into its own yaml with multiple translations.

spec/services/resource_recommendations_spec.rb
```diff
@@ -1,21 +1,19 @@
...
-    let(:moment) {FactoryBot.build(:moment,
-                                       name: 'SELF-INJURY@',
-                                       why: 'text',
-                                       fix: 'text' )}
+  let(:moment) { FactoryBot.build(:moment, name: 'SELF-INJURY@', why: 'text', fix: 'text' ) }
+  let(:current_user) { create(:user2) }
+  subject(:resources) { ResourceRecommendations.new(moment: moment, current_user: current_user).call}
 
-    subject(:resources) { ResourceRecommendations.new(moment).call}
+  describe 'A test for resources method from ResourceRecommendations class' do
```
bmwachajr, line 6: Could we add a spec for the crisis prevention modal, I believe this feature will grow.

AlineRibeiro:
It looks great to me. Thank you so much for sharing your code with us.

spec/services/resource_recommendations_spec.rb
julianguyen: Yes that is a good idea! I'll do that!

app/services/resource_recommendations.rb
julianguyen: Ah yes these have been translated to our other locales! But I could move these translation keys into it's own yaml. I think I'll revisit that in another PR if we end up adding a lot more. Thanks for bringing this up!

PR #1764, opened 2020-05-29
Changed BaseContainer to function component to use React Hooks

thesid01:
# Description
I have Changed the Basecontainer Class component to functional component in order to use React Hooks


julianguyen:
@thesid01  Sorry for the delayed response! Taking a look now, it's been a very busy few weeks for me.

PR #1773, opened 2020-06-06
Changed class to function for using hooks

thesid01:
# Description
I have Changed the Notification Class component to functional component in order to use React Hooks

## More Details
This error was there when converting class to function so **I made notifications, alreadyMounted, and open variable in state optiona**l.

Annotation 2020-06-06 160237


client/app/widgets/Notifications/index.jsx
```diff
@@ -11,124 +11,128 @@ export type Props = {
...
   modalKey?: string,
   signedInKey?: number,
 };
 
-export class Notifications extends React.Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
-    this.state = { notifications: '', alreadyMounted: false, open: false };
-  }
+export const Notifications = ({
```
julianguyen, line 21: So I think you're getting those errors you identified because of how you exported things.

You'll want to do something like [this](https://github.com/ifmeorg/ifme/blob/master/client/app/components/Modal/index.jsx#L174).

client/app/widgets/Notifications/index.jsx
```diff
@@ -11,124 +11,128 @@ export type Props = {
 };
 
 export type State = {
-  notifications: string,
-  alreadyMounted: boolean,
-  open: boolean,
+  notifications?: string,
```
julianguyen, line 14: I'm curious why these props became optional.

client/app/widgets/Notifications/index.jsx
```diff
@@ -11,124 +11,128 @@ export type Props = {
...
+}: Props, {
+  notifications, alreadyMounted, open, modalKey, signedInKey,
+}: State) => {
+  // error handle : notifications, alreadyMounted, open not defined
+  let newnotifications = notifications;
+  let newalreadyMounted = alreadyMounted;
+  let newopen = open;
+  if (!newnotifications) newnotifications = '';
+  if (!newalreadyMounted) newalreadyMounted = false;
+  if (!newopen) newopen = false;
```
julianguyen, line 32: I don't think you need lines 27-32.

client/app/widgets/Notifications/index.jsx
```diff
@@ -11,124 +11,128 @@ export type Props = {
...
-    this.setState({ signedInKey });
+  const setBody = (paramsNotifications: string[]) => {
+    let updatedNotifications = '';
+    paramsNotifications.forEach((item: string) => {
+      updatedNotifications += `<div>${item}</div>`;
+    });
+    setnotificationsState(updatedNotifications);
+  };
+
+  // eslint no-use-before-define: error
```
julianguyen, line 54: Can we avoid using this and just move this variable before?

client/app/widgets/Notifications/index.jsx
thesid01: I am removing these lines.

client/app/widgets/Notifications/index.jsx
thesid01: Ok I am changing this

client/app/widgets/Notifications/index.jsx
thesid01: These lines were because I did something wrong, and it created an error. I have resolved that error by making them optional I am looking into this.

client/app/widgets/Notifications/index.jsx
thesid01: Since setBosy function is used only once in fetchNotifications we can merge this function inside that function

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,117 +18,122 @@ export type State = {
...
-    super(props);
-    this.state = { notifications: '', alreadyMounted: false, open: false };
-  }
+export const Notifications = ({
+  element,
+}: Props) => {
+  const [notifications, setnotifications] = useState('');
+  const [alreadyMounted, setalreadyMounted] = useState(false);
+  const [open, setopen] = useState(false);
+  const [signedInKey, setsignedInKey] = useState(0);
```
julianguyen, line 27: Since this value can be undefined, we can just initialize it to undefined by doing so `useState()`.
Same for `modalKey` below.

client/app/widgets/Notifications/index.jsx
julianguyen: I think we should keep them separate even though it's only used once. This makes understanding the logic easier since there's a lot going on in `setBody` already.

You can just change the ordering of these functions.

client/app/widgets/Notifications/index.jsx
thesid01: Yeah sure

client/app/widgets/Notifications/index.jsx
thesid01: Ok.

thesid01:
circleci build is failing but I haven't changed the logic or anything I am not getting why it is being failing. pusher-key is defined in meta I have checked through Inspect element.

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,117 +18,122 @@ export type State = {
...
-    super(props);
-    this.state = { notifications: '', alreadyMounted: false, open: false };
-  }
+export const Notifications = ({
+  element,
+}: Props) => {
+  const [notifications, setnotifications] = useState('');
+  const [alreadyMounted, setalreadyMounted] = useState(false);
+  const [open, setopen] = useState(false);
+  const [signedInKey, setsignedInKey] = useState(undefined);
```
julianguyen, line 27: You can just call `useState()` to set it as `undefined` by default.

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,117 +18,122 @@ export type State = {
...
+  const [modalKey, setmodalKey] = useState(undefined);
```
julianguyen, line 28: Same as above.

client/app/widgets/Notifications/index.jsx
```diff
@@ -18,117 +18,122 @@ export type State = {
...
+
+  const getPusherKey = (paramSignedInKey: number) => {
+    setsignedInKey(signedInKey);
     const metaPusherKey = Array.from(
       window.document.getElementsByTagName('meta'),
     ).filter((item) => item.getAttribute('name') === 'pusher-key')[0];
     const pusherKey = metaPusherKey.getAttribute('content');
     const pusher = new window.Pusher(pusherKey, { encrypted: true });
-    const channel = pusher.subscribe(`private-${signedInKey}`);
+    const channel = pusher.subscribe(`private-${paramSignedInKey}`);
```
julianguyen, line 55: Shouldn't this be `signedInKey`?

client/app/widgets/Notifications/index.jsx
julianguyen: This should be removed.

client/app/widgets/Notifications/index.jsx
```diff
@@ -45,90 +54,85 @@ export class Notifications extends React.Component<Props, State> {
     const channel = pusher.subscribe(`private-${signedInKey}`);
```
julianguyen, line 54: So the issue is that `signedInKey` can either be undefined or a number. If it's undefined, you can't do string interpolation of this value. So to bypass this type issue, I would just use `paramSignedInKey` directly (which should always be a number).

Issue #1748, opened 2020-05-12
Cursor doesn't change when hovering over medication details


enoralecuyer:
# Description

The cursor doesn't change when hovering over the details of a new medication. To make this feature more accessible, I suggest to change the cursor to a type **pointer**.

On the New Medication page, the change is needed for:
- Strength/mg
- Total/tablets
- Dosage/tablets
- mm/dd/yyyy
- two check boxes Refill reminder email and Daily reminder email

# Screenshots
![Inked2020-05-12 15_31_15-Greenshot_LI](https://user-images.githubusercontent.com/48727972/81752500-688f0580-9466-11ea-9fe6-60c57a3a0995.jpg)

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

julianguyen:
I mentioned this already in Slack but I think we actually don't want the cursor type to be a pointer here.

The pointer cursor type should only be used on elements that are links. This custom select component is not a link. The native `<Select>` element actually uses the default cursor.

enoralecuyer:
In terms of accessibility and usability, it is not easy to recognize that it's a multiple-choice selection. I actually missed it the first time I went to this page. 

If the decision is made to not change the pointer, it would be good to come up with another solution to make this feature more accessible and help the user not miss this important step.

What about having all the choices visible in a single line (either horizontal or vertical) and the user can select the option of their liking?

julianguyen:
@enoralecuyer That's a great observation! I think improving the UX here is a good call. We would either:

1) Add a label to the existing component to indicate that there are multiple options
2) Change the component so that it's checkboxes

What do you think?

GuillermoCoding:
Just joining in the conversation and following up on @julianguyen. For the second option you suggested, do you think radio buttons would be more appropriate instead of check boxes ? since the use should choose between`mg` OR `ml` and not have the ability to select multiple options ?

enoralecuyer:
@GuillermoCoding I love your idea of using radio buttons! They are clear, easy to use and restrict to one option only. That sounds perfect to me!

GuillermoCoding:
@julianguyen any more thoughts on this?  I can start tackling this feature

[...]

PR #1776, opened 2020-06-08
[#1748] Adds InputRadio form component to new/edit medication form

GuillermoCoding:
# Description
* According to the discussion on #1748, the new/edit medication form has some accessibility issues when selecting the `strength unit`, `total unit`, and `dosage unit` inputs. The proposed solution consisted of changing these inputs to radio buttons in order to create a better user experience.
* Creates an `InputRadio` form component and adds this component to the new/edit medication form.
* Replaces the `InputSelect` form components in the medication form used for `strength unit`, `total unit` and `dosage unit` with new `InputRadio` component.
* Adds unit tests for `InputRadio` component.

## More Details
* Modifies `medications_form_helper.rb` to use radio inputs for `medication_strength_unit` and `medication_unit_field`.
* Adds more locales to `config/locales/en.yml `

## Corresponding Issue

#1748 

# Screenshots
### Before
image

### After
image

---

### DIFF
--- a/client/app/components/Input/InputRadioGroup.jsx
+++ b/client/app/components/Input/InputRadioGroup.jsx
@@ -1,0 +1,35 @@
+// @flow
+import React, { useState } from 'react';
+import css from './InputRadioGroup.scss';
+import type { Option } from './utils';
+
+export type Props = {
+  name?: string,
+  options: Option[],
+  value?: any,
+}
+
+export function InputRadioGroup({
+  name,
+  options,
+  value: propValue,
+}: Props) {
+  return (
+    <div role="radiogroup" className={css.wrapper}>
+      {options.map((option: Option) => (
+        <div key={option.value} className={css.content}>
+          <label className={css.label} htmlFor={option.id}>
+            {option.value}
+            <input
+              type="radio"
+              id={option.id}
+              name={name}
+              value={option.value}
+              defaultChecked={option.value == propValue}
+            />
+          </label>
+        </div>
+      ))}
+    </div>
+  );
+}
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -219,10 +219,9 @@ en:
     event_summary: 'Refill for %{name}.'
     form:
       strength: Strength
-      strength_placeholder: e.g. 100
-      total_placeholder: e.g. 30
+      strength_unit: 'Strength Unit'
       dosage: dosage
-      dosage_placeholder: e.g. 2
+      dosage_unit: 'Dosage Unit'
       refill: refill
       comments_hint: 'Add any additional comments here, e.g. logging your symptoms'
       add_to_google_cal: Google Calendar reminder
@@ -237,6 +236,7 @@ en:
         refilled.
       daily_reminder_hint: We'll email you a daily reminder to take this medication
       total: Total
+      total_unit: 'Total Unit'
     new: New Medication
     units:
       display: '%{count} %{unit}'
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -230,7 +230,7 @@ fr:
     event_summary: 'Recharge pour %{name}.'
     form:
       strength: Concentration
-      stregth_unit: 'Unité de concentration'
+      strength_unit: 'Unité de concentration'
       strength_placeholder: ex. 100
       total_placeholder: ex. 30
       dosage: dosage
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -230,9 +230,11 @@ fr:
     event_summary: 'Recharge pour %{name}.'
     form:
       strength: Concentration
+      stregth_unit: 'Unité de concentration'
       strength_placeholder: ex. 100
       total_placeholder: ex. 30
       dosage: dosage
+      dosage_unit: 'Unité de dosage'
       dosage_placeholder: ex. 2
       refill: renouvellement
       comments_hint: >-
@@ -254,6 +256,7 @@ fr:
         Nous vous enverrons un email de rappel journalier pour la prise de ce
         médicament.
       total: Total
+      total_unit: 'Unité totale'
     new: Nouvelle médication
     units:
       display: '%{count} %{unit}'
--- a/app/helpers/medications_form_helper.rb
+++ b/app/helpers/medications_form_helper.rb
@@ -121,7 +121,8 @@ module MedicationsFormHelper
 
   def medication_strength_unit
     {
-      type: 'select',
+      type: 'radio',
+      label: t('medications.form.strength_unit'),
       value: @medication.strength_unit || t('medications.units.mg'),
       options: [
         medication_type_unit_mg('strength'),
@@ -198,16 +199,21 @@ module MedicationsFormHelper
     }
   end
 
+  def medication_unit_options(type)
+    [
+      medication_type_unit_tablets(type),
+      medication_type_unit_mg(type),
+      medication_type_unit_ml(type)
+    ]
+  end
+
   def medication_unit_field(type)
     {
-      type: 'select',
+      type: 'radio',
+      label: t("medications.form.#{type}_unit"),
       value: @medication["#{type}_unit"] ||
         t('medications.units.tablets.other'),
-      options: [
-        medication_type_unit_tablets(type),
-        medication_type_unit_mg(type),
-        medication_type_unit_ml(type)
-      ]
+      options: medication_unit_options(type)
     }.merge(medication_basic_props("#{type}_unit"))
   end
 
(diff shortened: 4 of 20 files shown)


config/locales/en.yml
```diff
@@ -219,10 +219,12 @@ en:
     event_summary: 'Refill for %{name}.'
     form:
       strength: Strength
-      strength_placeholder: e.g. 100
-      total_placeholder: e.g. 30
+      strength_unit: 'Strength Unit'
+      strength_placeholder: 'e.g. 100'
+      total_placeholder: 'e.g. 30'
```
julianguyen, line 224: I don't think we need all these new placeholders. These fields must be numeric and users would be filling in this information from their prescription. But this was a good suggestion! :)

client/app/components/Input/InputRadio.scss
```diff
@@ -0,0 +1,20 @@
+@import "../../styles/_global.scss";
+
+.wrapper {
+  display: flex;
+  flex-direction: row;
+
+  .content {
+    margin-right: 15px;
```
julianguyen, line 8: You can replace this with `$size-16` which comes from [here](https://github.com/ifmeorg/ifme/blob/master/app/assets/stylesheets/base/_sizes.scss#L10).

client/app/components/Input/InputRadio.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  input {
+    margin-right: 5px;
```
julianguyen, line 12: You can replace this with `$size-6` which comes from [here](https://github.com/ifmeorg/ifme/blob/master/app/assets/stylesheets/base/_sizes.scss#L5).

client/app/components/Input/InputRadio.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  label {
+    all: unset;
+    color: rgba(109, 8, 57, 0.9);
```
julianguyen, line 17: You can replace this with `$mulberry-90` which comes from [here](https://github.com/ifmeorg/ifme/blob/master/app/assets/stylesheets/base/_colors.scss#L26).

client/app/components/Input/__tests__/InputRadio.spec.jsx
```diff
@@ -0,0 +1,62 @@
...
+        name={name}
+        id={id}
+        value={value}
+        options={options}
+        onChange={someEvent}
+      />,
+    );
+    expect(
+      wrapper.find('input[id="some-option-one-id"][type="radio"]').props().checked,
+    ).toBeTruthy();
```
julianguyen, line 30: I think you can be more explicit here and use the `.toBeTrue()` assertion since we expect booleans. Same case below too.

client/app/components/Input/InputRadio.jsx
```diff
@@ -0,0 +1,47 @@
...
+  onChange?: Function,
+}
+
+export function InputRadio({
+  name,
+  options,
+  onChange,
+  value: propValue,
+}: Props) {
+  const [value, setValue] = useState<any>(propValue);
```
julianguyen, line 19: I think we can make this uncontrolled instead of controlled. By that I mean, we can rely on the native `<input>` element to update the value for us instead of handling it ourselves in state.

Here's an [article](https://itnext.io/controlled-vs-uncontrolled-components-in-react-5cd13b2075f9) on uncontrolled vs. controlled components if you need it.

client/app/components/Input/InputRadio.jsx
```diff
@@ -0,0 +1,47 @@
+// @flow
+import React, { useState } from 'react';
+import css from './InputRadio.scss';
+import type { Option } from './utils';
+
+export type Props = {
+  name?: string,
+  value?: any,
```
julianguyen, line 8: We can replace `any` with `boolean`. If `value` is undefined, we can assume it's `false`.

GuillermoCoding: In this case, InputRadio.jsx represents a collection of radio buttons so value represents which radio button is checked (tables, mg, or ml) and not whether a singe radio button is checked or not. This suggestion made me think that maybe the naming of this component is misleading since it does not represent a single radio button but instead of a collection of radio buttons were value represents which radio button is checked.

GuillermoCoding: I can change the name of this component to `InputRadioGroup.jsx`? Let me know your thoughts

client/app/components/Input/InputRadio.jsx
GuillermoCoding: Thanks for this suggestion! Should I still make this change even though `InputRadio.jsx` represents a collection of radio buttons instead of a single radio button ?

config/locales/en.yml
```diff
@@ -219,10 +219,12 @@ en:
     event_summary: 'Refill for %{name}.'
     form:
       strength: Strength
-      strength_placeholder: e.g. 100
-      total_placeholder: e.g. 30
+      strength_unit: 'Strength Unit'
```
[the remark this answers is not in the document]
GuillermoCoding: I have added these to the other locales using Google translate on my latest commit👍

config/locales/en.yml
GuillermoCoding: I have removed these placeholders on my latest commit

client/app/components/Input/__tests__/InputRadio.spec.jsx
GuillermoCoding: Getting an error that `.toBeTrue()` is not a function using the current `jest-jasmine` testing library 🤔 but I did change this to `.toBe(true)` if that also is explicit in this case?

julianguyen: Oh that works too! :)

client/app/components/Input/InputRadio.jsx
julianguyen: Hmm good question. I think you should make that change. InputRadio should just refer to one radio button. We could make some kind of collection component later.

client/app/components/Input/InputRadio.jsx
julianguyen: Oh sorry I read this after the previous comment. Yeah I think you should rename it! I don't think it needs to be controlled component if you wrap things in a `<fieldset>`.

client/app/components/Input/InputRadio.jsx
GuillermoCoding: Sounds good!, I have changed the name to `InputRadioGroup.jsx` on my latest commit.

client/app/components/Input/InputRadio.jsx
GuillermoCoding: I have changed the name and made it uncontrolled on my latest commit 👍

client/app/components/Input/InputRadioGroup.jsx
```diff
@@ -0,0 +1,35 @@
...
+  value?: any,
+}
+
+export function InputRadioGroup({
+  name,
+  options,
+  value: propValue,
+}: Props) {
+  return (
+    <div className={css.wrapper}>
```
julianguyen, line 18: This should be replaced with a `<fieldset>`.

GuillermoCoding: Sounds good! 👍 I have added `<fieldset>` around this component. It did introduce some unwanted CSS styling that I disabled but it also prevents `display: flex` and `flex-direction: row` from being applied to it. That is why I had to keep the wrapper div to allow for `flex-direction: row`. Otherwise, items will be displayed in a column. Let me know your thoughts on this

julianguyen: Hmm okay, I did some more research and I think it should suffice to put `role="radiogroup"` on this div.

GuillermoCoding: Cool, I have made this change! Confused why the circleci build is not happening 🤔 I have squashed my commits. I also synced with the upstream master as explained in the wiki. I have rebased my branch with the master so it is on the top. None of this seems to reset the circleci build. I did some Google and it might be caused by something enabled under the repo's "settings" tab ?

config/locales/fr.yml
```diff
@@ -230,9 +230,11 @@ fr:
     event_summary: 'Recharge pour %{name}.'
     form:
       strength: Concentration
+      stregth_unit: 'Unité de concentration'
```
julianguyen, line 233: This should be `strength_unit`. This is causing a test in `./spec/compare_locales_support_spec.rb` to fail.

GuillermoCoding: Thanks, fixed this in my latest commit 👍

Issue #1758, opened 2020-05-21
Migrate from Enzyme to React Testing Library


julianguyen:
# Description

We should migrate to React Testing Library as it's officially recommended by React and it's a lot more usable than Enzyme.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

alekpentchev:
I can work on this! Does want to rewrite every component’s test file to use React Testing Library? Or do we want to focus on a specific set of components?

julianguyen:
Yeah we'll want to get rid of Enzyme and replace it with React Testing Library. The good thing is that we don't need to rely on element classes or IDs to identify them in our tests. We can identify elements by their appropriate aria labels, text content, etc.

Let me know if you have any questions or need help along the way. I use this library in my day job and have some experience with it :P

alekpentchev:
Thanks Julia for your support! I will probably need some onboarding in the topic of this library because I'm quite new to it. But I'm eager to work on it :)

Btw If someone else would like to jump in and to this in parallel, I'm absolutely ok with it!

PR #1780, opened 2020-06-13
Issue_1758 - Migration from Enzyme to React Testing Library pt. 1

alekpentchev:
# Description

First few components spec files rewritten from Enzyme to React Testing Library.

## More Details

The main purpose of this PR is to confirm and adjust the direction of the process of migrating from Enzyme to RTL.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1758

# Screenshots

N/A

---


client/app/components/Accordion/__tests__/Accordion.spec.jsx
```diff
@@ -1,7 +1,8 @@
 // @flow
-import { shallow } from 'enzyme';
 import React from 'react';
 import { Accordion } from '../index';
+import { render, screen, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
```
julianguyen, line 5: Is this necessary? If so I think we can place [/client/app/libs/testHelper.js](https://github.com/ifmeorg/ifme/blob/master/client/app/libs/testHelper.js) so that it's globally available.

client/app/components/Accordion/index.jsx
```diff
@@ -30,7 +30,7 @@ export const Accordion = ({
   const [open, setOpen] = useState(!!openProp);
 
   const displayContent = () => (
-    <div className={`${open ? 'accordionContent' : css.accordionClose}`}>
+    <div className={`${open ? 'accordionContent' : css.accordionClose}`} role="list">
```
julianguyen, line 33: So we shouldn't be using this list role on this accordion. It would be better to identify this element by an aria-label.

alekpentchev:
> This is a really great start! Thank yooou sooo much for working on this :)
> 
> Looks like there are linting errors, so I recommend running `(cd client && yarn lint)` in your local dev environment to fix this.

Thank you very much! Yeah, I've noticed them and will fix them tomorrow, probably. Besides, what do you think about the techniques I've used here and the methods of querying elements? Would you improve or replace something?

client/app/components/Accordion/__tests__/Accordion.spec.jsx
alekpentchev: Do you mean line 5? Because when I removed it from the spec file and tried to put it inside testHelper I get an error when running the tests (even after using another defineProperty for expect):

image

julianguyen: Interesting, what happens if you import `import '@testing-library/jest-dom'` instead? Does that work?

client/app/components/Accordion/index.jsx
```diff
@@ -30,7 +30,10 @@ export const Accordion = ({
   const [open, setOpen] = useState(!!openProp);
 
   const displayContent = () => (
-    <div className={`${open ? 'accordionContent' : css.accordionClose}`}>
+    <div
+      className={`${open ? 'accordionContent' : css.accordionClose}`}
+      role="list"
```
julianguyen, line 35: We should not be using the list role here. Let's use `role="region"` instead.

alekpentchev: 1. Extending expect globally. After a quick research it turned out that `"@testing-library/jest-dom/extend-expect"` should be specified in package.json under `setupFilesAfterEnv` field. Then it worked as expected.

2. Sure, will do. Role adjusted to `region`

alekpentchev:
@julianguyen Circleci:build is failing due to some vulnerabilities within some ruby-related packages. Should we handle this here or in a separate PR?

julianguyen:
Oof yeah, I can make a PR for this! Sorry about that.

alekpentchev:
> Excellent work! Thank you so much for taking this on!

Thank you! Now I have a green light to continue my work as until now. :) I can wait for the circleci issue to be resolved or can you merge this PR (I don't have such option)?

julianguyen:
Hey! Sorry I didn't get around to making a PR this weekend. Doing it right now with the gem updates. Then you can rebase and we can merge this. Great work!! :)

julianguyen:
You should be able to rebase now :)

PR #1841, opened 2020-08-29
Refactor membership controller spec to membership request spec #1816

bennpham:
# Description

Refactored `spec/controllers/groups/memberships_controller_spec.rb` to `spec/request/groups/membership_spec.rb`.

## More Details

- refactors `let!` in between `context` blocks into `it` blocks as an `Arrange` step instead
- adds **Arrange**, **Act**, **Assert** comments to easier see each steps of the [AAA unit test pattern](http://wiki.c2.com/?ArrangeActAssert)
- removes StubUserHelper since the following code is not supported by request specs
  - replaces StubUserHelper with `sign_in` method instead

```ruby
allow(request.env['warden']).to receive(:authenticate!).and_return(user)
allow(controller).to receive(:current_user).and_return(user)
```

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1816

---


spec/requests/groups/membership_spec.rb
```diff
@@ -1,117 +1,132 @@
...
-
-  before do
-    create_current_user
-  end
-
-  describe '#create' do
-    context 'new group member for existing group' do
-      let!(:leader_user) { create :user2 }
-      let!(:group_leader) { create :group_member, user_id: leader_user.id, leader: true }
-      let!(:group) { group_leader.group }
```
bennpham, line 14: There's a handful of `let` statements declared throughout the controller spec. I like let when there's a few it blocks or context blocks that uses the same defined variable, but for one it block, I think it's harder to read and can be an anti pattern. Also with the behavior of being threadsafe or not based on config as mentioend in [the rubydoc for rspec-core](https://www.rubydoc.info/github/rspec/rspec-core/RSpec%2FCore%2FMemoizedHelpers%2FClassMethods%3Alet), I feel like we should use it when needed. 

I've seen some flakey tests from a few disorganized `let` or ambigious tests.

I personally avoid using `let!` with the exclamation point unless I have to since that will always be evaluated.

---

I'm open to any opinions or counterexamples. :)

spec/requests/groups/membership_spec.rb
```diff
@@ -1,117 +1,132 @@
...
+  describe '#create' do
+    context 'new group member for existing group' do
+      it 'adds new member' do
+        # Arrange
+        leader_user = create :user2
+        group_leader = create :group_member, user_id: leader_user.id, leader: true
+        group = group_leader.group
+
+        # Act
+        post group_membership_path(group_id: group.id)
```
bennpham, line 17: Normally for most request specs I've wrote, I could've just done something like the following:
```suggestion
        post group_membership_path(group)
```

Rails magic would usually assume I'm passing the `group.id` to `:id` for that specified route. 

For some reason I'll get a failed test with the following instead:

```
Expected response to be a redirect to <http://www.example.com/groups/test-group> but was a redirect to <http://www.example.com/groups>.
       Expected "http://www.example.com/groups/test-group" to be === "http://www.example.com/groups".
```

For the other tests, if I just pass the model, I might get 500 error. I'm assuming this could depend on the `route.rb` setup, but base on the current setup, I can define the route's path variables and pass the corresponding `id` to the paths I'd like it to go to unless I'm missing something.

spec/requests/groups/membership_spec.rb
julianguyen: Thanks for sharing this insight! That makes a lot of sense!

spec/requests/groups/membership_spec.rb
julianguyen: Ah yeah the reason for this is cause we're expecting `group_id` instead of `id` like you pointed out. We could refactor [controller](https://github.com/ifmeorg/ifme/blob/master/app/controllers/groups/memberships_controller.rb) and the corresponding route so we're expecting an `id`. I don't think this is a pressing thing to fix to but we could make a note of it. Thanks for pointing this out!

PR #1842, opened 2020-09-01
Refactor google calendar event controller spec to google calendar event request spec

bennpham:
# Description

Refactored `spec/controllers/meetings/google_calendar_event_controller_spec.rb` to `spec/request/meetings/google_calendar_event_spec.rb`.

## More Details

- replaces 2 `let!` in `describe` block to just regular `let`
  - doesn't seem like `calendar_event` or `exception_message` needs to be initialized before the test starts (unless I'm missing something and ended up making the test flaky without knowing it after refactoring)
- removes `let!(:remove_response) { double }` since it seems to be dead code
- adds **Arrange**, **Act**, **Assert** comments to easier see each steps of the [AAA unit test pattern](http://wiki.c2.com/?ArrangeActAssert)

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1817

---

### DIFF
--- a/spec/requests/meetings/google_calendar_event_spec.rb
+++ b/spec/requests/meetings/google_calendar_event_spec.rb
@@ -4,8 +4,8 @@ RSpec.describe 'GoogleCalendarEvent', type: :request do
   let(:user) { create(:user_oauth) }
   let(:meeting) { create(:meeting) }
   let(:calendar_uploader) { double }
-  let!(:calendar_event) { double(id: 'someid') }
-  let!(:exception_message) { 'Exception message' }
+  let(:calendar_event) { double(id: 'someid') }
+  let(:exception_message) { 'Exception message' }
 
   before do
     sign_in user
@@ -63,7 +63,6 @@ RSpec.describe 'GoogleCalendarEvent', type: :request do
 
   describe '#destroy' do
     let!(:meeting_member) { create(:meeting_member, user_id: user.id, meeting_id: meeting.id, google_cal_event_id: calendar_event.id) }
-    let!(:remove_response) { double }
 
     context 'success' do
       it 'calls calendar_uploader#delete_event' do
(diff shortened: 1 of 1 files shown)


spec/requests/meetings/google_calendar_event_spec.rb
```diff
@@ -1,13 +1,14 @@
...
-RSpec.describe ::Meetings::GoogleCalendarEventController, type: :controller do
-  include_context(:logged_in_user)
+RSpec.describe 'GoogleCalendarEvent', type: :request do
   let(:user) { create(:user_oauth) }
   let(:meeting) { create(:meeting) }
   let(:calendar_uploader) { double }
-  let!(:calendar_event) { double(id: 'someid') }
-  let!(:exception_message) { 'Exception message' }
+  let(:calendar_event) { double(id: 'someid') }
+  let(:exception_message) { 'Exception message' }
```
bennpham, line 8: Not sure if I'm missing anything here, but I don't think this needs to be `let!` and just `let` seems to be enough. The test still passes this way and there doesn't seem to be any model here that needs to be created and initialize before each test. 

On the other hand, `let!(:meeting_member)` below is required since meeting_member needs to be created before the **act** step to be altered after. If we don't use `let!` there, `meeting_member` will be created during the **assert** step `expect(meeting_member.reload.google_cal_event_id).to eq(calendar_event.id)` and would've have never been altered to meet the proper assertion.

spec/requests/meetings/google_calendar_event_spec.rb
```diff
@@ -62,15 +71,17 @@
 
   describe '#destroy' do
     let!(:meeting_member) { create(:meeting_member, user_id: user.id, meeting_id: meeting.id, google_cal_event_id: calendar_event.id) }
-    let!(:remove_response) { double }
```
bennpham, line 65: This variable doesn't seem to be used anywhere so I think this is dead code ☠️ ⚰️

spec/requests/meetings/google_calendar_event_spec.rb
julianguyen: Good catch, thanks so much for cleaning this up!

PR #1845, opened 2020-09-25
Refactor Invitation controller spec to invitation request spec

bennpham:
# Description

Refactored `spec/controllers/users/invitations_controller_spec.rb` to `spec/requests/users/invitation_spec.rb`.

## More Details

- cleans up some of the tests with other RSpec matcher 
- adds some **Assume** steps in case false positives happen
- moves `let` variables into context blocks that uses it
- fixes `describe '#update' do` block by using `raw_invitation` token
- adds **Arrange**, **Act**, **Assume**, **Assert** comments to easier see each steps of the [AAA unit test pattern](http://wiki.c2.com/?ArrangeActAssert)

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1818

---

### DIFF
--- a/spec/requests/users/invitation_spec.rb
+++ b/spec/requests/users/invitation_spec.rb
@@ -4,10 +4,10 @@ RSpec.describe 'Invitation', type: :request do
   let(:user) { create(:user, name: 'Jane') }
   let(:invitee1) { 'invited_friend@gmail.com' }
   let(:invitee2) { 'other_friend@gmail.com' }
-  let(:invalid_email) { 'invalid_email.com' }
-  let(:invite_one_friend) { post user_invitation_path, params: { user: { email: invitee1 } } }
 
   describe '#create' do
+    let(:invalid_email) { 'invalid_email.com' }
+
     context 'when a user is not signed in' do
       before { post user_invitation_path }
       it_behaves_like :with_no_logged_in_user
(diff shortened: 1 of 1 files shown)


spec/requests/users/invitation_spec.rb
```diff
@@ -1,153 +1,186 @@
...
         it 'only creates a new User for the valid email' do
-          newest_users = User.order(invitation_sent_at: :desc).last(2)
-          emails = []
-          newest_users.each { |u| emails << u.email }
-          expect(emails.sort).to include(invitee1)
-          expect(emails.sort).not_to include(invalid_email)
+          # Assume
+          emails = User.pluck(:email)
+          expect(emails).not_to include(invitee1)
+          expect(emails).not_to include(invalid_email)
```
bennpham, line 102: I added an **Assume** step here  to make sure that there aren't any users with `invitee1` and `invalid_email` emails before `create` invitations path was called.

spec/requests/users/invitation_spec.rb
```diff
@@ -1,153 +1,186 @@
...
 
   describe '#update' do
-    include_context :logged_in_user
     let(:password) { 'passworD@99' }
-
-    before(:each) do
-      invite_one_friend
-    end
+    let(:name) { 'New Person' }
+    let(:invited_user) { User.invite!({ email: invitee1 }, user) }
```
bennpham, line 132: This fixed the failing test when I moved this over from controller_spec to request_spec since I needed the `raw_invitation_token` from the user itself. If I use a route directly instead of the model method, the `raw_invitation_token` would have been deleted at that point and I won't have it to reference in the params for the test.

bennpham:
I think we need to upgrade from Rails 6.0.3.2 to 6.0.3.3 to fix the Action View vulnerability?

julianguyen (approved):
Excellent work! This is so much more cleaner!

PR #1846, opened 2020-09-26
Update to Rails 6.0.3.3

julianguyen:
# Description

Updating to Rails 6.0.3.3 to get a patch for a security issue in Actionview. In addition to that, I updated other gems too.

I also updated our docs to reflect this change.

---


bennpham (approved):
Looks good. I think you can also run `bundle update rails` and it'll automatically get the latest build version number from `6.0.3` and just update the `Gemfile.lock` without having you modify the `Gemfile` itself. 

Up to you for preferences though.

Issue #1815, opened 2020-08-24
Refactor RSpec Controller Specs to Request Specs


bennpham:
# Description

Looking through the project, I notice that we use controller specs rather than request specs. Since we're on Rails 6 and reading through this [Stackoverflow discussion in regards to controller specs vs request specs](https://stackoverflow.com/questions/40851705/controller-specs-vs-request-specs), it seems that controller specs are obsolete now. 

Some pros and cons listed in [this medium article](https://medium.com/just-tech/rspec-controller-or-request-specs-d93ef563ef11) mentions that controller specs methods such as `assigns` and `assert_template` are deprecated now as well and request spec is what is recommended by Rspec and test the entire stack. Request spec since Rails 5 are now faster as well.

Refactoring all the controller specs would be a large PR on its own though. 

Request spec documentation on RelishApp: https://relishapp.com/rspec/rspec-rails/v/4-0/docs/request-specs/request-spec

---

- [x] https://github.com/ifmeorg/ifme/issues/1816 
Refactor spec/controllers/groups/memberships_controller_spec.rb to spec/request/groups/membership_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1817 Refactor spec/controllers/meetings/google_calendar_event_controller_spec.rb to spec/request/meetings/google_calendar_event_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1818 Refactor spec/controllers/users/invitations_controller_spec.rb to spec/request/users/invitation_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1819 Refactor spec/controllers/allies_controller_spec.rb to spec/request/allyship_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1820 Refactor spec/controllers/care_plan_contacts_controller_spec.rb to spec/request/care_plan_contact_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1821 Refactor spec/controllers/care_plan_controller_spec.rb to spec/request/care_plan_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1822 
Refactor spec/controllers/categories_controller_spec.rb to spec/request/category_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1823 Refactor spec/controllers/comment_controller_spec.rb to spec/request/comment_spec.rb

- [x] https://github.com/ifmeorg/ifme/issues/1824 
[...]

PR #1859, opened 2020-10-02
Refactor Strategies controller spec to strategy request spec

andrew-schutt:
# Description

This refactors the tests associated with the `StrategiesController` to use RSpec `request` type tests instead of `controller`

## More Details

Couple things of note:
- I used the built-in Rails routes helpers
- Instead of `include_context :logged_in_user` -> `before { sign_in user }` is used
- I eliminated some of the `before` blocks but could bring them back (I prefer explicitness 😸 ) 
- Added the ability to set `:focus` on a test for running _only_ the test focused.
- Moved `tagged` above `print_reminders` as `tagged` is a route in the controller vs an `include` in the controller

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1815 <- parent issue
https://github.com/ifmeorg/ifme/issues/1840 <- directly related to this refactor

---

### DIFF
--- a/spec/requests/strategies_spec.rb
+++ b/spec/requests/strategies_spec.rb
@@ -87,19 +87,20 @@ describe 'Strategy', type: :request do
 
       context 'when the request format is html' do
         it 'redirects to the strategies_path' do
-          post :premade
+          post premade_strategies_path
           expect(response).to redirect_to(strategies_path)
         end
 
         it 'returns status code 204 for json format' do
-          post :premade, as: :json
+          headers = { "ACCEPT" => "application/json" }
+          post premade_strategies_path, headers: headers
           expect(response).to have_http_status(204)
         end
       end
     end
 
     context 'when the user is not logged in' do
-      before { post :premade }
+      before { post premade_strategies_path }
       it_behaves_like :with_no_logged_in_user
     end
   end
--- a/spec/requests/strategies_spec.rb
+++ b/spec/requests/strategies_spec.rb
@@ -8,9 +8,7 @@ describe 'Strategy', type: :request do
     let(:strategy) { create(:strategy, name: 'test', user: user) }
 
     context 'when the user is logged in' do
-      before do
-        sign_in user
-      end
+      before { sign_in user}
 
       context 'when search params are provided' do
         it 'assigns @strategies' do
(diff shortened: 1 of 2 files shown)


tyler-wel (commented):
Overall looks good to me 🤔 pretty long spec lol

One thing I personally like to use is subject
https://www.betterspecs.org/#subject

What are your opinions on it?

andrew-schutt:
👋  @tyler-wel thanks for reviewing! 😄 

Yeah, I did not realize how long of a spec I was signing up to refactor 😆 

In regards to `subject`, I am familiar with it at all. Do you have an example of the usage within this spec?

tyler-wel:
> 👋 @tyler-wel thanks for reviewing! 😄
> 
> Yeah, I did not realize how long of a spec I was signing up to refactor 😆
> 
> In regards to `subject`, I am familiar with it at all. Do you have an example of the usage within this spec?

Yaa, I did the same thing at first before switching which spec I assigned myself to 😂 

I personally used `subject` in my refactor of `reports_spec` and my current implementation of comments spec here #1861 (even though I personally didn't use betterspecs `it { is_expected.to  }` 🤔 maybe I need to go back and change mine lol)

spec/requests/strategies_spec.rb
```diff
@@ -1,127 +1,128 @@
...
     context 'when the user is logged in' do
-      include_context :logged_in_user
+      before { sign_in user}
 
       context 'when search params are provided' do
-        before { get :index, params: { search: 'test' } }
-
         it 'assigns @strategies' do
+          get strategies_path, params: { search: 'test' }
           expect(assigns(:strategies)).to eq([strategy])
```
tyler-wel, line 16: Hmm I guess it really is no problem, because there are other render/data expectations in this context, but I'm not 100% sold on `assigns` matcher for request specs 🤔 
I've always believed it's best to just test the input and output for each possible case. But I guess it isn't hurting anyone still being there 🤷 lol ... especially if we're not rendering json, I guess that's one valid way to check that the data was actually assigned...hmmmm TIL

spec/requests/strategies_spec.rb
```diff
@@ -1,127 +1,128 @@
...
+describe 'Strategy', type: :request do
   let(:user) { create(:user) }
   let(:strategy) { create(:strategy, user: user) }
 
   describe '#index' do
     let(:strategy) { create(:strategy, name: 'test', user: user) }
 
     context 'when the user is logged in' do
-      include_context :logged_in_user
+      before { sign_in user}
```
julianguyen, line 11: Nice!!

spec/requests/strategies_spec.rb
julianguyen: Yeah in this [article](https://medium.com/just-tech/rspec-controller-or-request-specs-d93ef563ef11) that @bennpham linked in the original issue, it says that `assigns` and `assert_template` is deprecated. Is there another way to check for the value of instance variables?

spec/spec_helper.rb
```diff
@@ -90,6 +90,7 @@
   config.infer_spec_type_from_file_location!
 
   config.before { ActionMailer::Base.deliveries.clear }
+  config.filter_run_when_matching :focus
```
julianguyen, line 93: Good call!

julianguyen (dismissed):
Thanks for taking this on! :D Great work!

Everything looks good to me but I'm curious whether the tests for "assigns @strategies" can be tested without assigns / is truly necessary.

spec/requests/strategies_spec.rb
tyler-wel: As I mentioned to someone else on Slack, one way to check is check the actual response body (in the case of json) and confirm the returned items are what's expected.

```ruby
expect(JSON.parse(response.body['some_id']).to eq(expected_value)
```

For html expectations I'm a little more rusty, as most ofvmy experience in JSON apis

andrew-schutt:
I will get the `assigns` checks cleaned up quickly as well (today 10/3) before we merge

spec/spec_helper.rb
andrew-schutt: Thanks! I will be sure to update the wiki too

spec/requests/strategies_spec.rb
andrew-schutt: I used the _strategy_ in that article in which I expect an attribute to be included in the body of the response.

julianguyen (approved):

PR #1861, opened 2020-10-03
Refactor comments spec + controller fixes

### LINKED ISSUE #1823: Refactor spec/controllers/comment_controller_spec.rb to spec/request/comment_spec.rb
# Description

Controller Specs had been obsolete since Rails 5. It is best practice to create request spec instead of controller spec as recommended by the Rspec team.

Recommended Reading:

- [Stackoverflow Discussion in regards to Controller Specs vs Request Specs](https://stackoverflow.com/questions/40851705/controller-specs-vs-request-specs)
- [Medium article discussing request and controller specs](https://medium.com/just-tech/rspec-controller-or-request-specs-d93ef563ef11)

Resources on Request Spec formatting:

- [RelishApp](https://relishapp.com/rspec/rspec-rails/v/4-0/docs/request-specs/request-spec)

Original Issue: https://github.com/ifmeorg/ifme/issues/1815

---

tyler-wel:
# Description

The scope of the request spec refactor expanded by just a tad. 
- Refactored the original comment controller spec into comments request spec
- Expanded on the test cases to cover more of the controller logic
- Added failure to the controller where deletion has actually failed
  - Originally, if a comment failed to delete, the frontend had no idea and continued as normal

## More Details
One bigger change was made to the route. To me personally, I found it odd to have to do `comment_index_path` to get the path we wanted, especially since this route doesn't _actually have_ an index action. Because of this, I changed it to `comments`. This also is inline with most of the other routes

- comment_controller -> comments_controller
- paths are now called as `comments_path` and `delete_comments_path`
- changed all references of the existing paths to use the new pluralized form
- changed the action from destroy to delete

The reason this change was originally prompted was due to the following error while creating the request spec
```
ActionController::UrlGenerationError:
  No route matches {:action=>"destroy", :controller=>"comment"}, missing required keys: [:id]
```

## Corresponding Issue

fixes #1823 

---


config/routes.rb
```diff
@@ -14,7 +14,7 @@
     end
   end
 
-  resources :comment, only: %i[create destroy] do
+  resources :comments, only: %i[create delete] do
```
julianguyen, line 17: Yesss 🙏

app/controllers/comments_controller.rb
```diff
@@ -44,16 +45,21 @@ def handle_create(comment_params)
     comment.id
   end
 
+  # rubocop:disable Metrics/MethodLength
```
julianguyen, line 48: It makes sense to add this and this is better than making unnecessary helpers for the sake of preventing Rubocop from complaining. I wonder long term if we should just disable "Metrics/MethodLength" in our general config. I don't see this as fundamental. I think PR reviews should help identify whether to reduce the size of a method.

It looks like there are some other places where `# rubocop:disable Metrics/MethodLength` is used (`# rubocop:disable Layout/LineLength` too). 

This can be tackled in another PR :) Let me know what you think!

julianguyen (approved):
Excellent work, thanks for going above and beyond to improve error handling and routing 🙌

app/controllers/comments_controller.rb
tyler-wel: Yaaa, I agree with this. I was staring at this method like.... how... how can I reduce this even more??? lol

10 lines just seems waaaay too short, and you'll end up having a ton of little unnecessary "helper" methods. 

I think adding a PR to adjust the rubocop rules would be good! Maybe either a new max like...35? 40? Or just remove it all together 🤔

julianguyen: Good question. Yeah, maybe we could increase the max for now. Do you wanna take that on in another PR? :)

julianguyen (approved):

app/controllers/comments_controller.rb
tyler-wel: Ya sure :D

julianguyen: Looks like we can remove these comments with your rubocop config changes merged!

spec/requests/comments_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+      let(:user) { create(:user) }
+
+      # Shared examples reduce repeated code
+      shared_examples_for 'a comment is not created' do
+
+        it { is_expected.to eq 400 }
+
+        it 'returns an empty comment' do
+          subject
+          expect(JSON.parse(response.body)).to eq({})
```
bennpham, line 24:
```suggestion
          expect(JSON.parse(response.body)).to be_empty
```

Just a nitpick otherwise this is fine too.

bennpham (approved):
This looks great. Good job with fixing the routing 👍 👍 🐶 🐶 🐕

spec/requests/comments_spec.rb
tyler-wel: Ahhh good point 👍 if I get back to my PC before it's merged I'll fix it, it not maybe if I can do another rspec refactor i'll do it then... Or maybe we can do a like, once over refactor once all of the smaller issues are merged 🤔

PR #1862, opened 2020-10-03
feat: #1656 refactor ChartControl using Hooks

udayanshevade:
# Description

- Use `useState` for tracking selected chart data `type`
- Add tests, reuse mock data definitions *
- Minor changes: style, flow updates

## More Details

\* _Note_: test-related updates weren't needed for this PR. The existing test checked for a successful render and passed without issue. But I thought to add further expectations for:
- what initial data gets passed down to the Chart component,
- whether the rendered chart data is correspondingly updated when
   a specified ChartControlButton is clicked.

If anything should be in this PR (or shouldn't), please do LMK and I'll make the necessary changes.

Thank you! ☕ 

## Corresponding Issue

#1656 

---

### DIFF
--- a/client/app/components/Chart/ChartControl.jsx
+++ b/client/app/components/Chart/ChartControl.jsx
@@ -35,6 +35,7 @@ export const ChartControl = ({
   const onSelectType = (value: string) => () => {
     setType(value);
   };
+
   return (
     <div className={css.chartControl} role="presentation">
       <div>
--- a/client/app/components/Chart/ChartControl.jsx
+++ b/client/app/components/Chart/ChartControl.jsx
@@ -1,15 +1,10 @@
 // @flow
-import React from 'react';
+import React, { useState } from 'react';
 import { Chart } from './index';
 import globalCss from '../../styles/_global.scss';
 import css from './ChartControl.scss';
 
-type chartControlState = {
-  type: string,
-  data: any,
-};
-
-type chartControlProp = {
+type chartControlProps = {
   types: string[],
   initialParams: {
     type: string,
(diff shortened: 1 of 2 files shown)


client/app/components/Chart/ChartControl.jsx
```diff
@@ -31,45 +26,33 @@ const ChartControlButton = ({ type, onClick }: ChartControlButtonProps) => (
...
+            onClick={onSelectType(value)}
+          />
+        ))}
       </div>
-    );
-  }
-}
+      <Chart
+        ytitle={`${type}`}
+        xtitle="Date"
```
julianguyen, line 52: Ooh I realized we should be using an i18n key here instead of hardcoding to an English value.

You can use `I18n` from `client/app/libs/i18n/index.js`

`I18n.t('common.date')`

client/app/components/Chart/__tests__/ChartControl.spec.jsx
```diff
@@ -7,28 +7,79 @@ import React from 'react';
...
       );
     }).not.toThrow();
 
     expect(wrapper).not.toBeNull();
-    expect(wrapper.find('ChartControlButton').length).toEqual(1);
+    expect(wrapper.find('ChartControlButton').length).toEqual(types.length);
     expect(wrapper.find('Chart').length).toEqual(1);
   });
+
+  it('passes the expected data for the selected type', () => {
```
julianguyen, line 59: Thanks for adding this test :)

client/app/components/Chart/__tests__/ChartControl.spec.jsx
```diff
@@ -7,28 +7,79 @@ import React from 'react';
...
+    const wrapper = shallow(
```
julianguyen, line 60: So we actually have another [issue](https://github.com/ifmeorg/ifme/issues/1789) to use React Testing Library instead of Enzyme because we are moving away from it. Would you be interested in tackling this too? :)

udayanshevade:
Ah, I was wondering whether to refactor to use 'react-testing-library', but figured that could be done in a later PR. Looks like the issue for that exists already here #1789 :)

client/app/components/Chart/__tests__/ChartControl.spec.jsx
```diff
@@ -7,28 +7,79 @@ import React from 'react';
...
+      />,
+    );
+
+    // query child chart component to get current props
+    let chart = wrapper.find('Chart');
+    // test initial data passed down to the chart
+    expect(chart.prop('data')).toBe(data[MOMENTS_TYPE]);
+
+    // simulate click on ChartControlButton for a specified type
+    const typeToTest = types[Math.floor(Math.random() * types.length)];
```
julianguyen, line 73: Why not just specify a type instead of randomizing it?

julianguyen (commented):
Solid work! Thanks for testing the interaction too ^^

client/app/components/Chart/ChartControl.jsx
udayanshevade: Definitely! Let me include that in an update.

client/app/components/Chart/__tests__/ChartControl.spec.jsx
udayanshevade: Yes absolutely! I was just wondering about it when adding the new test. I can take that one on.

julianguyen: Yes feel free to take it on!

julianguyen: You can do it in another PR if you want!

client/app/components/Chart/__tests__/ChartControl.spec.jsx
udayanshevade: I...lack a concrete rationale for it :smile: I'd even specified `'Categories'` initially, but then got wondering if the test might be more generic if it didn't matter which type was selected?

Would it be better to specify a fixed value? I can include that in my updates as well 👍 

PS: I made a similar change for the expected number of chart control buttons too, tying it to `types.length`. I can also change that back to a constant number too if it makes more sense to do so.

client/app/components/Chart/__tests__/ChartControl.spec.jsx
udayanshevade: Awesome! Will do

client/app/components/Chart/__tests__/ChartControl.spec.jsx
julianguyen: I think it's fine to specify a fixed value, that way the assertion is more "reliable".

udayanshevade: Sounds good! I'll change these back

udayanshevade: On a related note, would it be better to keep the mock data specific to each test too, as opposed to the current update? I can revert the first test back to its original state and follow a similar pattern for the second test?

julianguyen (approved):
Excellent work! Thanks for working on this :)

PR #1870, opened 2020-10-03
Refactor Search controller spec to Search request spec

Ashwithabg:
#1838

### DIFF
--- a/spec/requests/search_spec.rb
+++ b/spec/requests/search_spec.rb
@@ -1,49 +1,56 @@
 # frozen_string_literal: true
-describe SearchController, type: :controller do
+
+describe 'Search', type: :request do
   let(:user) { create(:user) }
-  let(:user2) {  create(:user, email: 'bar@email.com') }
-  let(:user3) {  create(:user, email: 'foo@email.com', banned: true) }
+  let!(:user2) do
+    create(:user, email: 'bar@email.com',
+                  name: 'user 2 name')
+  end
+  let!(:user3) do
+    create(:user, email: 'foo@email.com',
+                  banned: true,
+                  name: 'user 3 name')
+  end
 
   describe '#index' do
     context 'when user is logged in' do
-      include_context :logged_in_user
+      before { sign_in user }
 
       context 'when have passed email' do
         it 'strip passed string' do
-          expect_any_instance_of(String).to receive(:strip)
-          get :index, params: { search: { email: '  hi@email.com ' } }
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
           expect(response).to render_template(:index)
         end
 
         it 'filters by non-banned user email' do
-          get :index, params: { search: { email: 'bar@email.com' } }
-          expect(assigns(:matching_users)).to include(user2)
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
           expect(response).to render_template(:index)
+          expect(response.body).to include(user2.name)
         end
 
         it 'does not filter banned user email' do
-          get :index, params: { search: { email: 'foo@email.com' } }
-          expect(assigns(:matching_users)).not_to include(user3)
+          get search_index_path, params: { search: { email: 'foo@email.com' } }
+          expect(response.body).not_to include(user3.name)
           expect(response).to render_template(:index)
         end
 
         it 'keeps a reference to the email queried' do
-          get :index, params: { search: { email: 'bar@email.com' } }
-          expect(assigns(:email_query)).to include('bar@email.com')
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
+          expect(response.body).to include(CGI.escape('bar@email.com'))
           expect(response).to render_template(:index)
         end
       end
 
       context 'when have no passed email' do
         it 'sets the correct instance variables' do
-          get :index, params: { search: {} }
+          get search_index_path, params: { search: {} }
           expect(response).to redirect_to('/allies')
         end
       end
     end
 
     context 'when user is not logged in' do
-      before { get :index, params: { search: {} } }
+      before { get search_index_path, params: { search: {} } }
 
       it_behaves_like :with_no_logged_in_user
     end
@@ -51,14 +58,13 @@ describe SearchController, type: :controller do
 
   describe '#posts' do
     context 'when user is logged in' do
-      include_context :logged_in_user
+      before { sign_in user }
       let(:word) { 'passed-word' }
 
       context 'when there is a searched name' do
         before do
           get(
-            :posts,
-            format: :html,
+            posts_search_index_path,
             params: { search: { name: word, data_type: data_type } }
           )
         end
@@ -100,8 +106,7 @@ describe SearchController, type: :controller do
       context 'when there is not a searched name' do
         before do
           get(
-            :posts,
-            format: :html,
+            posts_search_index_path,
             params: { search: { name: '', data_type: data_type } }
           )
         end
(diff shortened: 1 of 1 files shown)


spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
 # frozen_string_literal: true
-describe SearchController, type: :controller do
+
+describe 'Search', type: :request do
   let(:user) { create(:user) }
-  let(:user2) {  create(:user, email: 'bar@email.com') }
-  let(:user3) {  create(:user, email: 'foo@email.com', banned: true) }
+  let!(:user2) do
```
julianguyen, line 5: I don't think you need the `!`.

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
+    create(:user, email: 'bar@email.com',
+                  name: 'user 2 name')
+  end
+  let!(:user3) do
```
julianguyen, line 9: Same here.

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
+          get search_index_path, params: { search: { email: 'foo@email.com' } }
+          expect(response.body).not_to include(user3.name)
           expect(response).to render_template(:index)
         end
 
         it 'keeps a reference to the email queried' do
-          get :index, params: { search: { email: 'bar@email.com' } }
-          expect(assigns(:email_query)).to include('bar@email.com')
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
+          expect(response.body).to include(CGI.escape('bar@email.com'))
```
julianguyen, line 39: Why is `CGI.escape()` necessary?

julianguyen (commented):
Awesome work, thanks for taking this on! I just have a couple of questions.

bennpham (changes requested):
Nice job with picking up this issue 👍. I have a handful of suggestions that might make the tests more intuitive to read and follow.

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
     context 'when user is logged in' do
-      include_context :logged_in_user
+      before { sign_in user }
 
       context 'when have passed email' do
         it 'strip passed string' do
-          expect_any_instance_of(String).to receive(:strip)
-          get :index, params: { search: { email: '  hi@email.com ' } }
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
           expect(response).to render_template(:index)
```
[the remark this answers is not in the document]
bennpham: Forgot to mention, it is [recommended to avoid the rails-controller-testing gem](http://rspec.info/blog/2016/07/rspec-3-5-has-been-released/#comment-3869319771) so if you assert that going to this URL is successful, that should be sufficient 👍 

`render_template` is only supported by the gem I mentioned above.

spec/requests/search_spec.rb
Ashwithabg: Sure will do this change. Thanks for the feedback.

spec/requests/search_spec.rb
Ashwithabg: In the `response.body` the html response have email encoded format `bar%40email.com` instead of `bar@email.com`. Do you know another to do this assertion?

spec/requests/search_spec.rb
Ashwithabg: Sure will do this change. Thanks for the feedback.

spec/requests/search_spec.rb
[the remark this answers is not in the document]
Ashwithabg: got it. Will do this change.

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
+          expect(response.body).not_to include(user3.name)
           expect(response).to render_template(:index)
         end
 
         it 'keeps a reference to the email queried' do
-          get :index, params: { search: { email: 'bar@email.com' } }
-          expect(assigns(:email_query)).to include('bar@email.com')
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
+          expect(response.body).to include(CGI.escape('bar@email.com'))
           expect(response).to render_template(:index)
```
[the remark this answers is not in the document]
Ashwithabg: sure. will do this change

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
+          expect(response.body).to include(user2.name)
         end
 
         it 'does not filter banned user email' do
-          get :index, params: { search: { email: 'foo@email.com' } }
-          expect(assigns(:matching_users)).not_to include(user3)
+          get search_index_path, params: { search: { email: 'foo@email.com' } }
+          expect(response.body).not_to include(user3.name)
           expect(response).to render_template(:index)
```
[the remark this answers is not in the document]
Ashwithabg: sure. will do this change.

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
+                  name: 'user 3 name')
+  end
 
   describe '#index' do
     context 'when user is logged in' do
-      include_context :logged_in_user
+      before { sign_in user }
 
       context 'when have passed email' do
         it 'strip passed string' do
```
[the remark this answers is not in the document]
Ashwithabg: sure will do this change. Thank you

spec/requests/search_spec.rb
```diff
@@ -1,64 +1,70 @@
...
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
           expect(response).to render_template(:index)
         end
 
         it 'filters by non-banned user email' do
-          get :index, params: { search: { email: 'bar@email.com' } }
-          expect(assigns(:matching_users)).to include(user2)
+          get search_index_path, params: { search: { email: 'bar@email.com' } }
           expect(response).to render_template(:index)
+          expect(response.body).to include(user2.name)
```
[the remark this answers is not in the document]
Ashwithabg: got it. Sure will do this change. Thank you

spec/requests/search_spec.rb
bennpham: I tried just `expect(response.body).to include('bar@email.com')` and it works for me. 

When you type in the search box and it says `User bar@email.com not found`. You're just asserting the `response.body` that the following text is there which is in a string format. 

Here's a screenshot as reference for what we would be trying to assert.

image

julianguyen: I think if you do `JSON.parse(response.body)` it will use the actual `@` character. Let's use this instead since we do this in other similar tests.

bennpham: @julianguyen Interesting. `JSON.parse` actually fails for me while `response.body` itself works. It seems like using `JSON.parse(response.body)` throws an exception for `JSON::ParserError` `784: unexpected token at '<!DOCTYPE html>`.

`CGI.escape` wasn't needed when I tried testing it.

julianguyen: Thanks for verifying! :) +1 to removing. `CGI.escape` then!

bennpham (approved):
Everything else looks good. Setting my approval ahead before I take off, but pretty much if `CGI.escape` is remove then everything looks good to me afterwards. 👍 👍

spec/requests/search_spec.rb
Ashwithabg: yeah. It worked without `CGI.escape` 
sorry last time I tried I got `bar%40email.com` that's why I had added this logic.

julianguyen (approved):
Great work again! Feel free to merge whenever!

Issue #1844, opened 2020-09-21
Google Sign Up is not working


erickaorf:
![Screenshot 2020-09-21 185622](https://user-images.githubusercontent.com/58405291/93832800-6de23500-fc3c-11ea-9677-d193e1b67163.png)
<!--[
 I tried signing up via google, it did not work. I got the above screen.

julianguyen:
Thanks for reporting this! Is this happening on the live if-me.org site for you or in your development environment?

adedayomoshood:
This happens on the live if-me.org site. Googling the issue, the App needs to be reverified to use Google oauth login

julianguyen:
Oh no, I'll look into this. I wasn't able to reproduce this.

julianguyen:
I think there was an outage with some Google Cloud services this week.

I got this message in the dashboard:

> We are currently experiencing service disruptions and are unable to provide quota information at this time.

julianguyen:
Thanks, everyone for verifying! I just submitted the process to verify our app. In the past, we didn't have to do this but it looks like we have to now.  I'll keep everyone posted!

julianguyen:
We've been verified! Could folks test whether they can sign in/up now? :)

andrew-schutt:
Still had the same error for me 😅 Perhaps it takes a little while to take effect?

julianguyen:
I hope so lol. I double-checked the [Google OAuth scopes](https://github.com/ifmeorg/ifme/blob/695f7af517c74bbab982453f75ebe8b0790c0747/config/initializers/devise.rb#L318) we set and they seem correct.

They're appearing in GCP.
![image](https://user-images.githubusercontent.com/3010728/95031238-bef51e80-0669-11eb-9ae4-f433f86dd954.png)

julianguyen:
Turns out I need to go through another verification level to be able to use sensitive scopes (for the Calendar API). So I had to submit a video demonstrating how it's used in our app and do a short write-up on why we can't narrow down our scopes.

Our app was only verified on a "brand level" lol.

Here's [more info on verification levels and scopes for Google OAuth2](https://developers.google.com/identity/protocols/oauth2/scopes) if anyone is interested 💀 🤣

andrew-schutt:
Come on Google!

![flipceral](https://user-images.githubusercontent.com/413224/95114964-590ca380-070a-11eb-980f-4e82c44bba80.gif)

julianguyen:
[...]

PR #1873, opened 2020-10-04
Create OAuthButton component for Google and FB auth branded buttons

julianguyen:
# Description

Google authentication is not working in production for new users because we need to verify our application. I've submitted a request and updated our GCP account accordingly. The last step is to replace the "Sign in with Google" and "Sign up with Google" buttons with the [official styles](https://developers.google.com/identity/branding-guidelines#top_of_page). Sadly, I can't just download the SVGs they offer and stick it in our app. We support internationalization 🌐 

After that, our app will be verified 😅 

## More Details

I created a new React component called `OAuthButton` which will render the auth buttons for both Google and Facebook (might as well lol). I've added a story for this new component in Storybook 📖 

## Corresponding Issue

#1844

# Screenshots

## Sign in

image

image

## Sign up

image

image

## Invitation

image

image

---


client/app/components/OAuthButton/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+  signIn?: boolean,
+  type: 'facebook' | 'google',
+  action: string,
+  token: string,
+};
+
+export const OAuthButton = ({
+  signIn, type, action, token,
+}: Props) => {
+  const value = I18n.t(`devise.shared.sign_${signIn ? 'in' : 'up'}_${type}`);
```
andrew-schutt, line 18: Nitpick: Do you think that a more meaningful variable name would be useful here? I see that `value` is tied to the `button` attribute, but maybe like `oAuthButtonText` (not the best... 😄 )

andrew-schutt (approved):
Nice work! I really like how the `OAuthButton` abstraction encompasses both Facebook/Google and Sign In/Sign Up

PR #1874, opened 2020-10-04
feat: #1789 migrate chart tests to use react testing library

udayanshevade:
# Description

Migrate Chart and ChartControl tests to use React Testing Library instead of Enzyme

## More Details

In addition to updating the tests, this PR also adds a couple of other optional dev dependencies:

- [**@testing-library/user-event**](https://github.com/testing-library/user-event#the-problem):
Although the `fireEvent` that comes with RTL is sufficient for the purpose of this PR, I saw that this dependency came recommended, so I thought to include it here. If its inclusion here is premature, I can definitely roll it back.

- [**canvas**](https://www.npmjs.com/package/canvas):
Although the tests still pass without this one, it was displaying the error below (in the Screenshots section).

## Corresponding Issue

#1789 

# Screenshots

Screen Shot 2020-10-04 at 12 09 36 AM

As always, if anything looks out of place (or missing), please LMK and I'll make the necessary changes.

Thank you! ☕ 

---


udayanshevade:
I also noticed a really minor issue that the second hex string in these [colorSchemes](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Chart/index.jsx#L16) appears incorrect. It makes areas in [some charts](http://design.if-me.org/?path=/story/components-chart--default) appear grey. Rather than guessing a good value, I decided not to include that yet. I'm not as good with color and I kind of liked the grey.

julianguyen (approved):
Awesome work! Thank you for migrating this :D

PR #1882, opened 2020-10-06
Refactor profile_controller spec into profile request spec

DRBragg:
# Description

This PR refactors the tests in `profile_controller_spec.rb` to use RSpec request tests instead of controller tests.

## More Details

Confirmed coverage of ProfileController remained at 100% after refactor.

## Corresponding Issue

Closes #1834 

---

Reviewing this pull request? Check out our [Code Review Practices](https://github.com/ifmeorg/ifme/wiki/Code-Review-Practices) guide if you haven't already!

PR #1883, opened 2020-10-06
Refactor pages_controller tests into request tests

### LINKED ISSUE #1833: Refactor spec/controllers/pages_controller_spec.rb to spec/request/page_spec.rb
# Description

Controller Specs had been obsolete since Rails 5. It is best practice to create request spec instead of controller spec as recommended by the Rspec team.

Recommended Reading:

- [Stackoverflow Discussion in regards to Controller Specs vs Request Specs](https://stackoverflow.com/questions/40851705/controller-specs-vs-request-specs)
- [Medium article discussing request and controller specs](https://medium.com/just-tech/rspec-controller-or-request-specs-d93ef563ef11)

Resources on Request Spec formatting:

- [RelishApp](https://relishapp.com/rspec/rspec-rails/v/4-0/docs/request-specs/request-spec)

Original Issue: https://github.com/ifmeorg/ifme/issues/1815

---

DRBragg:
# Description

This PR refactors the tests in pages_controller_spec.rb to use RSpec request tests instead of controller tests.

## More Details

Had some interesting failures with in the `toggle_locale` tests.  For some reason when moving from a controller test to a request test the locale update was triggering the password validation.  Because of this, `current_user.update(locale: params[:locale])` was returning false, due to the password error being added. Removing `user.update!(locale: "en")` from each test and simply setting the locale to 'en' when the user was created bypassed this issue. Coverage for the `PagesController` is the same as before but futhur investigation into the cause of this issue may be needed.

## Corresponding Issue

Closes #1833 

---


spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  describe "#toggle_locale" do
+    context "When user is signed in" do
+      let(:user) { build(:user, locale: "en") }
+      before { sign_in user }
+
+      it "has a 200 status when the locale changes" do
+        post toggle_locale_path, params: {locale: "es"}
+        expect(user.locale).to eq("es")
+        expect(response.status).to eq(200)
```
bennpham, line 119:
```suggestion
        expect(response).to have_http_status(200)
```

Nitpick, but I thought this looks cleaner.

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+
+      it "has a 400 status when the locale is the same" do
+        post toggle_locale_path, params: {locale: "en"}
+        expect(user.locale).to eq("en")
+        expect(response.status).to eq(400)
```
bennpham, line 125:
```suggestion
        expect(response).to have_http_status(400)
```
Nitpick

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+    end
+
+    context "When not signed in" do
+      it "has a 200 status" do
+        post toggle_locale_path, params: {locale: "es"}
+        expect(response.status).to eq(200)
```
bennpham, line 132:
```suggestion
        expect(response).to have_http_status(200)
```
Nitpick

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+    end
+  end
+
+  describe "#resources" do
+    describe "when sending filter params" do
+      it "filters the aforementioned resources" do
+        get resources_path, params: {filter: %w[ADD english]}
+
+        expect(assigns(:keywords)).to match_array(%w[ADD English])
```
bennpham, line 142: I'd make sure the response.body contain ADD English or something of that sort since `assigns` is only for `rails-controller-test` gem which directly tests the variable in the controller itself instead of what is on the page itself.

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+
+      it "filters only existing resources" do
+        get resources_path, params: {filter: %w[ADD someUnexistentTag]}
+
+        expect(assigns(:keywords)).to match_array(["ADD"])
```
bennpham, line 148: Same with above in regards to `assigns` method being dependent on `rails-controller-test` gem.

bennpham (commented):
Good job on attempting this refactor into request tests. 👍 I left a handful of feedback in regards to `rails-controller-test` gem on what might need to be changed from controller specs and a few question on my confusion on test setup from the previous implementation. Let me know if you have any questions if I'm available to help out or give some finer examples to the feedback.

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+    end
+
+    context "logged in" do
+      before { sign_in user }
+
+      it "has no stories" do
+        list = double
+        expect(Kaminari).to receive(:paginate_array).and_return(list)
+        expect(list).to receive(:page)
+        get pages_home_path
```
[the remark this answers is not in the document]
tyler-wel: I'm also somewhat against these kinds of tests, as I follow the idea that tests should test for inputs/outputs and not the guts of the controller. The exception to this is 3rd party calls outside of the API, which should be mocked out with `allow` as benn pointed out.

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+
+      it "has stories" do
+        create(:strategy, user_id: user.id)
+        categories = create_list(:category, 2, user_id: user.id)
+        moods = create_list(:mood, 2, user_id: user.id)
```
tyler-wel, line 25: I'd also like to add that these should be moved outside of the "it" block and into `let()` assignments

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+    it "respond to request" do
+      get pages_home_path
+      expect(response).to be_successful
+    end
+
+    context "logged in" do
+      before { sign_in user }
+
+      it "has no stories" do
+        list = double
```
tyler-wel, line 16: should be moved to a `let()` block

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      get partners_path
+      expect(response).to be_successful
+    end
+
+    it "read external JSON file" do
+      data = []
+      file = File.read("doc/pages/partners.json")
+      expect(JSON).to receive(:parse).with(file).and_return(data)
+      expect(data).to receive(:sort_by!)
+      get partners_path
```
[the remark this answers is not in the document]
tyler-wel: The `expect().to receive()` are expectations for the future. It does the same as `allow().to receive()` but will fail if it is never called. This is not entirely intuitive, so there is also the `expect().to have_received()`

tyler-wel (changes requested):
A lot of the code is a bit inconsistent, so it'd be good to *actually* refactor the code to match and fix old syntax

spec/requests/pages_spec.rb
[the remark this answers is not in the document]
bennpham: Yeah, I was wondering since most patterns I see it would be like the following:

```ruby
# Arrange
allow().to receive()

# Act
get some_path

# Assert
expect().to have_received
```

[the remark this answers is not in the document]
tyler-wel: Ya, that's the pattern I've generally followed to, because most people see the other one and get confused. Not very intuitive, but it usually works the same way lol

tyler-wel:
@DRBragg Will you be continuing development on this?

DRBragg:
@tyler-wel I'll work on this one once I wrap up the refractors in #1882

spec/requests/pages_spec.rb
DRBragg: Where should those `let( )` assignments happen? The `user` variable won't be available in 'describe' or 'context' blocks

DRBragg: We could use `let!` but that differs from what @bennpham requested in their review from line 46

bennpham: I don't see it used anywhere else to put it in `let()` assignments looking at the test itself. I think we'd be over abstracting it at that point. If we had a context for maybe `when user have stories` and multiple `it` blocks, then I can see using `let!` here unless I'm missing something.

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+
+      it "has stories" do
+        create(:strategy, user_id: user.id)
+        categories = create_list(:category, 2, user_id: user.id)
+        moods = create_list(:mood, 2, user_id: user.id)
+        get pages_home_path
+        expect(assigns(:moment)).to be_a_new(Moment)
+        expect(assigns(:categories)).to eq(categories.reverse)
+        expect(assigns(:moods)).to eq(moods.reverse)
```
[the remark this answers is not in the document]
DRBragg: There might be a better place for some of this test now. Just looking at the response it seems the the categories and moods don't come over. Since we're moving away from controller testing perhaps we should move the parts of this test that covers the categories and moods and only test that the `StoryContainer` and/or the "Test Strategy" are in the response body

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+          contain_exactly(:link, :link_name, :author)
+        )
+        blurbs_file = File.read("doc/pages/blurbs.json")
+        expect(assigns(:blurbs)).to eq(JSON.parse(blurbs_file))
+      end
+    end
+  end
+
+  describe "#home_data" do
+    let!(:moment) { create(:moment, user: user) }
```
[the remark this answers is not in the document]
DRBragg: I can move it there but then I would also have to move both `before { }` blocks into the test itself. Totally fine by me but that would be inconsistent with the rest of the tests. TBH, I'm not a fan of having things in `before { }` bocks when there's only one `it`. I understand the point when there are multiple tests and keeping them DRY but with only one test it seems/feels forced. 

I'll go with whatever path you think is best.

bennpham:
@DRBragg Are you still working on this?

DRBragg:
Unfortunately I am not. I'm sorry for wasting your time... 😔

bennpham:
@DRBragg Oh no worries! My time isn't wasted at all.

bennpham:
I'm going to close this in this case.

PR #1891, opened 2020-10-09
feat: #1792 Migrate 'components/Inputs/' tests to React Testing Library

udayanshevade:
# Description

Updates the tests in ‘components/Input/’ to use React Testing Library instead of Enzyme.

## More Details

(1) I added comments where I ran into a couple of issues using `userEvent` and jest-dom matchers. I updated the dependencies to see if it would help (Narrator: it didn’t). But that amounted to just a minor version bump in '@testing-library/react' so I left it in.

(2) I also added some coverage for the Textarea component. I noticed a TODO there about stubbing the Pell editor methods. I can always revert the newly added tests, or make the necessary changes if it should be done a better way.

(3) Also, I was wondering whether to add more coverage to a couple of the Input components. I decided just to migrate the existing ones for starters, but I can definitely add more if time permits.

(4) Lastly, I updated minor details in a couple components themselves. They still work the same (including screenshots below).

Of course, if anything looks out of place (or missing), please LMK and I'll make the necessary changes.

Thanks! :coffee:

## Corresponding Issue

#1792

# Screenshots

Switch (updated):
Screen Shot 2020-10-08 at 7 02 46 PM

Tag (updated):
Screen Shot 2020-10-08 at 11 28 12 PM

---

### DIFF
--- a/client/app/components/Input/InputCheckboxGroup.jsx
+++ b/client/app/components/Input/InputCheckboxGroup.jsx
@@ -6,7 +6,7 @@ import type { Checkbox } from './utils';
 export type Props = {
   checkboxes: Checkbox[],
   required?: boolean, // At least one checkbox must be checked
-  hasError?: Function,
+  hasError?: (errorPresent: boolean) => void,
 };
 
 export function InputCheckboxGroup({
--- a/client/app/components/Input/InputSwitch.jsx
+++ b/client/app/components/Input/InputSwitch.jsx
@@ -37,7 +37,7 @@ export function InputSwitch({
     setKey(Utils.randomString());
   };
 
-  const onKeyPress = (e: SyntheticKeyboardEvent<HTMLInputElement>) => {
+  const onKeyDown = (e: SyntheticKeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Enter') {
       toggleChecked();
     }
@@ -48,7 +48,6 @@ export function InputSwitch({
       <InputCheckbox
         id={id}
         key={key}
-        type="checkbox"
         name={name}
         label={label}
         value={value}
@@ -70,7 +69,7 @@ export function InputSwitch({
       id={`${id}_switch`}
       className={`switchToggle ${css.switchToggle}`}
       onClick={toggleChecked}
-      onKeyPress={onKeyPress}
+      onKeyDown={onKeyDown}
       role="switch"
       aria-checked={checked}
       tabIndex={0}
--- a/client/app/components/Input/__tests__/Input.spec.jsx
+++ b/client/app/components/Input/__tests__/Input.spec.jsx
@@ -1,27 +1,40 @@
 // @flow
-import { mount } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { InputMocks } from 'mocks/InputMocks';
+import css from '../Input.scss';
+
+/**
+ * TODO: Follow up on an issue when using the matcher `.toBeVisible()` on the inputs directly.
+ * The components behave correctly, but the computed styles don't seem to correspond.
+ * Even if the accordion is closed, and uses the class 'accordionClosed', the actual style returns
+ * a visible value for the 'display' property, as opposed to the expected value 'none'.
+ * This is also seen with the helper `.toHaveStyle()` (in @testing-library/jest-dom@5.11.4).
+ * For now, checking whether div["role='region'"] from Accordion has the class 'accordionClosed'.
+ */
 
 describe('Input', () => {
   describe('InputDefault', () => {
     describe('with accordion prop', () => {
       it('toggles correctly', () => {
-        const wrapper = mount(
+        const { label } = InputMocks.inputTextProps;
+        render(
           InputMocks.createInput(InputMocks.inputTextProps, {
             accordion: true,
           }),
         );
-        expect(
-          wrapper.find('.accordionContent').find('input').exists(),
-        ).toEqual(false);
-        wrapper.find('.accordion').simulate('click');
-        expect(
-          wrapper.find('.accordionContent').find('input').exists(),
-        ).toEqual(true);
-        wrapper.find('.accordion').simulate('click');
-        expect(
-          wrapper.find('.accordionContent').find('input').exists(),
-        ).toEqual(false);
+
+        const input = screen.getByRole('textbox');
+        const button = screen.getByRole('button', { name: new RegExp(label) });
+        const accordionContainer = screen.getByRole('region');
+
+        expect(input).toBeInTheDocument();
+
+        expect(accordionContainer).toHaveClass(css.accordionClose);
+        userEvent.click(button);
+        expect(accordionContainer).not.toHaveClass(css.accordionClose);
+        userEvent.click(button);
+        expect(accordionContainer).toHaveClass(css.accordionClose);
       });
     });
   });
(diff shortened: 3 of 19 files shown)


client/app/components/Input/InputSwitch.jsx
```diff
@@ -37,7 +37,7 @@ export function InputSwitch({
     setKey(Utils.randomString());
   };
 
-  const onKeyPress = (e: SyntheticKeyboardEvent<HTMLInputElement>) => {
```
udayanshevade, line 40: Small update: replaces [onKeyPress](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onkeypress) with `onKeyDown`, should still work without issue

client/app/components/Input/InputCheckboxGroup.jsx
```diff
@@ -6,7 +6,7 @@ import type { Checkbox } from './utils';
 export type Props = {
   checkboxes: Checkbox[],
   required?: boolean, // At least one checkbox must be checked
-  hasError?: Function,
+  hasError?: (errorPresent: boolean) => void,
```
udayanshevade, line 9: Specified the type here to reflect the fix in the mock prop passed to the component in the test.

client/app/components/Input/__tests__/InputCheckboxGroup.spec.jsx
```diff
@@ -10,7 +10,8 @@ const label = 'Some Label';
 const idTwo = 'some-other-id';
 const nameTwo = 'some-other-name';
 const labelTwo = 'Some Other Label';
-const someEvent = () => {
+const someEvent = (hasError) => {
```
udayanshevade, line 13: Minor fix: previously, `window.alert` would be invoked below even if there was no error. The assertion on line 55 helped catch this issue.

client/app/components/Input/__tests__/InputCheckbox.spec.jsx
```diff
@@ -18,9 +19,13 @@ describe('InputCheckbox', () => {
     jest.spyOn(window, 'alert');
   });
 
+  afterEach(() => {
+    jest.clearAllMocks();
```
julianguyen, line 23: Is this necessary?

client/app/components/Input/__tests__/InputCheckbox.spec.jsx
```diff
@@ -57,19 +59,20 @@ describe('InputCheckbox', () => {
...
-      );
-      wrapper
-        .find('input[type="checkbox"]')
-        .simulate('change', { currentTarget: { checked: false } });
+
+      // ensures the input for the uncheckedValue is hidden
+      expect(screen.queryByRole('input')).not.toBeInTheDocument();
+      // Since '@testing-library/react' does not get hidden inputs,
+      // it can be queried directly from the container for this test.
+      const hiddenInput = container.querySelector('input[type="hidden"]');
```
julianguyen, line 67: Good call to put a comment here :)

client/app/components/Input/__tests__/InputCheckboxGroup.spec.jsx
```diff
@@ -35,44 +36,42 @@ describe('InputCheckboxGroup', () => {
...
-        wrapper.find(`input[name="${nameTwo}"]`).prop('onChange')({
-          currentTarget: { checked: false },
-        });
-      });
+      const checkbox = screen.getByRole('checkbox', { name: label });
+      const otherCheckbox = screen.getByRole('checkbox', { name: labelTwo });
+
+      // toggle both checkboxes false
+      userEvent.click(checkbox);
+      userEvent.dblClick(otherCheckbox);
```
julianguyen, line 73: Can this just be a regular click?

client/app/components/Input/__tests__/InputDefault.spec.jsx
```diff
@@ -18,9 +19,13 @@ describe('InputDefault', () => {
     jest.spyOn(window, 'alert');
   });
 
+  afterEach(() => {
+    jest.clearAllMocks();
```
julianguyen, line 23: Is this necessary?

client/app/components/Input/__tests__/InputSwitch.spec.jsx
```diff
@@ -1,30 +1,50 @@
...
-      expect(wrapper.find(input).props().defaultChecked).toEqual(false);
+      render(component);
+      const inputSwitch = screen.getByRole('switch');
+
+      expect(screen.getByRole('checkbox')).not.toBeChecked();
+
+      /**
+       * TODO: Follow up on `userEvent.type(inputSwitch, '{enter}')` in v12.1.7.
+       * Temporarily including `fireEvent` from RTL, for which the switch must have focus first:
+       * https://github.com/testing-library/react-testing-library/issues/376#issuecomment-541242684
```
julianguyen, line 40: Thanks for documenting this!

client/app/components/Input/__tests__/InputTextarea.spec.jsx
```diff
@@ -0,0 +1,173 @@
+// @flow
+import { render, screen } from '@testing-library/react';
```
julianguyen, line 2: Thank you for writing tests for this 🏆

client/app/components/Input/__tests__/InputTextarea.spec.jsx
```diff
@@ -0,0 +1,173 @@
...
+        { title: 'Unordered List', expectedArgs: ['insertUnorderedList'] },
+      ];
+
+      buttons.forEach(({ title, expectedArgs }) => {
+        const button = screen.getByTitle(title);
+        userEvent.click(button);
+        expect(pell.exec).toHaveBeenCalledWith(...expectedArgs);
+      });
+
+      jest.restoreAllMocks();
```
julianguyen, line 170: Is this necessary?

julianguyen (commented):
Fantastic work!! :D Thanks so much for taking this on!

I have some questions but overall it looks good to me!

client/app/components/Input/__tests__/Input.spec.jsx
```diff
@@ -1,119 +1,143 @@
...
-          wrapper.find('.accordionContent').find('input').exists(),
-        ).toEqual(false);
+
+        const input = screen.getByRole('textbox');
+        const button = screen.getByRole('button', { name: new RegExp(label) });
+        const accordionContainer = screen.getByRole('region');
+
+        expect(input).toBeInTheDocument();
+
+        expect(accordionContainer).toHaveClass(css.accordionClose);
```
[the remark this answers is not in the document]
udayanshevade: Absolutely, I agree. Initially I tried going the same route as well.

I found that the previous tests had been checking for the presence of a hierarchy in the DOM, rather than a target element:

```
wrapper.find('.accordionContent').find('input').exists()
```

It turns out the input is always present in the DOM (but '.accordionContent' in the parent gets swapped for css.accordionClose):

Screen Shot 2020-10-09 at 10 50 59 AM

The visibility [gets toggled](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Input/Input.scss#L111) by the class `accordionClose` in the parent, rather than removing the input.

So with `queryByLabelText` I get:

Screen Shot 2020-10-09 at 10 30 21 AM

(Aside: I was wondering whether it would be a good idea to ask and revisit the implementation itself, but I felt it was probably better to stay in the scope of the issue.)

That's when I thought to use the more idiomatic `toBeVisible` matcher from 'jest-dom', which ought to detect whether an element is viewable or accessible to a screen reader.

But there I ran into [a possible external issue](https://github.com/testing-library/jest-dom/issues/209#issuecomment-697325787) with styles not updating correctly (leaving a [comment](https://github.com/ifmeorg/ifme/pull/1891/files#diff-80573b41c43ce7334f2de9e18d973fc4R8) there for more visibility), even though the functionality itself seems to work:

Screen Shot 2020-10-09 at 11 06 48 AM

So in the end I thought it might be better to revisit. I can definitely keep digging in that direction? Please LMK if I'm missing something!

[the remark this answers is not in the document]
julianguyen: Ooo it looks like you're right, ok I think what you have in this PR is fine. Then in another PR you can revisit the implementation. Thanks for clarifying!

julianguyen (approved):

client/app/components/Input/__tests__/InputCheckbox.spec.jsx
udayanshevade: Oh right! Let me remove those. The tests should work without them. At one point, I think I ran into an issue with a subsequent assertion passing even though it shouldn't (due to a previous call). But then I think I might have resolved it and forgotten to remove these. Should have noted it down

client/app/components/Input/__tests__/InputCheckboxGroup.spec.jsx
udayanshevade: Oh right I think I should maybe just leave `otherCheckbox` alone, since it has an initial value of `false` anyways. This double click essentially just toggles it on and off again 😅 I think I transcribed the former test too literally, where it simulated the `onChange` prop firing for a `false` value. Let me change it so that the test just assumes `otherCheckbox` is false to begin with.

client/app/components/Input/__tests__/InputCheckbox.spec.jsx
udayanshevade: So I went back and double-checked here just to make sure. I think maybe this reset in particular makes sense, as both blocks of tests use the same assertions.

It's possible for the functionality underlying the second block to break (somehow?) and its tests could still pass due to the previous calls from the first block of tests.

client/app/components/Input/__tests__/InputDefault.spec.jsx
udayanshevade: Similarly I think I added this reset as a way to future-proof any extended tests relying on `window.alert`, which might accidentally pass when they shouldn't because of a separate call.

Should that situation be handled when the time comes? I'm worried someone might roll with a false positive. But that can be caught in code review, so I can still definitely remove this.

Issue #1885, opened 2020-10-08
Update password creation to NIST 2017 standards


FranceJ:
# Description

The current password creation step has outdated composition requirements It requires a number, special character, upper case, and lower case character The [current NIST standards](https://pages.nist.gov/800-63-3/sp800-63b.html#app) specifically call out these kinds of restrictions as commonly resulting in less secure user passwords

The password field should composition rules should be updated to be in line with the new recommendations Specifically, the composition requirements should be removed and the minimum character requirement should remain.

The NIST document contains other recommendations for improving the password system, such as checking against known leaked sources, providing a password strength meter to help users choose strong passwords, and checking against common passwords and providing advice (not restrictions) based on the results. These could also be implemented but likely add too much scope to the simple change of removing the complexity restrictions, which is a simple change which adds value.

bennpham:
I'll probably like to take this on sometime around Friday :) . I remember posting about this on the **Suggested Password Strength Features** issue thread in regards to password strength meter and XKCD comic summary on password strengths: https://github.com/ifmeorg/ifme/issues/1696#issuecomment-678841098

I think the [Devise Pwned Password](https://github.com/michaelbanfield/devise-pwned_password) gem would be nice to utilize [haveIBeenPwned](https://haveibeenpwned.com/) breached passwords API to check for known leaked sources.

XKCD comic in regards to passwords: https://xkcd.com/936/

Here's another article (which is about 3 years old by now since its in 2017) that offer some nice hints on password policies: https://medium.com/@securitystreak/password-policy-template-2017-best-practices-cbdcbb6beb49

For this issue, I'd probably like to try to implement the following:

- the Devise Pwned Password gem to prevent known hacked passwords
- remove the extra password requirements
- set minimum password to around 8 characters (NIST recommendation although [this article](https://medium.com/@securitystreak/password-policy-template-2017-best-practices-cbdcbb6beb49) prefers at least 16 to enforce long passwords)

Not sure if I'd go beyond scope, but you can also change some Devise settings to lock out an account after 10 failed login attempts or show CAPTCHA after 3 failed attempt. Maybe a log of some sort for login attempts (Rails have paper trails gem or something for logs).

bennpham:
Taken from NIST itself 

> Clearly communicate information on how to create and change memorized secrets.
> Clearly communicate memorized secret requirements, as specified in Section 5.1.1.
[...]

Issue #1696, opened 2020-03-17
Suggested Password strength features checks  should be visible to the first time user who are signing up


developer22-university:
# Description

on looking at the user signup, i discovered it gets very difficult sometime to first-time users while signing up, i discovered this issue, when the user can't set up password correctly for the first time,

PRESENT SITUATION
![Screenshot (105)](https://user-images.githubusercontent.com/51835554/76902488-16948e00-68c2-11ea-9e0f-4dcef09a10f0.png)

 to make sign up simple and easy for mental health community , it is considered to implement suggested password features list visible to all first time users signing up, which verifies all the conditions and approves .This will assist in password strength verification and would be more feasible for everyone. 
WHAT WE WANT:
![Screenshot (106)](https://user-images.githubusercontent.com/51835554/76902572-522f5800-68c2-11ea-8fdb-bc0dab4993aa.png)

Please feel free to reach out for discussions.  Issue Waiting for approval.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

developer22-university:
@julianguyen what do you think, is it be useful for community. please send remarks so i can proceed, as i really like to implement this missing feature

julianguyen:
Hey, I think this is a great improvement. Right now those checks are being made on the server-side when the user submits the form. We can do the checks you proposed in the React component. You would want to modify `InputPassword`.

We can discuss the design further too.

developer22-university:
Sure

developer22-university:
julia  going through the code base i need  your help

julianguyen:
I think we should leverage the existing `Tooltip` component to list the information. You find documentation for it [here](http://design.if-me.org/?path=/story/tooltip--element-is-text). We'll want to customize it for this use case of course.

kajatiger:
Hi I also had problems signing up as a first time user. There were 2 problems:
1. After submitting my name, password and email to the form, the error message said that my password requirements didn't match (although they did). The password that I wanted to chose had 1 special character, more than one uppercase, more than one lowercase and more than one digit. Altogether a bit more than 8 characters, but definitely less than 128.
[...]

PR #1895, opened 2020-10-09
Update password creation to NIST 2017 standards

bennpham:
# Description

Removes password histories and extraneous password validations to meet the [current NIST standards](https://pages.nist.gov/800-63-3/sp800-63b.html#app)

## More Details

I decided not to implement Devise pwned to check for existing compromised password since this PR is big enough as it is which would make it more confusing to review. I rather spin that off into a separate issue to work on to avoid going out of scope. 

I also updated the locales for the **password_errors** section (hopefully I didn't make errors there or missed anything) to only tell users that they need a password between 8 to 128 characters since the other requirements are not up to date with current NIST standards.

The only thing that remains after this PR would be Devise 8 character minimum password limit. 

**Other options we can explore:** 
- suggesting new passwords as an advice (not restriction) as a separate issue using [Devise PWNED password](https://github.com/michaelbanfield/devise-pwned_password). 
- [password strength meter issue](https://github.com/ifmeorg/ifme/issues/1696) which we already have an issue for on our project
- temporarily locking out users or blacklisting IP addresses after multiple fail login attempt
- 2 factor authentication option
- Recaptcha after multiple failed login attempts

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1885

---


app/views/pages/home.html.erb
```diff
@@ -18,5 +18,3 @@
 <% else %>
   <%= render partial: 'signed_in_empty' %>
 <% end %>
-
-<%= render partial: 'update_password_modal' %>
```
bennpham, line 22: I think we can add this back if we introduce [Devise Pwned Passwords](https://github.com/michaelbanfield/devise-pwned_password) to suggest to the user to change their password if it was found in a data breach. 

Otherwise, I'm removing this for now since there isn't a case for this anymore.

julianguyen: Maybe we can comment this out with a TODO indicating this just so we don't forget!

julianguyen (approved):
Strong work! Your PR description was very informative too!

To respond to your comments about next steps:

- suggesting new passwords as an advice (not restriction) as a separate issue using Devise PWNED password.
  - I'm curious why this should be advice instead of a restriction!
- password strength meter issue which we already have an issue for on our project
  - Yes! It would be great if this could be done at the same time. I'll have more bandwidth for the rest of the month so I think I could potentially knock this out.
- temporarily locking out users or blacklisting IP addresses after multiple fail login attempt
  - What do you recommend here?
- 2 factor authentication option
  - This would be fantastic!
- Recaptcha after multiple failed login attempts
  - This would also be fantastic!

bennpham:
> suggesting new passwords as an advice (not restriction) as a separate issue using Devise PWNED password.
>> I'm curious why this should be advice instead of a restriction!

Rereading **5.1.1.2 Memorized Secret Verifiers** in [NIST](https://pages.nist.gov/800-63-3/sp800-63b.html#app), you're right. This should probably be a restriction based on `If the chosen secret is found in the list, the CSP or verifier SHALL advise the subscriber that they need to select a different secret, SHALL provide the reason for rejection, and SHALL require the subscriber to choose a different value.`

We should restrict it instead. I was thinking advise since we were deciding at work whether to restrict or advise for existing passwords. I'm trying to think of the best way to tell that to the average user without technical experience to avoid freaking them out.

> password strength meter issue which we already have an issue for on our project
>> Yes! It would be great if this could be done at the same time. I'll have more bandwidth for the rest of the month so I think I could potentially knock this out.

[Dropbox's Zxcvbn](https://github.com/dropbox/zxcvbn) plugin would come in handy here :) 

> temporarily locking out users or blacklisting IP addresses after multiple fail login attempt
>> What do you recommend here?

This one would probably depends more on threat model. Like for HealthTech or FinTech I'd have stricter limits due to a lot of money involve or health care data is very sensitive information and is highly protected by data laws. 

I think [NIST](https://pages.nist.gov/800-63-3/sp800-63b.html#app) **5.2.2 Rate Limiting (Throttling)** might have some advice. Not more than 100 failed attempts. 

I'm just guessing, but maybe RECAPTCHA after 3-5 failed attempts? Lock account out for an hour after 10 failed attempts? Just possible strategies, but I think for that one, it really depends on the threat model.

julianguyen:
> We should restrict it instead. I was thinking advise since we were deciding at work whether to restrict or advise for existing passwords. I'm trying to think of the best way to tell that to the average user without technical experience to avoid freaking them out.

I think we should restrict for new passwords Then for existing passwords, we should put up that update password modal and enforce changing passwords.

> Dropbox's Zxcvbn plugin would come in handy here :)

Ooh thanks I'll check it out!

> This one would probably depends more on threat model. Like for HealthTech or FinTech I'd have stricter limits due to a lot of money involve or health care data is very sensitive information and is highly protected by data laws.

Because there is sensitive mental health data, I think we should treat this as strict.

> I think NIST 5.2.2 Rate Limiting (Throttling) might have some advice. Not more than 100 failed attempts.

> I'm just guessing, but maybe RECAPTCHA after 3-5 failed attempts? Lock account out for an hour after 10 failed attempts? Just possible strategies, but I think for that one, it really depends on the threat model.

Ah ok that makes sense!

bennpham:
@julianguyen I'll check on this once I get back home, but anything else before merging except making a new issue to implement breached password checking?

julianguyen:
Looks like you're good to merge!

PR #1897, opened 2020-10-10
Refactor InputTextArea to be function component

FranceJ:
# Description
Refactors InputTextArea from a class component to a function component

## Corresponding Issue
[#1656](https://github.com/ifmeorg/ifme/issues/1656)

# Screenshots

image

---

### DIFF
--- a/client/app/components/Input/InputTextarea.jsx
+++ b/client/app/components/Input/InputTextarea.jsx
@@ -66,13 +66,13 @@ export type Props = {
 export function InputTextarea({
   id,
   name,
-  value,
+  value: propsValue,
   required,
   hasError,
   myRef,
   dark,
 }: Props) {
-  const [currentValue, setValue] = useState<string>(value || '');
+  const [value, setValue] = useState<string>(propsValue || '');
   const editorRef = useRef(null);
   const editor = useRef(null);
 
@@ -82,7 +82,7 @@ export function InputTextarea({
 
   const onBlur = () => {
     if (required && hasError) {
-      hasError(!currentValue || currentValue === '<p><br></p>');
+      hasError(!value || value === '<p><br></p>');
     }
   };
 
@@ -103,7 +103,7 @@ export function InputTextarea({
         classes,
         actions,
       });
-      editor.current.content.innerHTML = currentValue;
+      editor.current.content.innerHTML = value;
     }
   }, []);
 
@@ -120,7 +120,7 @@ export function InputTextarea({
       <div className={`editor ${css.editor}`} />
       <input
         type="hidden"
-        value={currentValue}
+        value={value}
         name={name}
         required={required}
         ref={myRef}
--- a/client/app/components/Input/InputTextarea.jsx
+++ b/client/app/components/Input/InputTextarea.jsx
@@ -66,13 +66,13 @@ export type Props = {
 export function InputTextarea({
   id,
   name,
-  value: propsValue,
+  value: propValue,
   required,
   hasError,
   myRef,
   dark,
 }: Props) {
-  const [value, setValue] = useState<string>(propsValue || '');
+  const [value, setValue] = useState<string>(propValue || '');
   const editorRef = useRef(null);
   const editor = useRef(null);
 
--- a/client/app/components/Input/InputTextarea.jsx
+++ b/client/app/components/Input/InputTextarea.jsx
@@ -74,7 +74,7 @@ export function InputTextarea({
 }: Props) {
   const [currentValue, setValue] = useState<string>(value || '');
   const editorRef = useRef(null);
-  let editor;
+  const editor = useRef(null);
 
   const onChange = (updatedValue: string) => {
     setValue(updatedValue);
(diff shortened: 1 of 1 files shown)


client/app/components/Input/InputTextarea.jsx
```diff
@@ -67,84 +67,75 @@ export type State = {
...
+      value={currentValue}
+      name={name}
+      required={required}
+      ref={myRef}
+    />
+  );
+
+  useEffect(() => {
+    if (editorRef) {
+      editor = init({
```
udayanshevade, line 114: Defining `editor` inside `useEffect()` (after a dependency list is added) will cause `editor` to be `undefined` in subsequent renders, since if `useEffect` only runs on mount then this definition will be skipped and the newly defined instance will be inaccessible outside this function scope.

One [option](https://reactjs.org/docs/hooks-reference.html#useref) is to `useRef` to store the value:

>However, `useRef()` is useful for more than the `ref` attribute. It’s [handy for keeping any mutable value around](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) similar to how you’d use instance fields in classes.

```
const editor = useRef(null);
...
useEffect(() => {
  ...
  editor.current = init({ ... });
  editor.current.content.innerHTML = currentValue;
}, []);
```

client/app/components/Input/InputTextarea.jsx
```diff
@@ -67,84 +67,75 @@ export type State = {
...
+  id,
+  name,
+  value,
+  required,
+  hasError,
+  myRef,
+  dark,
+}: Props) {
+  const [currentValue, setValue] = useState<State>(value || '');
+  let editorRef;
```
udayanshevade, line 80: Rather than use a callback ref, `useRef()` can be used instead:

```
  const editorRef = useRef(null);
  ...
  return (
    <div
      id={id}
      ...
      ref={editorRef}
    >
      ...
    </div>
  );
```

Then references to `editorRef` would be replaced with `editorRef.current`, e.g. inside `onFocus`:

```
  if (editorRef.current) {
    editorRef.current.getElementsByClassName('editorContent')[0].focus();
  }
```

udayanshevade (changes requested):
Hey @FranceJ looks great! 👍 just a few notes below. Once a couple of issues are resolved, the PR should be good to go.

client/app/components/Input/InputTextarea.jsx
```diff
@@ -67,84 +67,75 @@ export type State = {
...
+    if (editorRef) {
+      editor = init({
+        element: editorRef.getElementsByClassName('editor')[0],
+        onChange,
+        classes,
+        actions,
+      });
+      editor.content.innerHTML = currentValue;
+    }
+  });
```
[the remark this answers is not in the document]
FranceJ: that makes a ton of sense, I'll update the code

client/app/components/Input/InputTextarea.jsx
FranceJ: Two questions on this:
* All this code is doing is initializing Pell, attaching it to the ref passed by React, passing in methods to deal with changes, and then filling it with the starting content right? Wouldn't this only ever need to be done when the component is initialized? Why would we need to access the editor variable outside of that scope, and on re-render? Or, is this just so that future people have access if they want it?
* If we do want to keep the reference to editor alive past the initial render, wouldn't it make more sense to keep it around as a state variable rather than useRef? If we were for some reason to change the value of editor later on wouldn't we want to trigger a re-render of the component?

udayanshevade: >Wouldn't this only ever need to be done when the component is initialized? Why would we need access to the editor variable outside the scope and on re-render? Or, is this just so that future people have access if they want it?

Right, definitely! I think it makes sense to store the `editor` outside the callback, for extending editor-derived functionality (and also because it is analogous to storing a value on an instance field in a class component, i.e. `this.editor` becomes `editor.current`).

But currently it would also work _not_ to assign the editor to a variable at all, since it isn't being used anywhere else. I think the original implementation stored it on the class as a way to keep it handy if needed.

>as a state variable rather than useRef? If we were for some reason to change the value of editor later on wouldn't we want to trigger a re-render of the component?

I think it makes sense to wait for that use case to become clearer (since there are different possible paths for the implementation to go). For a more dynamic handling of the `editor`, it may become necessary to use a different approach at that point. For now, updating the `value` in state is enough for rendering the hidden input.

I guess it goes back to `useRef` being analogous to an instance field. Would we store `editor` in the state if it were a class component? Since the `editor` is currently autonomous and doesn't directly affect what the component renders, I think it can remain a static value 👍

FranceJ: This is a good explanation, I agree. I'll make the change.

client/app/components/Input/InputTextarea.jsx
```diff
@@ -67,84 +67,75 @@ export type State = {
...
-      this.editor.content.innerHTML = value;
-    }
-  }
-
-  onChange = (value: string) => {
-    this.setState({ value });
+export function InputTextarea({
+  id,
+  name,
+  value,
```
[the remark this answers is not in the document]
FranceJ: Makes sense, I've added the alias now

udayanshevade (approved):
LGTM @FranceJ! :tada: I'm not sure why the CircleCI / CodeClimate checks are failing though 😕 I checked everything out locally and the tests were all passing. Not sure if it's just the flaky tests mentioned [here](https://github.com/ifmeorg/ifme/wiki/Code-Review-Practices#merging-to-main)? Maybe the pipeline just has to be rerun

FranceJ:
@udayanshevade looks like everything is passing except the code complexity check. Is this blocking? If so, do you have any recommendations?

udayanshevade:
@FranceJ I'm not sure 😅 I haven't seen this issue before and don't see why it's objecting to the changes, when nothing is striking me as problematic. Did the destructuring change help? I thought it was ok how it was before

julianguyen:
I just put a won't fix label on the Codeclimate issues because they aren't blockers! So feel free to merge when your'e ready.

julianguyen (approved):
Thanks @udayanshevade for your thoughtful reviews! :D

Issue #1560, opened 2019-09-27
Functionality to download your user data


julianguyen:
# Description 

We want to give users the option of being able to download their account data. Let's include a button on the Account page (when a user is signed in) that allows the user to download their data.

A good starting point is downloading the data in a CSV file. We'll want to give them the ability to download data regarding:
* Their Moments, Categories, Moods, Medications, and Strategies (and any comments in them)
* Information on groups (and corresponding meetings and comments) that are a leader of 
* Other than comments from allies in their posts, a user should not have access to another user's data

Acollie:
Hey a little new to Open source, but I have experience with rails 
I would gladly help out with this! 😄

james947:
Hey @Acollie still working on this? I'd love to give it a try cc: @julianguyen

Acollie:
I have mostly done with the task. I just got a little busy, to fully finish the task. 
Left todo
- there is currently a bug with the export I wrote, which I didn't have time to try and fix but it should be fairly simple to fix.
- Write a Test for the data export
I have a branch which has all the code I wrote called 'data_export'

james947:
This is well noted @Acollie. Good work 👍

julianguyen:
How's this going @Acollie? :)

Acollie:
Hey @julianguyen , I really appreciate you checking up on me :) 
I reached a bug which found it hard to fix. I also found it hard to write tests, for the feature.

Would it be possible if you or someone could try and give me a hand, since I am keen to finish this feature.

julianguyen:
No worries and thanks for the update @Acollie! I can help take a look, do you have an existing branch? :)

Acollie:
I do the 'data_export' should have my version but its quite out of date with the main branch.

julianguyen:
I have some bandwidth to help finish this, starting sometime this week.

Acollie:
@julianguyen I would like to try to finish as well. DM me on Slack and we can try and finish this!

akp2603:
Hi! I would be really interested to work on this one.
Is this still open?
(Will be needing a little guidance, this is going to be my first open source contribution)

chethanagopinath:
Hey @akp2603, I was a bit lost on this one for a while, and I've been meaning to work on this forever, but yea you can pick this up if you want! I'll unassign myself.

akp2603:
hi @julianguyen! Can you assign me the issue and can we discuss this over slack once?

[...]

PR #1900, opened 2020-10-10
Data to csv

akp2603:
# Description

The following PR includes backend APIs for the aforementioned issues. 

## More Details

 - Provided 3 apis. One to generate a request for data to be generated, one to fetch the status of the request being queued(CSV generation happens async, hence this api for polling.), and last one to download the file.
- The following PR enables a user to generate a csv to get his/her data. With every new csv generated, the old csv get deleted, to prevent disk space.
- Files, 30 days old, will be deleted through a cron.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1560

---


bennpham (commented):
Hi Aashish, this PR looks nice for a proof of concept for a job scheduler or data backup. Good work on it 👍 . If I made a scheduler, I'd technically store it on Amazon S3 or Google Cloud service instead where data storage would be cheaper and less costlier than storing everything on disc. Especially when more users were to register, it'd be difficult to scale. 

On the other hand, for an actual csv download itself, I'm wondering but couldn't you just do the following (modify below to restrict it to only users who owned that data):

**Model.rb**
```ruby
def self.to_csv
  heading = ["Col1", "Col2", "Col3]

  CSV.generate do |csv|
    csv << heading
      
    Model.all.each do |model|
      csv << [model.data1, model.data2, model.data3]
    end
end
```

**Controller.rb**
```ruby
def index
  ...

  respond_to do |format|
    format.html
    format.csv {
      send_data Model.to_csv, filename: "model-data-#{Date.current.strftime('%Y%m%d')}.csv"
    }
  end
end
```

**index.html.erb**
```
<%= link_to "Export to CSV", { format: :csv }, class: "btn" %>
```

akp2603:
> If I made a scheduler, I'd technically store it on Amazon S3 or Google Cloud service instead where data storage would be cheaper and less costlier than storing everything on disc. Especially when more users were to register, it'd be difficult to scale.

Hi @bennpham firstly thanks for taking the time to review my PR. Really appreciated. Secondly, regarding what you said, I second this, but I asked @julianguyen regarding the same before implementing it. I suggested what you suggested. But she preferred it storing it on the disk only. Hence I implemented it this way. Maybe we can switch to S3 as an enhancement over this?

Also, regarding the csv file, what you've suggested is separated files for separate models. But I implemented it all in a single file. (The format for which, was again, confirmed with Julia.). Can share it with you on slack too, if you want.

Please let me know if you still want any changes regarding the same.

app/models/allyship.rb
```diff
@@ -12,6 +12,14 @@
 #
 
 class Allyship < ApplicationRecord
+  DISPLAY_ATTRIBUTES = %w[
```
julianguyen, line 15: Let's rename this to something like `USER_DATA_ATTRIBUTES` so that it's more obvious that this is being used for that.

julianguyen (commented):
I don't have deep expertise this kind of data storage, so I'm definitely open to whatever option is free 🙃 Just cause we have limited funds at the moment as an open source project.

julianguyen:
In terms of recommendations for learning how to test, I recommend looking at our existing specs for general patterns and ideas. The official RoR docs has a good starter [doc](https://guides.rubyonrails.org/v3.2/testing.html) on testing. 

Feel free to ask for help in our #dev channel on Slack!

config/env/development.example.env
```diff
@@ -39,3 +39,7 @@ RAISE_DELIVERY_ERRORS="false"
 PSQL_HOST=""
 PSQL_USERNAME=""
 PSQL_PASSWORD=""
+
+# REDIS
```
julianguyen, line 43: You'll want to add this to `config/env/test.example.env` as well.

akp2603: Done.

app/helpers/users/reports_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+module Users
+  module ReportsHelper
+    def submit_request_helper(user)
+      [200, { request_id: user.generate_data_request }]
+    rescue StandardError => e
+      [422, { error: e.message }]
+    end
+
+    def fetch_request_status_helper(user, request_id)
+      return 400, { error: 'Request id can be blank.' } if request_id.blank?
```
julianguyen, line 11: Should this be "Request id can't be blank" instead?

app/helpers/users/reports_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+      data_request = user.data_requests.find_by(request_id: request_id)
+      if data_request.blank?
+        return 404, { error: 'No such request exists for current user.' }
+      end
+
+      [200, { current_status: data_request.status_id }]
+    end
+
+    def download_data_helper(user, request_id)
+      return 400, { error: 'Request id can be blank.' } if request_id.blank?
```
julianguyen, line 22: Same question as above.

app/workers/process_data_request_worker.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class ProcessDataRequestWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: 'critical'
+
+  def perform(request_id)
+    data_request = Users::DataRequest.find_by(request_id: request_id)
+    return if data_request.blank?
```
julianguyen, line 8: Lines 8 and 9 could be combined with an `||`.

spec/models/users/data_request_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+RSpec.describe Users::DataRequest, type: :model do
+  context 'validations' do
+    it 'is invalid without a request_id' do
+      data_request = build(:empty_request_id_data_request)
+      expect(data_request).to have(1).error_on(:request_id) 
+    end
+
+    it 'is invalid without a status_id' do
+      data_request = build(:partial_data_request)
+      expect(data_request).to have(2).error_on(:status_id)
```
julianguyen, line 22: What are there two errors?

spec/requests/users/reports_request_spec.rb
```diff
@@ -0,0 +1,94 @@
+RSpec.describe "Users::Reports", type: :request do
+  let(:user) { create(:user) }
+
+  describe '#submit_request' do
+    context 'when the user is logged in' do
+      before { sign_in user }
+
+      it 'creates a data download request' do
+        post "/users/data"
```
julianguyen, line 9: Let's replace this and other occurrences with `users_data_path`.

spec/requests/users/reports_request_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      it_behaves_like :with_no_logged_in_user
+    end
+  end
+
+  describe '#fetch_request_status' do
+    context 'when the user is logged in' do
+      before { sign_in user }
+
+      it 'fetches the status of data request with a blank request_id' do
+        get "/users/data/status"
```
julianguyen, line 28: Let's replace this and other occurrences with `users_data_status`.

spec/requests/users/reports_request_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      it_behaves_like :with_no_logged_in_user
+    end
+  end
+
+  describe '#download_data' do
+    context 'when the user is logged in' do
+      before { sign_in user }
+
+      it 'fetches the file with a blank request_id' do
+        get "/users/data/download"
```
julianguyen, line 61: Let's replace this and other occurrences with `users_data_download`.

julianguyen (commented):
Strong work! Thanks for looking into writing tests :D

spec/models/users/data_request_spec.rb
akp2603: 1. presence.
2. inclusion.
Both are treated separatedly. There are 2 validations on status_id. `nil` is blank and also not amongst the included values.

julianguyen (approved):
Fantastic work! Thanks so much for taking this on and going through all the review cycles here :)

PR #1903, opened 2020-10-10
Refactor notification controller spec to notification request spec

Ashwithabg:
#1831

### DIFF
--- a/spec/requests/notifications_spec.rb
+++ b/spec/requests/notifications_spec.rb
@@ -1,24 +1,23 @@
 # frozen_string_literal: true
 
-RSpec.describe NotificationsController, type: :controller do
+describe 'Notifications', type: :request do
   describe '#destroy' do
-    let(:user) { FactoryBot.create(:user1) }
-    let(:other_user) { FactoryBot.create(:user2) }
+    let(:user) { create(:user1) }
+    let(:other_user) { create(:user2) }
     let(:notification_owner) { user }
     let!(:notification) do
-      FactoryBot.create(:notification, user: notification_owner)
+      create(:notification, user: notification_owner)
     end
 
     context 'when the user is signed in' do
       let(:previous_page) { 'http://example.com/previous' }
       let(:notification_id) { notification.id }
-      let(:format) { 'html' }
+      let(:headers) { { 'HTTP_REFERER' => previous_page } }
 
       before do
-        allow(controller).to receive(:current_user).and_return(user)
-        request.env['HTTP_REFERER'] = previous_page
+        sign_in user
 
-        delete :destroy, format: format, params: { id: notification_id }
+        delete notification_path(id: notification_id), headers: headers
       end
 
       context 'and the notification to be deleted exists' do
@@ -28,7 +27,7 @@ RSpec.describe NotificationsController, type: :controller do
           end
 
           context 'and the requested format is html' do
-            let(:format) { 'html' }
+            let(:headers) { { 'HTTP_REFERER' => previous_page } }
 
             it 'redirects the user back to where they were before' do
               expect(response).to redirect_to previous_page
@@ -36,7 +35,10 @@ RSpec.describe NotificationsController, type: :controller do
           end
 
           context 'and the requested format is json' do
-            let(:format) { 'json' }
+            let(:headers) do
+              { 'HTTP_REFERER' => previous_page,
+                'ACCEPT' => 'application/json' }
+            end
 
             it 'renders a HEAD response with :no_content' do
               expect(response).to have_http_status 204
@@ -52,7 +54,7 @@ RSpec.describe NotificationsController, type: :controller do
           end
 
           context 'and the requested format is html' do
-            let(:format) { 'html' }
+            let(:headers) { { 'HTTP_REFERER' => previous_page } }
 
             it 'redirects the user back to where they were before' do
               expect(response).to redirect_to previous_page
@@ -60,7 +62,10 @@ RSpec.describe NotificationsController, type: :controller do
           end
 
           context 'and the requested format is json' do
-            let(:format) { 'json' }
+            let(:headers) do
+              { 'HTTP_REFERER' => previous_page,
+                'ACCEPT' => 'application/json' }
+            end
 
             it 'renders a HEAD response with :no_content' do
               expect(response).to have_http_status 204
@@ -73,7 +78,7 @@ RSpec.describe NotificationsController, type: :controller do
         let(:notification_id) { 'something-fake' }
 
         context 'and the requested format is html' do
-          let(:format) { 'html' }
+          let(:headers) { { 'HTTP_REFERER' => previous_page } }
 
           it 'redirects the user back to where they were before' do
             expect(response).to redirect_to previous_page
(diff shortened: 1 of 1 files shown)


spec/requests/notifications_spec.rb
```diff
@@ -1,24 +1,23 @@
 # frozen_string_literal: true
 
-RSpec.describe NotificationsController, type: :controller do
+describe 'Notifications', type: :request do
```
bennpham, line 3:
```suggestion
describe 'Notification', type: :request do
```

Nitpick: I noticed that some of the request _spec are plural and some are singular. I think for consistency, let's stick to singular to match the rest of the example request specs: https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec . Same goes for filename.

bennpham (commented):
I'll probably get through reviewing the rest of the tests later, but I'm finding the multiple context blocks a bit tricky to follow. It's DRY, but not DAMP so it's hard for me to follow which action I'm trying to assert as I go down the tests. 

I think multiple nested context with `and` is an anti pattern. There are some projects I work on with default rubocop rule to prevent more than 2-3 nested context/describe blocks. I'd change some of them to say `context 'when ...' do` and merge some of those together.

spec/requests/notifications_spec.rb
tyler-wel: I've always done it to match the controllers/routing. So if it's `notifications_controller` or `GET /notifications`, I'd do `notifications_spec`. But there's also some inconsistent naming in the controllers too so 🤷‍♀️

spec/requests/notifications_spec.rb
```diff
@@ -1,24 +1,23 @@
...
       before do
-        allow(controller).to receive(:current_user).and_return(user)
-        request.env['HTTP_REFERER'] = previous_page
+        sign_in user
 
-        delete :destroy, format: format, params: { id: notification_id }
+        delete notification_path(id: notification_id), headers: headers
       end
 
       context 'and the notification to be deleted exists' do
```
tyler-wel, line 23: These names should be refactored. "and" naming should not be used.
https://www.betterspecs.org/#contexts

spec/requests/notifications_spec.rb
```diff
@@ -28,15 +27,18 @@
...
               expect(response).to redirect_to previous_page
             end
           end
 
           context 'and the requested format is json' do
-            let(:format) { 'json' }
+            let(:headers) do
+              { 'HTTP_REFERER' => previous_page,
+                'ACCEPT' => 'application/json' }
+            end
```
tyler-wel, line 41: I'm personally not a fan of this code format
```suggestion
            let(:headers) do
              { 
                'HTTP_REFERER' => previous_page,
                'ACCEPT' => 'application/json' 
              }
            end
```

spec/requests/notifications_spec.rb
```diff
@@ -52,15 +54,18 @@
...
+            let(:headers) { { 'HTTP_REFERER' => previous_page } }
 
             it 'redirects the user back to where they were before' do
               expect(response).to redirect_to previous_page
             end
           end
 
           context 'and the requested format is json' do
-            let(:format) { 'json' }
+            let(:headers) do
```
tyler-wel, line 65: I feel like this headers let block is repeated way too many times. Could be extracted to a higher context and shared. 
Make the accept block `'ACCEPT' => format` and define `let(:format)` where needed

spec/requests/notifications_spec.rb
```diff
@@ -124,7 +132,7 @@
 
     context 'when the user is signed in' do
       before do
-        allow(controller).to receive(:current_user).and_return(user)
+        sign_in user
```
tyler-wel, line 135: Don't need `do...end` here

```suggestion
      before { sign_in user }
```

tyler-wel (changes requested):

spec/requests/notifications_spec.rb
Ashwithabg: Have updated name of the spec to singular.

spec/requests/notifications_spec.rb
Ashwithabg: Changed `and the` to `when`  in context names.

spec/requests/notifications_spec.rb
Ashwithabg: Formatted the headers.

spec/requests/notifications_spec.rb
Ashwithabg: Updated this

spec/requests/notifications_spec.rb
Ashwithabg: In few context blocks, we need headers to have JSON format. And for a few HTML. That's the reason why headers are duplicated.

tyler-wel: My recommendation is to reduce the amount of duplicated long blocks of code by moving the headers block to a higher context and defining the format where needed
```
context 'when user is signed in' do
  let(:headers) do
    { 
      'HTTP_REFERER' => previous_page,
      'ACCEPT' => format
    }
  end

  context '...' do
    let(:format) { 'application/json' }
  end

  context '...' do
    let(:format) { 'text/html' }
  end
end
```

Ashwithabg: okay thank you. Will do this change.

Ashwithabg: I have pushed above changes. Could you please review.

spec/requests/notifications_spec.rb
```diff
@@ -69,19 +71,17 @@
...
       end
 
-      context 'and the notification to be deleted does not exist' do
+      context 'when notification to be deleted does not exist' do
         let(:notification_id) { 'something-fake' }
 
-        context 'and the requested format is html' do
-          let(:format) { 'html' }
-
+        context 'when requested format is html' do
```
tyler-wel, line 77: you should specify for all html contexts
```ruby
let(:format) { 'text/html' }
```

this ensure the expected behavior is actually happening

tyler-wel (changes requested):
@bennpham you good?

bennpham (approved):
@tyler-wel If all of your other reviews look good. I'm overall fine with this. Just a few nitpicks like some do/end could be reduce to { } to save some line spaces but nothing seem logically off for me now when I'm going through 👍 👍

spec/requests/notifications_spec.rb
```diff
@@ -1,66 +1,68 @@
...
   describe '#destroy' do
-    let(:user) { FactoryBot.create(:user1) }
-    let(:other_user) { FactoryBot.create(:user2) }
+    let(:user) { create(:user1) }
+    let(:other_user) { create(:user2) }
     let(:notification_owner) { user }
     let!(:notification) do
-      FactoryBot.create(:notification, user: notification_owner)
+      create(:notification, user: notification_owner)
     end
```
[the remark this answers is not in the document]
tyler-wel: Agreed, do..end here isn't necessary 👍

spec/requests/notifications_spec.rb
Ashwithabg: have made the changes. Thanks for the feedback

tyler-wel (approved):

PR #1904, opened 2020-10-11
Issus 1804 - migrate ToggleLocale to RTL

eric199wen:
# Description

Migrate `ToggleLocale` from Enzyme to React Test Library

## More Details

This is the first PR I created for this project and the first time I use RTL, please leave comments if I am not following the best practice.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1804

# Screenshots

NA
---

### DIFF
--- a/client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
+++ b/client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
@@ -1,9 +1,9 @@
 // @flow
 import React from 'react';
 import axios from 'axios';
-import { mount } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import Cookies from 'js-cookie';
-import { act } from 'react-dom/test-utils';
 import { ToggleLocale } from 'widgets/ToggleLocale';
 
 const locales = ['en', 'es', 'it', 'nb', 'nl', 'pt-BR', 'sv', 'vi'];
@@ -17,23 +17,17 @@ describe('ToggleLocale', () => {
   });
 
   it('does nothing if the previous locale is the same as the selected', () => {
-    const wrapper = mount(component);
-    act(() => {
-      wrapper.find('select').prop('onChange')({
-        currentTarget: { value: 'en' },
-      });
-    });
+    render(component);
+    userEvent.selectOptions(screen.getByRole('combobox'), 'en');
     expect(axios.post).not.toHaveBeenCalled();
   });
 
   it('sets the locale cookie and makes a post request if the selected locale is different from the previous', () => {
-    const wrapper = mount(component);
-    act(() => {
-      wrapper.find('select').prop('onChange')({
-        currentTarget: { value: 'es' },
-      });
-    });
+    render(component);
+    userEvent.selectOptions(screen.getByRole('combobox'), 'es');
     expect(Cookies.set).toHaveBeenCalledWith('locale', 'es');
-    expect(axios.post).toHaveBeenCalledWith('/toggle_locale', { locale: 'es' });
+    expect(axios.post).toHaveBeenCalledWith('/toggle_locale', {
+      locale: 'es',
+    });
   });
 });


client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
```diff
@@ -17,23 +22,25 @@ describe('ToggleLocale', () => {
   });
 
   it('does nothing if the previous locale is the same as the selected', () => {
-    const wrapper = mount(component);
+    render(component);
```
udayanshevade, line 25: 👍

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
```diff
@@ -17,23 +22,25 @@ describe('ToggleLocale', () => {
...
     act(() => {
-      wrapper.find('select').prop('onChange')({
-        currentTarget: { value: 'en' },
+      fireEvent.select(screen.getByRole('combobox'), {
```
udayanshevade, line 27: The dev dependency '@testing-library/user-event' is also available for simulating user behavior:

```
import userEvent from '@testing-library/user-event';

...

userEvent.selectOptions(screen.getByRole('combobox'), 'en');
```

It has some useful abstractions outside of `fireEvent`, as seen with the `selectOptions` [method](https://github.com/testing-library/user-event#selectoptionselement-values) used in this example.

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
```diff
@@ -17,23 +22,25 @@ describe('ToggleLocale', () => {
   });
 
   it('does nothing if the previous locale is the same as the selected', () => {
-    const wrapper = mount(component);
+    render(component);
     act(() => {
```
udayanshevade, line 26: `act` is not needed here 👍

See [this point](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily).

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
```diff
@@ -17,23 +22,25 @@ describe('ToggleLocale', () => {
...
   });
 
-  it('sets the locale cookie and makes a post request if the selected locale is different from the previous', () => {
-    const wrapper = mount(component);
+  it('sets the locale cookie and makes a post request if the selected locale is different from the previous', async () => {
+    render(component);
     act(() => {
-      wrapper.find('select').prop('onChange')({
-        currentTarget: { value: 'es' },
+      fireEvent.change(screen.getByRole('combobox'), {
```
udayanshevade, line 37: Same as above here:

`act` can be removed and `userEvent` can be used instead of `fireEvent`.

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
```diff
@@ -17,23 +22,25 @@ describe('ToggleLocale', () => {
...
+        target: { value: 'es' },
       });
     });
     expect(Cookies.set).toHaveBeenCalledWith('locale', 'es');
-    expect(axios.post).toHaveBeenCalledWith('/toggle_locale', { locale: 'es' });
+    await waitFor(() => expect(axios.post).toHaveBeenCalledWith('/toggle_locale', {
```
udayanshevade, line 42: Since this test is only checking if the request gets triggered, it is not necessary to `await` the response. The request does not need to conclude, as the response itself is not used here. We can simply expect `axios.post` to have been called (even if it concludes async). For now, I think this assertion can be kept synchronous like it was before 👍

udayanshevade (changes requested):
Looks great @eric199wen! 🎉 checked out your changes and the tests pass. Just a couple of small notes below on usage. Then it should be good to go 👍 

(Btw I found [this article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) useful while writing tests with RTL, if you haven't checked it out already).

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
eric199wen: I didn't notice that we are already using it! Thank you for pointing that out.

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
eric199wen: Good call! I didn't think about it since I think it is not really related to the migration.

client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
eric199wen: Makes sense. Thanks!

udayanshevade (approved):
LGTM @eric199wen :tada: thanks for making those small changes!

PR #1905, opened 2020-10-11
Migrate comments test

eric199wen:
# Description

Migrate `Comments` component from Enzyme to RTL.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1800

# Screenshots

NA

---

### DIFF
--- a/client/app/widgets/Comments/__tests__/Comments.spec.jsx
+++ b/client/app/widgets/Comments/__tests__/Comments.spec.jsx
@@ -1,12 +1,11 @@
 // @flow
 import React from 'react';
 import axios from 'axios';
-import { mount } from 'enzyme';
-import { act } from 'react-dom/test-utils';
+import {
+  act, render, screen, waitFor,
+} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import Comments from 'widgets/Comments';
-
-let axiosPostSpy;
-let axiosDeleteSpy;
 
 const formProps = {
   inputs: [
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
 
 describe('Comments', () => {
   beforeEach(() => {
-    axiosPostSpy = jest.spyOn(axios, 'post').mockImplementation(() => Promise.resolve({
+    jest.spyOn(axios, 'post').mockImplementation(() => Promise.resolve({
       data: { comment },
     }));
-    axiosDeleteSpy = jest.spyOn(axios, 'delete').mockImplementation(() => Promise.resolve({
+    jest.spyOn(axios, 'delete').mockImplementation(() => Promise.resolve({
       data: { id },
     }));
   });
 
   it('renders correctly', () => {
-    let wrapper = null;
-    expect(() => {
-      wrapper = mount(component);
-    }).not.toThrow();
-    expect(wrapper).not.toBeNull();
+    render(component);
+    expect(() => component).not.toThrow();
+    expect(component).not.toBeNull();
   });
 
   it('add and delete a comment', async () => {
-    const wrapper = mount(component);
-    expect(wrapper.find('.comment').exists()).toEqual(false);
-    act(() => {
-      wrapper
-        .find('input[name="comment[comment]"]')
-        .simulate('change', { currentTarget: { value } });
+    render(component);
+    expect(screen.queryByRole('article')).toBeNull();
 
-      wrapper.find('select#comment_visibility').prop('onChange')({
-        currentTarget: { value: 'private' },
-      });
+    act(() => {
+      userEvent.type(screen.getByRole('textbox'));
+      userEvent.selectOptions(screen.getByRole('combobox'), 'private');
+      userEvent.click(screen.getByRole('button', { name: 'Submit' }));
     });
-    wrapper.find('input[type="submit"]').simulate('click');
-    await axiosPostSpy();
-    wrapper.update();
-    expect(wrapper.find('.comment').exists()).toEqual(true);
-    wrapper.find('.storyActionsDelete').find('a').simulate('click');
-    await axiosDeleteSpy();
-    wrapper.update();
-    expect(wrapper.find('.comment').exists()).toEqual(false);
+
+    await waitFor(() => expect(screen.queryByRole('article')).not.toBeNull());
+
+    userEvent.click(screen.getByRole('link', { name: 'Delete' }));
+
+    await waitFor(() => expect(screen.queryByRole('article')).toBeNull());
   });
 });
--- a/client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
+++ b/client/app/widgets/ToggleLocale/__tests__/ToggleLocale.spec.jsx
@@ -1,9 +1,9 @@
 // @flow
 import React from 'react';
 import axios from 'axios';
-import { mount } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import Cookies from 'js-cookie';
-import { act } from 'react-dom/test-utils';
 import { ToggleLocale } from 'widgets/ToggleLocale';
 
 const locales = ['en', 'es', 'it', 'nb', 'nl', 'pt-BR', 'sv', 'vi'];
@@ -17,23 +17,17 @@ describe('ToggleLocale', () => {
   });
 
   it('does nothing if the previous locale is the same as the selected', () => {
-    const wrapper = mount(component);
-    act(() => {
-      wrapper.find('select').prop('onChange')({
-        currentTarget: { value: 'en' },
-      });
-    });
+    render(component);
+    userEvent.selectOptions(screen.getByRole('combobox'), 'en');
     expect(axios.post).not.toHaveBeenCalled();
   });
 
   it('sets the locale cookie and makes a post request if the selected locale is different from the previous', () => {
-    const wrapper = mount(component);
-    act(() => {
-      wrapper.find('select').prop('onChange')({
-        currentTarget: { value: 'es' },
-      });
-    });
+    render(component);
+    userEvent.selectOptions(screen.getByRole('combobox'), 'es');
     expect(Cookies.set).toHaveBeenCalledWith('locale', 'es');
-    expect(axios.post).toHaveBeenCalledWith('/toggle_locale', { locale: 'es' });
+    expect(axios.post).toHaveBeenCalledWith('/toggle_locale', {
+      locale: 'es',
+    });
   });
 });


julianguyen (approved):
Excellent work! Thanks so much for taking this on!

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
 
 describe('Comments', () => {
   beforeEach(() => {
-    axiosPostSpy = jest.spyOn(axios, 'post').mockImplementation(() => Promise.resolve({
+    jest.spyOn(axios, 'post').mockImplementation(() => Promise.resolve({
```
udayanshevade, line 85: There is also a nice [shorthand](https://jestjs.io/docs/en/mock-function-api#mockfnmockreturnvaluevalue) for this

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
...
   });
 
   it('renders correctly', () => {
-    let wrapper = null;
-    expect(() => {
-      wrapper = mount(component);
-    }).not.toThrow();
-    expect(wrapper).not.toBeNull();
+    render(component);
+    expect(() => component).not.toThrow();
```
udayanshevade, line 95: Q: Is this needed? I'm wondering if this assertion is making a difference.

The previous 'enzyme' version of this test would `mount` the component and expect it not to throw. Now this appears to be equivalent to passing in an un-invoked function component:

```
const NewComponent = () => component;
expect(NewComponent).not.toThrow();
```

Does that work? 🤔 I'm not sure. I haven't used it this way before.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
...
 
   it('add and delete a comment', async () => {
-    const wrapper = mount(component);
-    expect(wrapper.find('.comment').exists()).toEqual(false);
-    act(() => {
-      wrapper
-        .find('input[name="comment[comment]"]')
-        .simulate('change', { currentTarget: { value } });
+    render(component);
+    expect(screen.queryByRole('article')).toBeNull();
```
udayanshevade, line 101: There is also the matcher `toBeInTheDocument()` that can be used 👍

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
...
 
   it('renders correctly', () => {
-    let wrapper = null;
-    expect(() => {
-      wrapper = mount(component);
-    }).not.toThrow();
-    expect(wrapper).not.toBeNull();
+    render(component);
+    expect(() => component).not.toThrow();
+    expect(component).not.toBeNull();
```
udayanshevade, line 96: Since `component` is defined above, it appears this assertion can't fail, since `component` can never be `null`.

Should this instead be checking against the result of _rendering_ the component, e.g.:

```
const wrapper = render(component);
expect(wrapper).not.toBeNull();
```

But from the docs, I couldn't tell if `wrapper` can ever be `null` either. I tried checking what happens without any arguments to `render` and it still worked.

Would it make more sense to check if the expected child components are rendered instead?

```
it('renders correctly', () => {
    render(component);
    expect(screen.queryByRole('article')).not.toBeInTheDocument(); // no comment yet
    expect(screen.getByRole('button', { name: 'Submit' })).toBeInTheDocument(); // form exists
});
```

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
...
-      wrapper
-        .find('input[name="comment[comment]"]')
-        .simulate('change', { currentTarget: { value } });
+    render(component);
+    expect(screen.queryByRole('article')).toBeNull();
 
-      wrapper.find('select#comment_visibility').prop('onChange')({
-        currentTarget: { value: 'private' },
-      });
+    act(() => {
```
udayanshevade, line 103: Is this needed? I think `act` can be removed here 👍

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +82,34 @@ const component = <Comments formProps={formProps} />;
...
-    wrapper.find('input[type="submit"]').simulate('click');
-    await axiosPostSpy();
-    wrapper.update();
-    expect(wrapper.find('.comment').exists()).toEqual(true);
-    wrapper.find('.storyActionsDelete').find('a').simulate('click');
-    await axiosDeleteSpy();
-    wrapper.update();
-    expect(wrapper.find('.comment').exists()).toEqual(false);
+
+    await waitFor(() => expect(screen.queryByRole('article')).not.toBeNull());
```
udayanshevade, line 109: Would it also make sense to check how the comment is rendered, e.g. get the text displayed in the comment?

udayanshevade (commented):
@eric199wen looks great! 👍 just left a couple of notes/questions that occurred to me as I was going through the updates. Lmk what you think!

client/app/widgets/Comments/__tests__/Comments.spec.jsx
eric199wen: I am not really familiar with `jest` and `RTL` best practice but I think your suggestion makes more sense based on the test description `it renders correctly`.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
eric199wen: TBH I am not sure about this. Based on the [documentation](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme) `render` in RTL is more like `mount` than `shallow`. But I am not really familiar with jest also so I don't know the answer.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
eric199wen: Seems like we don't need it. I still don't understand when to use `act`. I thought we change state multiple times in three events so that was why I used act. I think the purpose is to make sure assertion after all the state update is done right? Or is it because we've already used `await` and `waitFor` to wait until comment is present so there is no need of `act`?

client/app/widgets/Comments/__tests__/Comments.spec.jsx
eric199wen: That is more comprehensive. I will add it.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
udayanshevade: >render in RTL is more like `mount`

I think you're right 👍  whereas enzyme's `shallow` avoids rendering the component's children, both `render` and `mount` do render the full tree.

Btw I tried a quick analog to see if that previous assertion on line 95 works:

```
  it('See if component throws', () => {
    const Foo = () => {
      throw new Error('oops'); // definitely throws an error
    };
    const foo = <Foo />;
    // render(foo); // commenting this line out for now
    const WrappedFoo = () => foo;
    expect(WrappedFoo).not.toThrow(); // similar scenario
  });
```

And the test still passes -- even though it should fail.

That suggests only passing in a component, without calling or rendering it, to an assertion does not `throw`. And if it can never fail, the test can't be reliable.

Now if you uncomment the `render` line though, the test _does_ fail. Because rendering it actually throws the error inside `Foo` and causes the test as a whole to fail. So previously, it looks like line 94 was doing line 95's work.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
udayanshevade: Awesome! In a rush, I missed that the component also renders a `section` around all comments. That was another 'role' that could be queried. But I think `article` still works 👍

client/app/widgets/Comments/__tests__/Comments.spec.jsx
udayanshevade: >I still don't understand when to use `act`

Yeah for sure it's confusing.

>the purpose is to make sure assertion after all the state update is done right?

Exactly, it flushes batched changes that may occur during async code outside of React's awareness.

>is it because we've already used `await` and `waitFor`

Right, I think it's because RTL's `waitFor` comes wrapped in `act` already, so it takes care of that for us. From what I know `act` is covered when working within RTL's async methods.

I'm going to be checking out [this article](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning) as I haven't read it yet. It seems to go in-depth on the use of `act` in specific scenarios. If you get a chance to go through it too (or if you already have), lmk if it makes sense 👍

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +80,33 @@ const component = <Comments formProps={formProps} />;
 
 describe('Comments', () => {
   beforeEach(() => {
-    axiosPostSpy = jest.spyOn(axios, 'post').mockImplementation(() => Promise.resolve({
+    jest.spyOn(axios, 'post').mockResolvedValue({
```
udayanshevade, line 83: 👍

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +80,33 @@ const component = <Comments formProps={formProps} />;
...
   });
 
   it('renders correctly', () => {
-    let wrapper = null;
-    expect(() => {
-      wrapper = mount(component);
-    }).not.toThrow();
-    expect(wrapper).not.toBeNull();
+    render(component);
+    expect(screen.queryByRole('article')).not.toBeInTheDocument();
```
udayanshevade, line 93: My mistake. I didn't see earlier that this [component also renders](https://github.com/ifmeorg/ifme/blob/main/client/app/widgets/Comments/index.jsx#L135) a `section` around all the comments, which would make sense to check too. But I think checking for an `article` also works 👍

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +80,33 @@ const component = <Comments formProps={formProps} />;
...
-    wrapper.find('.storyActionsDelete').find('a').simulate('click');
-    await axiosDeleteSpy();
-    wrapper.update();
-    expect(wrapper.find('.comment').exists()).toEqual(false);
+    userEvent.type(screen.getByRole('textbox'));
+    userEvent.selectOptions(screen.getByRole('combobox'), 'private');
+    userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+
+    await waitFor(() => expect(screen.queryByRole('article')).toBeInTheDocument());
+    expect(screen.queryByRole('article')).toHaveTextContent('Hey');
```
udayanshevade, line 106: Nice 👍 I like this!

Minor note: when querying for presence, `getByRole` [can be used](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence) instead.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -83,41 +80,33 @@ const component = <Comments formProps={formProps} />;
...
+
+    userEvent.click(screen.getByRole('link', { name: 'Delete' }));
+
+    await waitFor(() => expect(screen.queryByRole('article')).not.toBeInTheDocument());
```
udayanshevade, line 110: 👍

udayanshevade (approved):
LGTM too @eric199wen! Just left a couple of minor notes.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
eric199wen: I'll keep it as is.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
eric199wen: I updated to use `getByRole`.

PR #1911, opened 2020-10-14
feat: #1803 Migrate client/app/widgets/Resources/__tests__ to RTL

udayanshevade:
# Description

Migrates test suite from enzyme to RTL

## More Details

I was wondering about the repetition in the tests. While trying to understand the functionality and test code, I thought maybe there was a way to simplify things. So I experimented with abstracting some reusable helpers.

For example, instead of:

```
resourcesArticles = screen.getAllByRole('article');
expect(resourcesArticles).toHaveLength(1);
// results count/pagination
expect(screen.getByText(/1 of 1/)).toBeInTheDocument();
```

I tried moving it into a helper that could be reused across the tests:

```
const testResourcesCount = (count) => {
  resourcesArticles = screen.getAllByRole('article');
  expect(resourcesArticles).toHaveLength(count);
  // results count/pagination
  expect(screen.getByText(new RegExp(`${count} of ${count}`))).toBeInTheDocument();
};

...

testResourcesCount(3);
```

Wasn't feeling too sure about it though, so only pushing the refactor at first without any abstractions 👍 

## Corresponding Issue

#1803 

Would appreciate any thoughts on that issue. And as always, if any changes are needed please LMK and I'll follow up with the updates.

Thanks! :coffee:

---

### DIFF
--- a/client/app/components/Input/__tests__/InputTag.spec.jsx
+++ b/client/app/components/Input/__tests__/InputTag.spec.jsx
@@ -19,7 +19,6 @@ describe('InputTag', () => {
 
     // simulate searching for an option
     userEvent.type(input, 'Two');
-    userEvent.click(input);
 
     const queryOptions = { name: checkboxLabelTwo };
     let checkbox = screen.queryByRole('checkbox', queryOptions);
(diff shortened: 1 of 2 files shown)


julianguyen (approved):
Awesome work! Thanks for all these great RTL refactors!

So I personally think you don't need to abstract that assertion into a helper. Though repetitive, I think it's more readable to not thave that abstraction. 

I think the helper would be useful if there was repetition across test files.

udayanshevade:
> So I personally think you don't need to abstract that assertion into a helper. Though repetitive, I think it's more readable to not thave that abstraction.
> I think the helper would be useful if there was repetition across test files.

For sure! 👍 will keep it as is. Makes sense, if there is repetition across files then maybe that can be its own issue if it becomes needed

Issue #1909, opened 2020-10-14
Fix autosuggest selection using 'enter' keydown in QuickCreate widget


udayanshevade:
# Description

**Type**: Bug
**Environment**: Development/local
**Branch**: 'main' (up to date)
**Browser**: Chrome 85.0.4183.121
**OS**: MacBook Pro Catalina 10.15.7

Note: I saw this happen while testing a PR locally, but wasn't able to reproduce it on the live site.

In the `QuickCreate` widget, pressing 'enter' while selecting a dropdown menu option has unexpected behavior:
- if the input value is empty:
    - and a dropdown list item is selected, it throws an error (screenshot below).
- or if the input value matches one of the list options:
    - and that matching item is selected, the create modal is still opened

I believe these are two separate problems, but they are originating from the same component/behavior.

If it makes sense, they can also be split into different tasks, but I wasn't sure whether they're distinct enough.

# Reproduction Steps

1. Try to create/edit a Strategy in the Strategies page.
2. Focus on the Autocomplete input, or press 'down' if already focused, to open the dropdown menu
3. Try to select one of the options from the dropdown menu (without an input, or with an input).

# Screenshots

<img width="934" alt="Screen Shot 2020-10-13 at 4 57 54 PM" src="https://user-images.githubusercontent.com/11330558/95928052-48e16d80-0d75-11eb-8cf3-24e49bae29f1.png">

If there is no input when an option is selected, it throws the following error:

<img width="529" alt="Screen Shot 2020-10-13 at 4 51 31 PM" src="https://user-images.githubusercontent.com/11330558/95928063-4e3eb800-0d75-11eb-8941-c4cca3c0685e.png">

If there is an input value that matches one of the options, the option is selected ('Public Speaking' in the screenshot below), but the create modal is still shown:

<img width="800" alt="Screen Shot 2020-10-13 at 5 06 41 PM" src="https://user-images.githubusercontent.com/11330558/95928568-87c3f300-0d76-11eb-84a0-daf932ada104.png">

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #1913, opened 2020-10-17
fix: #1909 autosuggest behavior on keydown

udayanshevade:
# Description

Fixes 2 issues with InputTag behavior inside a parent component, e.g. in this case QuickCreate:

1. The default value of `autocompleteLabel` has been [reverted](https://github.com/ifmeorg/ifme/commit/4a31511966790a844d96a65a7f4d1021804f1c24#diff-009b07c356e1317043bd5c8c0ff3e0c76c03144bce19a23fc8a514afdb27f1eaL30) to an empty string to avoid an error
2. `onKeyDown` is now being prevented when `onSuggestionSelected` fires, i.e. when a suggestion from the list is selected.

## More Details

1. The `undefined` value of `autocompleteLabel` in InputTag would [throw an error](https://github.com/ifmeorg/ifme/blob/main/client/app/widgets/QuickCreate/index.jsx#L51) when passed into the `onChange` prop on 'Enter' keydown: >Cannot read property 'toLowerCase' of undefined. The helper `labelExists` expects the value to be a string.

2. Recently, the deprecated `onKeyPress` event handler was renamed to use `onKeyDown` instead. But that caused an unexpected issue. It gets called [alongside](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Input/InputTag.jsx#L105) `onSuggestionSelected` from Autosuggest, which also derives from `onKeyDown`.
Previously `onKeyPress` would be ignored any time `onSuggestionSelected` was fired. So this fix re-includes that expectation with `onKeyDown`.

(New tests have been added to validate the behavior.)

## Corresponding Issue

#1909 

## Screenshots

Showing three expected behaviors of QuickCreate/InputTag:

after_fix

- Selecting a highlighted list option on 'Enter' keydown when there is no input, e.g. 'Test Category'
- Selecting a matching list option with 'Enter' keydown, e.g. 'Public Speaking'
- Opening the create form modal on 'Enter' keydown when there is no match, e.g. 'New Category'

_(Apologies for the tiny gif made for ants!)_

If anything looks missing or out of place, please LMK and I'll make the changes.

Thanks! :coffee:

---

### DIFF
--- a/client/app/components/Input/__tests__/InputTag.spec.jsx
+++ b/client/app/components/Input/__tests__/InputTag.spec.jsx
@@ -3,24 +3,43 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { InputMocks } from 'mocks/InputMocks';
 
-const component = InputMocks.createInput(InputMocks.inputTagProps);
-const { checkboxes } = InputMocks.inputTagProps;
+const { inputTagProps, createInput } = InputMocks;
+const getComponent = (extraProps = {}) => createInput(inputTagProps, extraProps);
+// baseline for most tests
+const component = getComponent();
+const { checkboxes } = inputTagProps;
 
 describe('InputTag', () => {
-  it('type in value and select it', () => {
-    const [, { label: checkboxLabelTwo }] = checkboxes;
+  it('renders input correctly', () => {
     render(component);
 
     const combobox = screen.getByRole('combobox');
     const input = screen.getByRole('textbox');
+    const listbox = screen.getByRole('listbox');
 
     expect(combobox).toBeInTheDocument();
     expect(input).toBeInTheDocument();
+    expect(listbox).toBeInTheDocument();
+  });
+
+  it('toggles menu options correctly', () => {
+    render(component);
+    const input = screen.getByRole('textbox');
+    expect(screen.queryByRole('option')).not.toBeInTheDocument();
+    // toggle options by focusing on input
+    userEvent.click(input);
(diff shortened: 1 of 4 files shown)


client/app/components/Input/__tests__/InputTag.spec.jsx
```diff
@@ -3,24 +3,43 @@ import { render, screen } from '@testing-library/react';
...
+    render(component);
+    const input = screen.getByRole('textbox');
+    expect(screen.queryByRole('option')).not.toBeInTheDocument();
+    // toggle options by focusing on input
+    userEvent.click(input);
+    expect(screen.getAllByRole('option')).toHaveLength(2);
+  });
+
+  it('type in value and select it', () => {
+    const [, { label: labelTwo }] = checkboxes;
```
julianguyen, line 35: Can this just be `[{ label: labelTwo }]`?

julianguyen (approved):
Thanks for catching this and fixing it! 🎉  Awesome work as usual :)

client/app/components/Input/__tests__/InputTag.spec.jsx
udayanshevade: For sure! I used the second checkbox label as the rendered mock component in these tests already selects the first checkbox value by default. But now that I look at it... the next test block overlaps with this one.

I'll try to separate those expectations, or maybe I can just remove this new test ("type in value and select it"), since the expectation is already essentially covered by the next block 👍

julianguyen (approved):

PR #1918, opened 2020-10-18
migrate Modal tests to RTL

eric199wen:
# Description

Migrate Modal component tests from Enzyme to RTL.
## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1794

---

### DIFF
--- a/client/app/components/Modal/__tests__/Modal.spec.jsx
+++ b/client/app/components/Modal/__tests__/Modal.spec.jsx
@@ -1,6 +1,9 @@
 // @flow
-import { mount } from 'enzyme';
 import React from 'react';
+import {
+  fireEvent, render, screen, waitFor,
+} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import Modal from 'components/Modal';
 
 const title = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit';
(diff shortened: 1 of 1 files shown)


client/app/components/Modal/__tests__/Modal.spec.jsx
```diff
@@ -39,19 +42,24 @@ describe('Modal', () => {
     jest.spyOn(window, 'alert');
   });
 
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
```
eric199wen, line 48: Seems like before we don't restore the mocks so `toHaveBeenCalled` was not actually a useful assertion.

client/app/components/Modal/__tests__/Modal.spec.jsx
```diff
@@ -304,11 +332,11 @@ describe('Modal', () => {
           />
         );
 
-        const wrapper = mount(component);
-        expect(wrapper.find('.avatar').exists()).toEqual(true);
-        wrapper.find('.avatar').simulate('click');
-        expect(wrapper.find('.modalBackdrop').exists()).toEqual(true);
-        expect(wrapper.find('.modal').exists()).toEqual(true);
+        const { container } = render(component);
+        expect(container.querySelector('.avatar')).toBeInTheDocument();
```
udayanshevade, line 336: With testing-library, it's preferable to query the rendered image itself, but in this case I'm having trouble querying the image because the Avatar is also wrapped inside the LazyLoad component.

I tried using `findByRole` which [combines](https://testing-library.com/docs/dom-testing-library/api-queries#findby) `waitFor` and `getByRole`:

```
const image = await screen.findByRole('image', {}, { timeout: 4000 });
expect(image).toBeInTheDocument();
```

Which should work afaik (and I tried it with a longer timeout too), but it wasn't cooperating. Maybe this test can be revisited 👍

client/app/components/Modal/__tests__/Modal.spec.jsx
```diff
@@ -373,13 +404,13 @@ describe('Modal', () => {
...
-        expect(wrapper.find('.modalBackdrop').exists()).toEqual(true);
-        expect(wrapper.find('.modal').exists()).toEqual(true);
-        wrapper.find('.modal').simulate('mouseover');
-        wrapper.find('.modalBackdrop').simulate('click');
-        expect(wrapper.find('.modalBackdrop').exists()).toEqual(true);
-        expect(wrapper.find('.modal').exists()).toEqual(true);
+        const { container } = render(component);
+        expect(container.querySelector('.modalBackdrop')).toBeInTheDocument();
+        expect(container.querySelector('.modal')).toBeInTheDocument();
+
```
udayanshevade, line 410: Likewise, there should be a `hover` event here to simulate the `mouseover` as before. In the component, the mouseover sets the `modalHasFocus` state to `true` and prevents the modal from being closed on click.

client/app/components/Modal/__tests__/Modal.spec.jsx
```diff
@@ -373,13 +404,13 @@ describe('Modal', () => {
...
+        userEvent.click(container.querySelector('.modal'));
```
udayanshevade, line 411: And the click here should be on the '.modalBackdrop' element

udayanshevade (changes requested):
Looks great @eric199wen! :tada: just a few comments below. Once those are addressed I think it should be good to go 👍

Issue #1920, opened 2020-10-19
Lock Account After Failed Password Attempts


bennpham:
# Description

Spun off after merging https://github.com/ifmeorg/ifme/pull/1895. [Current NIST guidelines](https://pages.nist.gov/800-63-3/sp800-63b.html#throttle) recommend no more than 100 failed login attempts as an upper bound before locking accounts.

Reading through [stackexchange account lockout policy guidelines](https://security.stackexchange.com/a/158084), 100 attempts can be a pretty high bound. I personally prefer 10 since I feel like its common for a user to incorrectly type their passwords multiple times. The balance here is between usability and security so if we set the rate too low, it's not very usable. 

I think lockout time of up to an hour would be good from the example NIST standards provided. The default settings in `config/intializers/devise.rb` if you enable it is 30 minutes.

## Additional Notes

You should be able to configure this in `config/initializers/devise.rb`

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #1923, opened 2020-10-19
[Issue #1920] Devise configuration for locking account after failed password attempts

camilacampos:
# Description

As stated at #1920, [current NIST guidelines](https://pages.nist.gov/800-63-3/sp800-63b.html#throttle) recommend no more than 100 failed login attempts as an upper bound before locking accounts. This PR implements devise lockable configuration for users.

## More Details

The values I used here are totally discussable, but I stuck mostly to default options.
### Lock Strategy
**Failed Attempts:** After 20 failed attempts, user will be locked (20 is the default value)

### Unlock Strategy
**Time:** If user is locked, he/she can try to log-in again after 1 hour (1 hour is the default)

_PS. We could also have an option to unlock by email, but I'm not sure if we need to configure something to set it up_

## Corresponding Issue

Fixes https://github.com/ifmeorg/ifme/issues/1920

# Screenshots

Screen Recording 2020-10-19 at 20 49 40

---

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -206,7 +206,7 @@ Devise.setup do |config|
   # Defines which strategy will be used to lock an account.
   # :failed_attempts = Locks an account after a number of failed attempts to sign in.
   # :none            = No lock strategy. You should handle locking by yourself.
-  # config.lock_strategy = :failed_attempts
+  config.lock_strategy = :failed_attempts
 
   # Defines which key will be used when locking and unlocking an account
   # config.unlock_keys = [ :email ]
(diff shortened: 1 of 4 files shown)


config/initializers/devise.rb
```diff
@@ -216,17 +216,17 @@
...
-  # config.maximum_attempts = 20
+  config.maximum_attempts = 20
 
   # Time interval to unlock the account if :time is enabled as unlock_strategy.
-  # config.unlock_in = 1.hour
+  config.unlock_in = 1.hour
 
   # Warn on the last attempt before the account is locked.
-  # config.last_attempt_warning = false
+  config.last_attempt_warning = false
```
bennpham, line 229: I think I like to warn users before the last failed attempt to give them a heads up. My personal opinion. @julianguyen do you prefer this on or off?

bennpham (approved):
This looks great. @julianguyen do you have any opinions on any of the current settings you'd prefer? 

> PS. We could also have an option to unlock by email, but I'm not sure if we need to configure something to set it up

I think we can write up another issue as a feature for this one too! We'd need a new migration to include an `unlock_token` `string` for users then setup Devise Mailer to send users the email.

To test that locally, I'd install the gem [letter_opener](https://github.com/ryanb/letter_opener) or [letter_opener_web](https://github.com/fgrehm/letter_opener_web) on `development` to test emails sent which I see that we already have. 

For mails, there should be some mailers in `app/mailers` to send users emails. Usually I see these setup with some external services or something that handles the emailing or something such as Mailchimp, Mailtrap, Mailgun, Sendgrind or something of that sort.

julianguyen (approved):

PR #1925, opened 2020-10-22
test: fix act warnings in Jest tests

### LINKED ISSUE #1857: Fix act warnings in Jest tests
# Description

If you run `cd client && yarn test`, you'll notice in the output the following warnings in various tests:

```
When testing, code that causes React state updates should be wrapped into act(...):

      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
```

Even though this doesn't block our tests from running and passing, we should fix our tests so that we use `act()` correctly.

Here's recommended [reading](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning) on how to address this issue with React Testing Library.

This warning is appearing in:
* `client/app/widgets/Comments/__tests__/Comments.spec.jsx`
* `client/app/components/BaseContainer/__tests__/index.spec.jsx`

`client/app/widgets/Resources/__tests__/Resources.spec.jsx` is an example, among some others, where `act()` is used correctly :)

---

meanmachin3:
# Description
Fixed act warnings in Jest tests

## More Details
While running `cd client` && `yarn test`, you'll notice in the output the following warnings in various tests:

When testing, code that causes React state updates should be wrapped into act(...):

      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */

In this pull, my attempt is to remove that warning by wrapping react state updates in `act(() => {}`

## Corresponding Issue
Fixes #1857

---

### DIFF
--- a/client/app/components/BaseContainer/index.jsx
+++ b/client/app/components/BaseContainer/index.jsx
@@ -2,6 +2,7 @@
 import React, { useState } from 'react';
 import axios from 'axios';
 import { Utils } from 'utils';
+import { act } from 'react-dom/test-utils';
 import StoryContainer from 'components/BaseContainer/StoryContainer';
 import { LoadMoreButton } from 'components/LoadMoreButton';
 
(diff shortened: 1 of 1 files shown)


client/app/components/BaseContainer/index.jsx
```diff
@@ -37,9 +38,11 @@ export const BaseContainerComponent = ({
     Utils.setCsrfToken();
     axios.get(url).then((response: any) => {
       if (response.data) {
-        setlastPage(response.data.lastPage);
-        setpage(page + 1);
-        setdata(data.concat(response.data.data));
+        act(() => {
```
julianguyen, line 41: You'll want to put an `act()` in the test file, not the component itself.

PR #1926, opened 2020-10-22
Issue-1827 refactor medications_controller_spec to requests spec

charlesdyan:
# Description

Refactoring spec/controllers/medications_controller_spec.rb -> spec/requests/medictions_spec.rb

## Corresponding Issue

[issue 1827](https://github.com/ifmeorg/ifme/issues/1827)
---


spec/requests/medications_spec.rb
```diff
@@ -1,87 +1,27 @@
 # frozen_string_literal: true
-describe MedicationsController do
-  describe '#print_reminders' do
```
charlesdyan, line 3: Moved to bottom to be consistent with strategies_spec. Being totally new to ruby/rails and rspec, would be curious to understand how a single-method unit test like this fits into request spec model.

spec/requests/medications_spec.rb
```diff
@@ -1,87 +1,27 @@
...
     let!(:medication) { create(:medication, user: user) }
 
     context 'when signed in' do
       before { sign_in user }
 
-      it 'renders index page' do
-        get :index
-        expect(response).to render_template(:index)
+      it 'returns a successful response' do
+        get medications_path
```
charlesdyan, line 11: where are these path variables defined? I see them used throughout the other request spec migrations, but can't find where in the project they're actually defined or what their values should be

spec/requests/medications_spec.rb
```diff
@@ -144,7 +95,7 @@ def post_create(medication_params)
           it 'has no validation errors and has correct refill' do
             post_create valid_medication_params
             expect(assigns(:medication).errors).to be_empty
-            expect(assigns(:medication)[:refill].strftime('%d/%m/%Y')).to eq('01/01/2020')
+            expect(response.body).to include(valid_medication_params.refill)
```
charlesdyan, line 98: Is this sufficient by itself (i.e. if we just drop the line above to be rid of `assigns`)? Is there a more complete expected body we could use with `eq` here instead that captures the lack of validation errors?

spec/requests/medications_spec.rb
```diff
@@ -144,7 +95,7 @@ def post_create(medication_params)
...
           end
         end
```
charlesdyan, line 101: At a bit of a loss on this one. Have several thoughts, none of which seems obviously best to me:

1. calculate from and to dates using `valid_medication_params` field values. Generate list of dates spanning the two dates. Check that at least one date on the list is included in the `response.body`.

2. make it a json request. Parse body for medication refill date. Calculate range as in 1. and check that refill falls in that range.

3. scrap all expectations. say responsibility for testing refill auto-gen lies with some method-level unit test instead. Here it is sufficient that the response successfully redirects to `medication_path(valid_medication_params.name`

4. something else entirely

spec/requests/medications_spec.rb
placy2: I would recommend taking a look at [this page in the Rails docs](https://guides.rubyonrails.org/routing.html) for information about how this routing is set up and used. A big part of Rails relies on naming and structure conventions to perform *\"magic\"* supported by the framework like this.

tyler-wel: 100% magic

spec/requests/medications_spec.rb
tyler-wel: Technically ya. The "it" is specifically saying there is no validation errors, which that second expectation isn't. I'd say you're fine just removing

tyler-wel: On a related note:

https://www.betterspecs.org/#single

spec/requests/medications_spec.rb
tyler-wel: See my other comment.

I'd say maybe 3. I'm on mobile and haven't checked the controller code, but the responsibilites of the request spec should be to ensure the request passes and/or fails. It can be argued it would be good to check the output of  requests, especially for json requests (I'm in favor of having output checkes for json)

My belief is to think of these tests like black boxes. Check the in, and check the out. Ensure it succeeded or failed with the correct output. That's another reason we're removing all the `assigns` checks. That and those matchers are part of a gem that's being removed.

cc @bennpham

spec/requests/medications_spec.rb
bennpham: TLDR answer: It doesn't. I'd delete it.

---

They're duplicate of https://github.com/ifmeorg/ifme/blob/60f8c9b067a9d77d9e1681d954b3346227e47ec1/spec/helpers/reminder_helper_spec.rb#L44-L84

so they're redundant.

----

I actually was confused initially by `#print_reminders` since it was found in `reminder_helper.rb` and was included in the controller via `include ReminderHelper`. 

Looking at the available routes for medications:

image

I don't see `#print_reminders` and they're already and should be handled in `reminder_helper_spec.rb` itself. 

Offtopic to the above, usually I see concerns included in models and controllers more than helpers, but I suppose sometimes helpers are occasionally included too, but I see them used for views more.

Source: https://stackoverflow.com/a/18751095/9463774

spec/requests/medications_spec.rb
```diff
@@ -368,4 +319,45 @@ def put_update(medication_params)
...
+      let(:medication) { create(:medication, :with_both_reminders, user_id: user.id) }
+
+      it is_expected.to eq(
+        '<div>' \
+        '<i class="fa fa-bell smallMarginRight"></i>' \
+        'Refill reminder email, Daily reminder email</div>'
+      )
+    end
+  end
+end
```
bennpham, line 363: Nitpick: I think we should have a newline at end of file?

spec/requests/medications_spec.rb
bennpham: Looking at the controller itself, wouldn't response be something like

```ruby
expect(response).to have_http_status(302)
```

Maybe we can replace the assigns with that or remove the assigns and put that assertion. If we remove the assigns, let's just called the test:

```ruby
it 'has correct refill' do
```

otherwise with the redirect check

```ruby
it 'redirects to medication and has correct refill' do
```

----

If it did have an error, wouldn't it be `have_http_status(422)` for unprocessable entity?

I'm basing on the condition below in the **create** method:

```ruby
if @medication.save
else
  render_unprocessable_medication
end
```

bennpham:
Not ready for merge yet. I think the following would need to be done first: 
- delete `#print_reminders` 
- finish up `#destroy` (path should be `delete medication_path(medication)`)
- clean up the `assigns` in `#update` and `#create`

spec/requests/medications_spec.rb
```diff
@@ -153,51 +138,39 @@ def post_create(medication_params)
...
 
     def put_update(medication_params)
-      put :update, params: { id: medication.id, medication: medication_params }
+      put medication_path(medication), params: { medication: medication_params }
     end
 
-    context 'when the user is logged in' do
-      include_context :logged_in_user
+    context 'when signed in' do
+      before { sign_in user }
```
charlesdyan: not sure if there's a functional difference between these two, but opting for consistency across the spec

spec/requests/medications_spec.rb
```diff
@@ -153,51 +138,39 @@ def post_create(medication_params)
...
 
       context 'when valid params are supplied' do
+
         it 'updates a medication' do
           put_update valid_medication_params
-          expect(assigns(:medication)[:name]).to eq(valid_medication_params[:name])
+          expect(response.body).to include(valid_medication_params[:name])
```
charlesdyan, line 151: This test currently doesn't pass because the updated medication name appears in the response body reformatted from `Updated Medication Name` to `updated-medication-name` (in redirect url). I assume this is a transformation applied by FriendlyId, but I'm not sure how to mirror that transformation here.

Alternatively, I could either drop this example, or change the redirect example below to something like
```suggestion
          it 'redirects to renamed medication page' do
            let!(renamed_medication) { 
              create(:medication, user_id: user.id, name: valid_medication_params[:name])
            }
            put_update valid_medication_params
            expect(response).to redirect_to(medication_path(renamed_medication))
          end
```

spec/requests/medications_spec.rb
```diff
@@ -52,17 +57,17 @@
...
           expect(response).to be_successful
         end
       end
 
       context 'when the medication does not exist' do
-        it 'redirects to the medications page'
+        it 'redirects to the medications page' do
           get medication_path(medication.id + 1)
-          expect(response).to redirect_to(medications_path)
+          expect(response).to have_http_status(302)
```
charlesdyan, line 70: is one of these preferable to the other? Seems to me that `redirect_to` tests output more thoroughly where it is possible to use it.

spec/requests/medications_spec.rb
```diff
@@ -153,51 +138,39 @@ def post_create(medication_params)
...
-        end
-
-        it 'adds errors to the medication ivar' do
-          expect(assigns(:medication).errors).not_to be_empty
+        let(:invalid_medication_params) { 
+          valid_medication_params.merge(name: nil, dosage: nil)
+        }
+        it 'returns 422, unprocessable entity' do
+          put_update invalid_medication_params
+          expect(response).to have_http_status(422)
```
charlesdyan, line 173: Currently receiving 200 response instead of 422 here and in twin example in #create. Investigating why...

charlesdyan:
Thanks for the eyes, y'all! Learning a lot =)

spec/requests/medications_spec.rb
```diff
@@ -13,9 +14,13 @@
       end
 
       context 'when request type is JSON' do
-        before { get medications_path, params: { page: 1, id: medication.id }, format: :json }
+        before { get medications_path,
+          params: { page: 1, id: medication.id },
+          headers: { "ACCEPT" => "application/json" }
+        }
```
tyler-wel, line 20: https://github.com/rubocop-hq/ruby-style-guide#single-line-blocks-delimiters
```suggestion
        before do 
          get medications_path,
          params: { page: 1, id: medication.id },
          headers: { "ACCEPT" => "application/json" }
        end
```

spec/requests/medications_spec.rb
```diff
@@ -91,39 +96,19 @@ def post_create(medication_params)
...
-
-        it 're-renders the new medication form' do
-          expect(response).to render_template(:new)
-        end
-
-        it 'adds errors to the medication ivar' do
-          expect(assigns(:medication).errors).not_to be_empty
+        let(:invalid_medication_params) {
+          valid_medication_params.merge(name: nil, dosage: nil)
+        }
```
tyler-wel, line 108: https://github.com/rubocop-hq/ruby-style-guide#single-line-blocks-delimiters
```suggestion
        let(:invalid_medication_params) do
          valid_medication_params.merge(name: nil, dosage: nil)
        end
```

spec/requests/medications_spec.rb
```diff
@@ -153,51 +138,39 @@ def post_create(medication_params)
...
-
-        it 're-renders the new medication form' do
-          expect(response).to render_template(:new)
-        end
-
-        it 'adds errors to the medication ivar' do
-          expect(assigns(:medication).errors).not_to be_empty
+        let(:invalid_medication_params) { 
+          valid_medication_params.merge(name: nil, dosage: nil)
+        }
```
tyler-wel, line 170: https://github.com/rubocop-hq/ruby-style-guide#single-line-blocks-delimiters
```suggestion
        let(:invalid_medication_params) do
          valid_medication_params.merge(name: nil, dosage: nil)
        end
```

spec/requests/medications_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+
+    context 'when signed in' do
+      before { sign_in user }
+
+      context 'when valid params are supplied' do
+
+        it 'updates a medication' do
+          updated_name_slug = valid_medication_params[:name]
+            .downcase
+            .gsub(' ', '-')
```
charlesdyan, line 164: this feels hacky

charlesdyan:
Think this should be just about there now. @tyler-wel or @bennpham would be dope if you could give this a final (hopefully) pass when you've got the time =) Thanks again for all the tips!

spec/requests/medications_spec.rb
bennpham: I'd go with ```before { sign_in user }``` 👍

spec/requests/medications_spec.rb
bennpham: You can use [parameterize](https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-parameterize) to "slugify" the title.

So the following would result:

```ruby
include ActiveSupport::Inflector
parameterize("Updated Medication Name")

=> "updated-medication-name"
```

spec/requests/medications_spec.rb
bennpham:
```suggestion
          updated_name_slug = (valid_medication_params[:name]).parameterize
```

spec/requests/medications_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+          put_update valid_medication_params
+          expect(response.body).to include(updated_name_slug)
+        end
+
+        it 'redirects to medication' do
+          put_update valid_medication_params
+          expect(response).to have_http_status(302)
```
bennpham, line 171: I think `redirect_to` seems to be the better case if the test reads `redirects to medication`, then I'd probably expect it to be the medication page.

spec/requests/medications_spec.rb
bennpham: I think if I were reading a test and it says `redirects to the medication page`, I'd probably expect it to use `redirect_to` whereas if the test reads `redirects the user` then I'd just use a generic 302.

Reading rspec tests is kind of like reading a sentence in English so I'd probably expect the test to do what it reads to do.

bennpham (approved):
A few nitpicks on my comments otherwise I approve the rest 👍 . Looks good.

Issue #1893, opened 2020-10-09
RSpec test runner resets sessions after every example


SherSpock:
# Description
Currently the RSpec test runner resets sessions after every example.
This behavior is currently configured in the `spec_helper`:

```ruby
config.after(:each) do
    Capybara.reset_sessions!
    Capybara.use_default_driver
    Capybara.app_host = nil
end
```

# Impact
This would appear to have minimally the following effects:

## Performance
Without benchmarking it's unclear how much performance overhead is involved here, but it seems reasonable to expect that it is measurable but marginal.

## Maintainability
Resetting sessions everywhere is convenient in that we don't need to figure out where the resets are actually needed, and then invoke this behavior only in those specific cases. 

It is old code, added somewhat mysteriously in de57f77cc4e9f912ca8ffb25e5bcd317d598c6b6/#332. Being old, important in terms of its location, and of not-readily-discernible intent, it is likely by its presence to invite confusion and potentially worse.

# Desired Functionality
The test runner only resets sessions where necessary to ensure the nominal and non-flaky behavior of the test suite.

# What has already been done
* Chatted about this in #dev. 
* Commented out the config block and ran the test suite, but the results were non-deterministic. Mainly because my local run had failures both before and after, and I'm not certain if the number remained the same. There were 8 failures on the second run.

PR #1934, opened 2020-10-28
Issue-1893 only reset session for feature tests

charlesdyan:
# Description

Modified test config in spec_helper.rb to only reset session on Spec tagged `type: :feature`. 

## More Details

While specs using feature/scenario are all automatically tagged type: :feature for the purpose of running or excluding tagged tests, i.e. with `$ rspec --tag type:feature`, simply modifying spec_helper.rb was insufficient and broke several tests. I suspect the case may be that there is some distinction between tags and spec metadata, something to the effect that all spec metadata become tags but not all tags become spec metadata. Accordingly, explicitly adding metadata `type: :feature` to feature specs got session resets properly running after each.

## Corresponding Issue

[Issue 1893](https://github.com/ifmeorg/ifme/issues/1893)

---


SherSpock (commented):
Just my quick two cents: I think this is fantastic -- very targeted and clean and addresses the issue.

The only thing I might suggest after glancing at it and the RSpec feature docs is potentially altering the config rather than specifying the feature your in every spec: `config.infer_spec_type_from_file_location!`

Assuming that does the trick, of course!

julianguyen (commented):
I like @SherSpock's recommendation! Overall, thank you for working on this! :D

charlesdyan:
Thanks for the look @julianguyen! Initially, I didn't make the change @SherSpock recommended because of a followup discussion we had on Slack. Looking into it a bit more I've noticed that `config.infer_spec_type_from_file_location!` is actually already in the config (`spec_helper.rb:90` on my branch). Despite that, however it is assigning metadata seems not to be picked up by the filter block I added to reset sessions for feature tests. That is, before explicitly adding the metadata to these specs, many would fail in the same manner they did when I initially disabled session resets to identify impact.

That said, I'm not sure how best to proceed here... I may try to investigate deeper to figure out why the inferred metadata isn't picked up by the filter, but I haven't found anything illuminating yet.

SherSpock:
> Thanks for the look @julianguyen! Initially, I didn't make the change @SherSpock recommended because of a followup discussion we had on Slack. Looking into it a bit more I've noticed that `config.infer_spec_type_from_file_location!` is actually already in the config (`spec_helper.rb:90` on my branch). Despite that, however it is assigning metadata seems not to be picked up by the filter block I added to reset sessions for feature tests. That is, before explicitly adding the metadata to these specs, many would fail in the same manner they did when I initially disabled session resets to identify impact.

@charlesdyan 
Very interesting. I wonder at what point in the RSpec lifecycle the `infer_spec_type_from_file_location!` call is happening, where and how the after each with context of type feature is happening, and where and how the feature tag assignment by inference is being done. That could be checked through static analysis or dynamic exploration -- whichever you prefer :) Perhaps on the House-ian "diagnose-by-treatment" side of things, we could see a meaningful change in behavior by merely changing where that call is happening, or by changing the after each -- I don't know if contexts inherit the spec metadata (although I would assume they do), or if all the specs/examples we want to have the feature type metadata even have contexts. Maybe I sound foolish here and am making some obvious conceptual mistakes, I just am not an old hand at RSpec and it's been quite some time since I last worked with it before ifme. 

I also think, as another option, it would be totally reasonable to leave things as they are, but turn *off* feature tag inference in the `spec_helper` since its presence is doing nothing for us and we're assigning the feature tags manually anyway. Leaving the app in its current state is a recipe for confusion -- my knee-jerk is to remove the noop magic if we can't figure out a way to make it "op" to our needs, but that's certainly not the only way.

@julianguyen 
Please feel free to add or amend anything!

SherSpock:
For some transparency and completeness of relevant info in one place, surfacing the relevant part of the referenced Slack conversation:

```
@Charley Yan The more I think about this the more I like it the way you have it already. Tagging each spec is more noise for the PR, but if we were to instead infer the feature tag from each feature spec's presence in the features directory, then we would have two ways tags are getting specified in the project -- one magic and one explicit. Avoiding magic is usually best, as is minimizing the number of ways of doing something, as is avoiding going against the grain in an ecosystem. Magic is widespread in Rails and allowing lots of ways to do something is widespread in Ruby, but it seems to me that it's better to go against the ecosystem grain here in deference to the broader best practice
In short, pending other input, I'll cast my vote as I like it how you have it.
```

julianguyen (approved):

Issue #1921, opened 2020-10-19
Setup Recaptcha After A Certain Amount of Failed Login Attempts


bennpham:
# Description

Spunned off after merging https://github.com/ifmeorg/ifme/pull/1895. It'd probably be a nice feature to implement recaptcha after a certain amount of failed logins. I think 3 failed attempts would be a good time to ask user to correctly answer Recaptcha to make sure they're not a bot. 

## Additional Notes

Gem for Recaptcha: https://github.com/ambethia/recaptcha

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

seannemann21:
Hello, I started working on a PR, and I was wondering if you could provide any feedback as to whether this is what you had in mind? https://github.com/ifmeorg/ifme/pull/1936. I still need to add tests and clean up a few things, but the logic is mostly there. It was based off of https://github.com/heartcombo/devise/wiki/How-To:-Use-Recaptcha-with-Devise, but I had to add logic for the 3 failed attempts. One concern I have is if a user fails to login 3 times, prompting the recaptcha, then leaves before successfully logging in. The next time they come to the site, they won't see the recaptcha until they attempt to login again and fail (due to the request not having a recaptcha). Now, the form will reload with the recaptcha, but it's kinda an awkward experience since they'll have to reenter their username/password. This is probably a pretty rare scenario, though, so I'm not sure how much it matters

bennpham:
> The next time they come to the site, they won't see the recaptcha until they attempt to login again since failed attempts are tied to emails, and we won't know their email until they attempt to login.

I like the cookie idea for this.

> Now, the form will reload with the recaptcha once they try to login with that email once, but it's kinda an awkward experience since they'll have to reenter their email/password

I'd prefer not to make the user experience discomforting for this matter especially since we eventually have 20 fails to lock. I think that'd be good enough. I'd rather not punish the user unfairly from a gimmick as long as other procedures are in place. Sometimes misspelling happens a lot multiple times, especially if you have a very long sentence for a password.

----

[...]

PR #1936, opened 2020-11-08
Require recaptcha after 3 failed logins

seannemann21:
# Description

This PR adds a recaptcha requirement if a user has failed to login 3 times. Failed attempts are tracked by Devise on a per user basis, so 3 failed logins with the same email will trigger the recaptcha requirement but 3 failed logins with different usernames will not.

## More Details

This PR adds the [ambethia/recaptcha](https://github.com/ambethia/recaptcha) gem for recaptchas. This gem was suggested in both the issue and in the [Devise Wiki](https://github.com/heartcombo/devise/wiki/How-To:-Use-Recaptcha-with-Devise). The implementation is based off of the Devise [example for logins](https://github.com/heartcombo/devise/wiki/How-To:-Use-Recaptcha-with-Devise#devisesessionscontroller). However, additional logic was required since we only want to show the recaptcha after 3 failed logins. A cookie is given to users who needed a recaptcha after their last failed login in case the user leaves the site before logging in and then comes back. Additionally, the recaptcha only display/is required if the recaptcha site key/secret key are set. This should ease local development.

This PR uses recaptcha V2, which requires the user to click an "I am not a robot" check box or solve a "challenge" by selecting images that contain a certain item. The frequency with which the image challenges appear can be adjusted within the recaptcha admin console.

Note:  We will need to set the site key and secret key ENV variables in prod. This is my first PR here, so I could probably use some help. This change could still be merged without them, we just won't show/require the recaptcha until then.

If you are reviewing this PR, leave a comment here and I will grant you access to the admin console for our recaptcha. Once you have been granted access, you can go to https://www.google.com/recaptcha/admin to view the keys.

## Validation Steps
It's up to you if you actually want to run through these steps. I'm not positive what the process for reviewing PRs consists of, but here's how you can check out the changes locally...

Once given access to recaptcha site/secret keys (comment and I'll add you)...
1. In `config/env/development.env` Set `RECAPTCHA_SITE_KEY` and `RECAPTCHA_SECRET_KEY` ENV variables
2. Try to sign in as a user with an incorrect password 3 times
3. Verify that after the 3rd failed attempt a recaptcha will appear
4. Before signing in, navigate away from the page, then come back to the sign in page. The recaptcha should still be there since you have a cookie set
5. Try to sign in with a different email. Since this other email doesn't require a recaptcha, the cookie will be cleared. Go to the sign in page and try to sign in with the original email that requires a recaptcha. The recaptcha field will not be present since the last user you attempted to sign in with did not require it, but if you try to sign in as the original user again, you'll be prompted to fill in a recaptcha
6. Ensure you can't submit the sign in form until the recaptcha has been completed
7. Sign in successfully

Might be a good idea to validate that the recaptcha doesn't appear/isn't required if the RECAPTCHA_SITE_KEY/RECAPTCHA_SECRET_KEY ENV variables aren't present

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1921

# Screenshots

Screen Shot 2020-11-27 at 6 02 37 PM

recaptcha-challenge

recaptcha-example

---


app/assets/javascripts/require_recaptcha.js
```diff
@@ -0,0 +1,7 @@
+function recaptcha_verified () {
+  document.querySelector('.recaptcha-submit').removeAttribute('disabled')
+}
+
+function recaptcha_expired () {
+  document.querySelector('.recaptcha-submit').setAttribute('disabled', 'disabled')
+}
```
seannemann21, line 7: I wasn't quite sure where to add this javascript for enabling/disabling the submit button based on the status of the recaptcha.

app/controllers/sessions_controller.rb
```diff
@@ -15,4 +22,36 @@ def set_user_locale
...
+
+    @display_recaptcha = true
+    cookies.permanent[:login_recaptcha] = true
+  end
+
+  def show_recaptcha?
+    RecaptchaService.recaptcha_configured? &&
+      (recaptcha_required_for_user? ||
+      params[:recaptcha] ||
+      cookies[:login_recaptcha])
```
seannemann21, line 46: This handles three different scenarios...
1. User enters incorrect password, in which case the the email is still in the request parameters, so we can use `recaptcha_required_for_user?` to look up the number of failed attempts
2. If a user tries to login without filling out the recaptcha (which shouldn't really happen since it's disabled), then we redirect to the `new` action with the `recaptcha` param
3. If a user needs a recaptcha, they'll have the `login_recaptcha` cookie set, so if they leave before successfully signing in, then when they return, we'll know to sure the recaptcha

Admittedly, there is probably a cleaner way to handle these scenarios, so I'm open to any suggestions!

app/controllers/sessions_controller.rb
```diff
@@ -15,4 +22,34 @@ def set_user_locale
...
+
+    @display_recaptcha = true
+    cookies.permanent[:login_recaptcha] = true
+  end
+
+  def show_recaptcha?
+    RecaptchaService.recaptcha_configured? &&
+      (cookies[:login_recaptcha] ||
+      params[:recaptcha] ||
+      recaptcha_required_for_user?)
```
seannemann21, line 46: This handles three different scenarios...
1. User enters incorrect password, in which case the the email is still in the request parameters, so we can use `recaptcha_required_for_user?` to look up the number of failed attempts
2. If a user tries to login without filling out the recaptcha (which shouldn't really happen since it's disabled), then we redirect to the `new` action with the `recaptcha` param
3. If a user needs a recaptcha, they'll have the `login_recaptcha` cookie set, so if they leave before successfully signing in, then when they return, we'll know to sure the recaptcha

Admittedly, there is probably a cleaner way to handle these scenarios, so I'm open to any suggestions!

app/views/devise/sessions/new.html.erb
```diff
@@ -34,7 +35,12 @@
         } %>
       </div>
     <% end -%>
-    <%= f.submit t('account.sign_in'), class: 'buttonGhostM marginTop' %>
+
+    <% if @display_recaptcha %>
+      <%= recaptcha_tags id: 'recaptcha-form', callback: 'recaptcha_verified', expired_callback: 'recaptcha_expired' %>
```
seannemann21, line 40: `recaptcha_verified` and `recaptcha_expired` are the names of the javascript functions that run when a recaptcha is completed/expires. Recaptchas expire after about 2 minutes.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    context 'when recaptcha is not configured' do
+      before do
+        allow(RecaptchaService).to receive(:recaptcha_configured?).and_return(false)
+      end
+
+      it 'does not display recaptcha' do
+        cookies[:login_recaptcha] = true
+        get new_user_session_path
+
+        expect(response.body).to_not include recaptcha_id
```
seannemann21, line 18: This doesn't feel great, but I wasn't quite sure how to verify that the recaptcha loaded on the page. I'm open to any suggestions!

app/controllers/sessions_controller.rb
```diff
@@ -1,11 +1,18 @@
...
+  before_action :set_recaptcha, only: [:new]
+
   def create
     super
     set_user_locale if user_signed_in?
   end
 
+  def new
+    super
+  end
```
seannemann21, line 14: Rubocop required we specify `new` since we have a `before_action` associated with it

app/assets/javascripts/require_recaptcha.js
```diff
@@ -0,0 +1,7 @@
+function recaptcha_verified () {
+  document.querySelector('.recaptcha_submit').removeAttribute('disabled')
+}
+
+function recaptcha_expired () {
+  document.querySelector('.recaptcha_submit').setAttribute('disabled', 'disabled')
+}
```
seannemann21, line 7: I wasn't quite sure where to add this javascript for enabling/disabling the submit button based on the status of the recaptcha.

app/views/devise/sessions/new.html.erb
```diff
@@ -34,7 +35,12 @@
         } %>
       </div>
     <% end -%>
-    <%= f.submit t('account.sign_in'), class: 'buttonGhostM marginTop' %>
+
+    <% if @display_recaptcha %>
+      <%= recaptcha_tags id: 'recaptcha_form', callback: 'recaptcha_verified', expired_callback: 'recaptcha_expired' %>
```
seannemann21, line 40: `recaptcha_verified` and `recaptcha_expired` are the names of the javascript functions that run when a recaptcha is completed/expires. Recaptchas expire after about 2 minutes.

app/controllers/sessions_controller.rb
bennpham: I don't think I got my environment setup to play around with Recaptcha on local, but from what I'm guessing, would this be the implemented scenario?

**Scenario 1: Incorrect email**

3 failed attempt, we get a cookie to enable a Recaptcha for that session on that specific computer/IP?

**Scenario 2: Correct email, incorrect password**

3 failed attempt, we attach a required Recaptcha to that specified account so no matter what computer someone uses, that account will now require Recaptcha

Case #3 would be an example where someone uses a different browser or refresh the website to not require Recaptcha again, right?

-----

I like that approach, we'll just need to add the tests for each use case. 

At first I was thinking probably just lock it after 3 for that IP, but I suppose we can handle by user basis as well since that is pretty consistent with the current Devise lock settings per user. 

I suppose if the or statements get too big, I suppose you could wrap it in `recaptcha_required?` private method or something to be more intuitive to read?

app/views/devise/sessions/new.html.erb
bennpham: I wonder if this could fit into one of the React components, but I'm still learning React at the moment so I don't have too much opinion on that at the moment. I think @julianguyen might have some better opinions on where this could possibly go?

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+    end 
+
+    context 'when recaptcha is configured' do
+      before do
+        allow(RecaptchaService).to receive(:recaptcha_configured?).and_return(true)
+      end
+
+      context 'and user has recaptcha cookie' do
```
bennpham, line 27:
```suggestion
      context 'when user has recaptcha cookie' do
```

Let's change these, line 36-37, 51-52, 60-61, 71, 112, 124, 136, 142 to say `when` instead. For those 2 nested context, let's just merge to 2 ands into one lines as well since we don't need to distinguish the other cases since we don't have other `it` blocks to separate the 2 cases. 

See https://www.betterspecs.org/#contexts for RSpec best practices.

spec/services/recaptcha_service_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+        site_key: site_key,
+        secret_key: secret_key,
+      )
+    }
+
+    before do
+      allow(Recaptcha).to receive(:configuration).and_return(configuration)
+    end
+
+    context 'site key and secrete key are present' do
```
bennpham, line 26:
```suggestion
    context 'when site key and secrete key are present' do
```

Also lines 33, 41, 51, 61

spec/requests/sessions_spec.rb
bennpham: Seems good enough for a request spec to me.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+          before do
+            get new_user_session_path, params: { recaptcha: true }
+          end
+
+          it 'displays recaptcha' do
+            expect(response.body).to include recaptcha_id
+          end
+
+          it 'sets recaptcha cookie' do
+            expect(cookies[:login_recaptcha]).to eq 'true'
```
bennpham, line 47: Why is `'true'` a string instead of a boolean?

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    let(:user) { create(:user, failed_attempts: 5, password: password) }
+
+    before do
+      allow(RecaptchaService).to receive(:recaptcha_configured?).and_return(true)
+    end
+
+    context 'when user does not require recaptcha' do
+      let(:failed_attempts) { 1 }
+      it 'does not check recaptcha' do
+        post "/users/sign_in", params: { user: { email: user.email, password: password } }
```
bennpham, line 104:
```suggestion
        post user_session_path, params: { user: { email: user.email, password: password } }
```

Let's use Rails routes prefixes instead. You can see what they are when you run `bundle exec rake routes`.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+          expect(response).to redirect_to new_user_session_path(recaptcha: true)
+        end
+      end
+
+      context 'and recaptcha is filled out' do
+        let(:entered_password) { password }
+
+        before do
+          allow_any_instance_of(SessionsController).to receive(:verify_recaptcha).and_return(true)
+          post "/users/sign_in", params: { user: { email: user.email, password: entered_password } }
```
bennpham, line 129:
```suggestion
          post user_session_path, params: { user: { email: user.email, password: entered_password } }
```

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+        end
+
+        it 'clears recaptcha cookie' do
+          expect(cookies[:login_recaptcha]).to eq ''
+        end
+
+        context 'and credentials are valid' do
+          it 'signs user in' do
+            expect(response).to redirect_to '/'
```
bennpham, line 138:
```suggestion
            expect(response).to redirect_to root_path
```

bennpham (changes requested):
I made some comments mainly in regards to better spec practices. I have a few questions about the scenario cases that the Recaptcha covers as well.

I haven't gotten to test the Recaptcha on my local machine since I haven't setup the ENV variables for it, but I like that the recaptcha doesn't appear or come into any conflict with my workflow when it's not setup. Thanks for providing a gif on the details!

Main changes I'd like to see is some spec clean up unless I miss any coverage for other parts.

app/controllers/sessions_controller.rb
seannemann21: Sorry, I don't think I explained that clearly. All failed attempts are based on username. Right now, there's no tracking of failed attempts on an IP basis. My thought was that a bot could change their IP to get around it, but I can try to add it in if you think it's worthwhile?

The scenarios are really all the same (user tries to log in with the same email with an incorrect password 3 times). I was just trying to explain why each condition in the `if` block was necessary. `params[:recaptcha]` handles the scenarios in which the user redirects to the login page from `redirect_to new_session_path(resource_name, recaptcha: true)` in the `check_recaptcha` up above.  `recaptcha_required_for_user?` handles when a user makes it past that check, but is redirected to the login page by Devise due to an invalid password. Then the cookie is for when the user leaves and comes back. If they cleared the cookie, then they'll get caught by `check_recaptcha` and the form will reload with the recaptcha. Let me know if that's not clear!

app/views/devise/sessions/new.html.erb
seannemann21: Potentially, but then I'm not sure if we could use the `recaptcha_tags` helper that's provided by the recaptcha gem. Although, I don't think it'd be too hard to recreate what that wrapper does.

spec/requests/sessions_spec.rb
seannemann21: I think it's because cookies are always stored as strings, so even though the cookie is being stored with a boolean (i.e. `cookies.permanent[:login_recaptcha] = true`), it's being converted to a string. Maybe I should use "1" rather than "true"?

seannemann21:
@bennpham thanks for the review! Sorry, I was traveling yesterday and was busier than expected today. I'm hoping to make the suggested changes tomorrow or Thursday. Let me know if you think I should add IP tracking in this PR. Thanks for all of the help so far!

spec/requests/sessions_spec.rb
bennpham: This is fine to me then. It got the point across. I just wanted to make sure 👍

app/controllers/sessions_controller.rb
bennpham: Ah ok, that makes sense. Good handling on the edge cases 👍

bennpham: I think handling IP would cause scope creep for this PR. I rather stick to what we already have since it's large enough. I think if we think it's worthwhile, we can make another issue if it sounds nice.

seannemann21:
Ok, I've updated the specs. I think it should be ready to review again! Also I've have the section I'll add to https://github.com/ifmeorg/ifme/wiki/Installation ready to go. I'm going to wait to add it until this PR is in, though.

----------------------------

#### Google reCAPTCHA

[Google reCAPTCHA v2](https://developers.google.com/recaptcha) is used to verify users are not bots. This is triggered after a certain number of failed logins for a user.

If you would like to use this feature in your local environment, please [register your own site](https://www.google.com/recaptcha/admin/create) to generate keys, and then update RECAPTCHA_SITE_KEY and RECAPTCHA_SECRET_KEY in config/env/development.env.

When registering your site, be sure to select reCAPTCHA v2/"I'm not a robot" Checkbox and list "localhost" as the domain.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+        it 'sets recaptcha cookie' do
+          expect(cookies[:login_recaptcha]).to eq 'true'
+        end
+      end
+
+      context 'when user is specified in sign in params' do
+        let(:failed_attempts) {}
+        let(:user) { create(:user, failed_attempts: failed_attempts) }
+
+        before { get new_user_session_path, params: { user: { email: user.email } } }
```
bennpham, line 54: I feel like the **act** part of [Arrange Act Assert](http://wiki.c2.com/?ArrangeActAssert) might be called before the Arrange looking through the rest of the tests below. I'm wondering if it actually tests the edge case where you setup the attempts or something.

spec/services/recaptcha_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        it { is_expected.to be true }
+      end
+
+      context 'when the user has more than 3 failed logins' do
+        let(:failed_attempts) { 4 }
+
+        it { is_expected.to be true }
+      end
+    end
+  end
```
bennpham, line 81: You can ignore this as this is just a personal opinion of mine: I suppose this does follow [Better Spec convention](https://www.betterspecs.org/#contexts) although there is a [topic of debate](https://stackoverflow.com/a/51182031/9463774) in regards to too many nested context blocks. I personally am not too much of a fan of it for these case and rather just have `it 'passes if user has more than 3 failed logins'` since it seems short enough to me and nesting seem overkill for single cases, but that's just a matter of opinion. In this case, I prefer to disagree with better specs, but again, just an opinion.

bennpham (commented):
Overall looks good, but I do have a few comments/questions just for the specs.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+        context 'when user does not require recaptcha' do
+          let(:failed_attempts) { 0 }
+
+          it 'does not display recaptcha' do
+            expect(response.body).to_not include recaptcha_id
+          end
+        end
+
+        context 'when user requires recaptcha' do
+          let(:failed_attempts) { 3 }
```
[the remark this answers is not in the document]
seannemann21: I'll add a constant to `RecaptchaService` and reference it here

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    let(:user) { create(:user, failed_attempts: failed_attempts, password: password) }
+
+    before do
+      allow(RecaptchaService).to receive(:recaptcha_configured?).and_return(true)
+    end
+
+    context 'when user does not require recaptcha' do
+      let(:failed_attempts) { 1 }
+      it 'does not check recaptcha' do
+        post user_session_path, params: { user: { email: user.email, password: password } }
```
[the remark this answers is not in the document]
seannemann21: Good catch, it might have gotten lost while I was refactoring these.

spec/services/recaptcha_service_spec.rb
seannemann21: Hmm.. yeah I have a tendency to go back and forth on nesting contexts, which I've noticed gets me into trouble from a consistency standpoint sometimes, so lately I've been embracing them. In this case, I see your point. Having it a bit more explicit makes it easier to follow what's happening. I also don't know how I feel about using `is_expected` rather than an explicit `expect`. I didn't even know it was a thing until reading through Better Specs. I think I prefer the explicit assertion, but that might just be because `is_expected` is new to me.

seannemann21: Oh wait, I just realized that `failed_attempts` is used when creating the user. Are you suggesting I create the user within the spec? Or update the value on the user within the spec? I actually might just leave this one as is.

bennpham: Whatever is the more explicit solution :) . I prefer to be more explicit as well.

seannemann21: Yeah, I'll just create a new user in each spec.

seannemann21: Ok, I swung the complete opposite way. Should be about as explicit as you can make it now 😂

bennpham (approved):
skimmed through the rest and looks good to me 👍

seannemann21:
> skimmed through the rest and looks good to me 👍 

Ok, now I just need to figure out how to update the production ENV variables before I deploy, but I can ask in the slack channel for help with that. Thank you so much for all of the help! This was my first open source contribution in a couple years (and I think only my second ever), so I really appreciate it!

app/assets/javascripts/require_recaptcha.js
julianguyen: I think this is the appropriate place to put this since `recaptcha_tags` is expecting a `callback` and `expired_ callback` function. Though it would be ideal to have this handled on the React side in [startup scripts](https://github.com/ifmeorg/ifme/tree/main/client/app/startup) since we want to use the Rails asset pipeline for JS as little as possible. We could explore doing this in another PR :)

app/views/devise/sessions/new.html.erb
julianguyen: Thanks for bringing this up and I have a similar sentiment. Yeah, I don't think it's worth the time trying to extract what's going on with the `recapthca_tags` helper so it can be used in React so I think what you have is fine :)

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        before do
+          get new_user_session_path, params: { recaptcha: true }
+        end
+
+        it 'displays recaptcha' do
+          expect(response.body).to include recaptcha_id
+        end
+
+        it 'sets recaptcha cookie' do
+          expect(cookies[:login_recaptcha]).to eq 'true'
```
julianguyen, line 46: Should this be boolean instead of a string?

spec/requests/sessions_spec.rb
julianguyen: This is a good point. Yeah I'm curious if these tests pass without the `before`.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+        context 'when user requires recaptcha' do
+          let(:failed_attempts) { RecaptchaService::MIN_ATTEMPTS_FOR_RECAPTCHA }
+
+          it 'displays recaptcha' do
+            expect(response.body).to include recaptcha_id
+          end
+
+          it 'sets recaptcha cookie' do
+            expect(cookies[:login_recaptcha]).to eq 'true'
```
julianguyen, line 72: Similar question as above, should this be a boolean?

julianguyen (approved):
Fantastic work! Thanks so much Sean for taking this on 🙌🏼 I learned a lot looking at your PR. 

So I'll take on setting up the production keys. Let me know what information we need to add.

spec/requests/sessions_spec.rb
seannemann21: Whoops, sorry, I had missed this comment before. To my knowledge, the `before` block will run in whatever context your specs run, so they will use the innermost nested `failed_attempts`/`user` value in these specs. Turns out, I could even delete `let(:failed_attempts) {}` up above and the specs would run just fine even though I don't declare `failed_attempts` until a nested context.

spec/requests/sessions_spec.rb
seannemann21: That's what I thought, as well, but I was running into issues with it coming back as a string. I think cookies can only be stored as strings, which was why this was necessary, but I could be wrong.

julianguyen:
I just added the keys to production! Feel free to merge when you're ready :)

Issue #1601, opened 2019-11-05
Add time support for Medications


julianguyen:
# Description

Currently, users can specify the days they want to be notified of Medication reminders. This notification is sent at 12:00 am, at the beginning of each day. We want to add ability to specify the time for each day of the week.

# Screenshots

Currently: 

![image](https://user-images.githubusercontent.com/3010728/68189807-05756100-ff61-11e9-96d2-315b2383f5de.png)

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

anGie44:
@julianguyen is there a specific uicomponent in mind for listing out the times of day? dropdown or checkboxes for example

julianguyen:
@anGie44 Yes, good question! You should be able to use the `<Input />` component and pass in `type="time"`. We should add examples for this and `type="date"` in our Storybook but here's a link to `<Input type="text" />` to get you started: http://design.if-me.org/?path=/story/input--text

SonuToor:
Hi @anGie44,

I was wondering if you're still working on this issue? It's been a while since I contributed to if-me and I was looking for something to work on. If you'd like help or if you're not working on this issue anymore let me know!

SonuToor:
Hi @julianguyen, 

It's been a while since I contributed and I was looking for something to do. I reached out to @anGie44 to see if she was still active on this issue, I don't think she is.

Should I wait a bit longer or do you think it's safe to take this issue on?

julianguyen:
Go ahead and assign it to yourself @SonuToor. Thanks for reaching out! :D

SonuToor:
Hi @julianguyen , 

Just have a couple of questions to clarify the nature of the new feature.

When a user sets a time, is that time for every single day? Or should a user be able to set a separate time for every single day?

julianguyen:
> Or should a user be able to set a separate time for every single day?

This is what we want! Good question :)

We could add a checkbox for setting the same time daily, but we should give users the ability to fine-tune time for each day.

SonuToor:
@julianguyen so the time input that I am creating is for the reminder to take the medication not the refill reminder correct? 

[...]

PR #1946, opened 2020-12-29
Feature/add time to medication alerts

marthaerm:
# Description
First approach to add time support for medication alerts

## More Details
I created the `db/migrate/20201229031222_add_time_to_medication_daily_reminders.rb` migration that adds a column for each day to add the time for daily reminders. Also, the frontend time inputs were added after the daily checkbox.
The related issue is not completely solved yet but I would like to have a review on the approach.
I believe the next steps to close the issue are:
- Refactor the way to add the 6 inputs and checkbox for each day
- Add a default value to the time inputs
- Store the time on the database on the correspondent field
- Modify the email sender to send it on the time selected
- Create a migration to add a time for each day on the current data (and decide if to add a default value or let the user chose it).

I would like to keep working on this, I'm not sure if I'll be able to finish it but, I'll keep you updated

## Corresponding Issue
#1601.

# Screenshots
image

---

### DIFF
--- a/db/migrate/20201229031222_add_time_to_medication_daily_reminders.rb
+++ b/db/migrate/20201229031222_add_time_to_medication_daily_reminders.rb
@@ -1,0 +1,11 @@
+class AddTimeToMedicationDailyReminders < ActiveRecord::Migration[6.0]
+  def change
+    add_column :medications, :sunday, :time
+    add_column :medications, :monday, :time
+    add_column :medications, :tuesday, :time
+    add_column :medications, :wednesday, :time
+    add_column :medications, :thursday, :time
+    add_column :medications, :friday, :time
+    add_column :medications, :staurday, :time
+  end
+end
--- a/app/helpers/medications_form_helper.rb
+++ b/app/helpers/medications_form_helper.rb
@@ -87,7 +87,7 @@ module MedicationsFormHelper
   def medication_weekly_dosage
     {
       type: 'checkboxGroup',
-      checkboxes: days_checkbox,
+      # checkboxes: days_checkbox,
       label: t('common.days'),
       info: t('medications.form.weekly_dosage_hint'),
       required: true
@@ -95,7 +95,7 @@ module MedicationsFormHelper
   end
 
   def extra_fields
-    [medication_weekly_dosage, medication_refill, medication_comments]
+    [medication_refill, medication_comments]
   end
 
   def common_fields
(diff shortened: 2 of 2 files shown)


db/migrate/20201229031222_add_time_to_medication_daily_reminders.rb
```diff
@@ -0,0 +1,11 @@
+class AddTimeToMedicationDailyReminders < ActiveRecord::Migration[6.0]
```
julianguyen, line 1: You'll have to run a migration (`rake db:migrate`) in order for the database schema to be updated (https://github.com/ifmeorg/ifme/blob/main/db/schema.rb). This why when you run our RSpec tests you get the following error:

```
ActiveRecord::PendingMigrationError:


  Migrations are pending. To resolve this issue, run:

          rails db:migrate RAILS_ENV=test
# ./spec/spec_helper.rb:24:in `<top (required)>'
No examples found.

```

db/migrate/20201229031222_add_time_to_medication_daily_reminders.rb
```diff
@@ -0,0 +1,11 @@
...
+  def change
+    add_column :medications, :sunday, :time
+    add_column :medications, :monday, :time
+    add_column :medications, :tuesday, :time
+    add_column :medications, :wednesday, :time
+    add_column :medications, :thursday, :time
+    add_column :medications, :friday, :time
+    add_column :medications, :staurday, :time
```
julianguyen, line 9: Looks like a typo, with the word `saturday`.

db/migrate/20201229031222_add_time_to_medication_daily_reminders.rb
julianguyen: I think the changes you made to `app/helpers/medications_form_helper.rb` will result in failing tests that need to be updated. You should run `rspec` in your dev environment!

julianguyen (commented):
Thanks for taking this on Martha 🎉  This is a good start! Let me know if you need any help or want to do any pairing.

For the Codeclimate issues you're getting, you can also run `rubocop -a` in your dev environment to catch and auto-fix some of those errors.

marthaerm:
Thanks for the comments @julianguyen ! I'm a bit busier now, so I'll not be able to make the changes quickly, I'll let you know as soon as I can work on them, but If meanwhile, someone else wants to work on them, t also would be great! Thank you so much :)

Issue #1947, opened 2021-01-01
Fix update in 'Customize your account' view redirects incorrectly


sebassebas1313:
# Description

When inserted a valid current password, `update` in `Customize your account` view tries to render the result of `POST` to `/users` resulting in a `This page is not working` .

PR #1948, opened 2021-01-01
refactor #update_resource to return boolean

sebassebas1313:
# Description

`cookies.delete(:pwned)` was the problem since this method returns `nil`. This was making `after_update_path_for` to not be called and therefore `update` was returning to `/users` instead of `/users/edit`

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1947
---


bennpham (approved):
Looks like I got the ordering mixed up with the cookies delete on my Devise pwned password issue 👀 . Nice catch! 🎣 🥅 🧤

bennpham:
I just noticed. I approved a bit early, but after the test catches this case it should be good. I forgot to mention, but it seems like when I checked in the `spec/requests/registrations_spec.rb` it seems like we're missing a test case for a regular login without oauth. Let's add to that file (somehow to make it work, the current implementation I have here don't seem to work which I'm a bit confuse why since redirect path should still be `edit_user_registration_path` according to the override on Devise):

```ruby
context "when the user is logged in" do
  let(:user) { create(:user, provider: nil) }
  it_behaves_like "sucessful password update"
end
```

That should cover this regression. Also note for small typo change in that file. `sucessful` should read `successful` (it's missing an extra `c`). 

Also `describe "update password" do` should be `describe "#update_resource" do` base on https://www.betterspecs.org/#describe

sebassebas1313:
> I just noticed. I approved a bit early, but after the test catches this case it should be good. I forgot to mention, but it seems like when I checked in the `spec/requests/registrations_spec.rb` it seems like we're missing a test case for a regular login without oauth. Let's add to that file (somehow to make it work, the current implementation I have here don't seem to work which I'm a bit confuse why since redirect path should still be `edit_user_registration_path` according to the override on Devise):
> 
> ```ruby
> context "when the user is logged in" do
>   let(:user) { create(:user, provider: nil) }
>   it_behaves_like "sucessful password update"
> end
> ```
> 
> That should cover this regression. Also note for small typo change in that file. `sucessful` should read `successful` (it's missing an extra `c`).
> 
> Also `describe "update password" do` should be `describe "#update_resource" do` base on https://www.betterspecs.org/#describe

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,46 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/BlockLength
 RSpec.describe 'Registrations', type: :request do
-
-  shared_examples "sucessful password update" do
-    let(:update_params) { { password: new_password, password_confirmation: new_password } }
+  shared_examples 'successfully updates user and redirects' do
```
sebassebas1313, line 5: This description represents more accurately the test case.

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,46 @@
...
     before do
       sign_in user
       put user_registration_path, params: update_params
+      user.reload
     end
 
     it { expect(response).to redirect_to(edit_user_registration_path) }
+    it { expect(user.comment_notify).to be(false) }
```
sebassebas1313, line 13: Random attribute to check if attribute's update is ok.

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,46 @@
...
   end
 
-  describe "update password" do
-    let(:new_password) { "foobar_w!th_@" }
+  describe '#update' do
```
sebassebas1313, line 16: Since the test flow goes from the controller method calling until the redirection

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,46 @@
...
+    let(:update_params) { { user: { comment_notify: 'false' } } }
 
-    context "when the user is logged in with Google OAuth" do
+    context 'when the user is logged in with Google OAuth' do
       let(:user) { create(:user_oauth, provider: 'google_oauth2') }
-      it_behaves_like "sucessful password update"
```
sebassebas1313, line 20: User's attribute update when user is logged through a third party is not allowed. In this case #update_without_password is called in the controller which by default does not allows password updates.

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,46 @@
...
+    context 'when the user is logged in with Facebook' do
       let(:user) { create(:user_oauth, provider: 'facebook') }
-      it_behaves_like "sucessful password update"
+      it_behaves_like 'successfully updates user and redirects'
+    end
+
+    context 'when the user is logged in' do
+      let(:user) { create(:user, provider: nil) }
+      let(:update_params) do
+        { user: { name: user.name, email: user.email, password: new_password, password_confirmation: new_password,
```
sebassebas1313, line 32: Add 'user' as parent in payload. Before this spec was buggy because the tests were actually testing a user update with an empty hash.

spec/requests/registrations_spec.rb
bennpham: I think this makes the test a bit tricky to follow. It doesn't feel explicit to me since now every test would be dependent on `comment_notify` to be false otherwise it'll be a flaky test which fails once that case isn't met. 

I think following [Single expectation test](https://www.betterspecs.org/#single), it's best to just isolate to just one assertion and test that it redirects. 

Maybe for this, we should ditch the 

```ruby
it { expect(user.comment_notify).to be(false) }
```

and rename this to `shared_examples "successfully redirects"`

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,46 @@
...
+                  current_password: user.password, comment_notify: 'false' } }
+      end
+      let(:new_password) { 'foobar_w!th_@' }
+      it_behaves_like 'successfully updates user and redirects'
+
+      it 'successfully changes password' do
+        sign_in user
+        put user_registration_path, params: update_params
+        user.reload
```
bennpham, line 41: I think we're missing the assertion here `expect` command, but I'd probably ditch the test altogether base on my thoughts below.

```suggestion
      it 'successfully changes password' do
        sign_in user
        expect { put user_registration_path, params: update_params }.to change { user.password }
```

---

While trying this myself, I feel like there's a problem here and that the current `update_params` password isn't going to change. I feel like we're trying to test Devise implementation here which we should expect and trust that the gem tests the password change implementation itself. 

You'd have to look at [Devise gem](https://github.com/heartcombo/devise) itself here to see exactly what `update_params` to use. My question would be why would this be necessary? This is something that I'd expect Devise to test itself. From what I remember one time, I forgot if it was the password or password confirmation that was hashed and passed on the way (you'll have to reread Devise implementation again), but that'd be overkill for this implementation test.

bennpham (commented):
Rereading through the test itself. 2 things I'll note. 

- Let's change shared_examples to `successfully redirects` to stick to just [single expectation](https://www.betterspecs.org/#single) & just test redirects
- Let's skip the check successful update part since I think I feel like we're testing the implementation of Devise gems and Devise is a well tested and accepted standard for user handling/authentication/etc which it should be well tested in itself in its implementation

I think the main issue and defects was that updating something didn't return the correct response. In such case, my previous cookie implementation caused a 500 error when it should've been a successful status.

spec/requests/registrations_spec.rb
sebassebas1313: I agree with this

spec/requests/registrations_spec.rb
sebassebas1313: I agree with you. We are going too way beyond the issue's scope. The only test here should be if the action is redirecting properly. I am going to delete all the unnecessary testing. 
Regarding your last point, I have look into the actual request's payload (deploying locally the app) and the payload was exactly the way I set it up. Either way, as discussed above this is out of scope.

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,52 @@
...
+# rubocop:disable Metrics/BlockLength
 RSpec.describe 'Registrations', type: :request do
-
-  shared_examples "sucessful password update" do
-    let(:update_params) { { password: new_password, password_confirmation: new_password } }
+  shared_examples 'successfully redirects' do
     before do
       sign_in user
       put user_registration_path, params: update_params
+      user.reload
```
bennpham, line 9:
```suggestion
```

If we're just checking redirects, this is probably dead code now.

spec/requests/registrations_spec.rb
```diff
@@ -1,28 +1,52 @@
...
+                    password: new_password, current_password: user.password} }
+        end
+        it_behaves_like 'successfully redirects'
+      end
+
+      context 'with invalid params' do
+        it 'does not redirects' do
+          sign_in user
+          put user_registration_path, params: {}
+          user.reload
```
bennpham, line 45:
```suggestion
```

I think this is dead code now since we don't need to check assertion for the user model itself.

bennpham (commented):
A few nitpicks on `user.reload`. I haven't checked whether the edge cases pass/fail if I rearrange the cookies delete yet until I switch computer to check out your branch, but if that case does cover it, I'm good with that 👍

Issue #425, opened 2017-03-04
More writing prompts for Moments


julianguyen:
# Description

We currently use static writing prompts in Moments.
![image](https://cloud.githubusercontent.com/assets/3010728/23581694/d36c5324-00cd-11e7-8014-d87b9c85acfe.png)

We want to add the ability to choose from a dropdown of writing prompts and have that saved in the user's Moment.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

jnamster:
![](https://github.trello.services/images/mini-trello-icon.png) [(4) More writing prompts for moments](https://trello.com/c/2Q6ki0qV/22-4-more-writing-prompts-for-moments)

Paarmita:
@julianguyen can I work on this issue?

PR #1972, opened 2021-04-06
Ability to use templates for moments 

julianguyen:
# Description

Currently, the experience for writing Moments is limited to two questions: "What happened and how do and you feel?" and "What thoughts would you like to have?"

image

We want to make this feature more flexible and add the ability to apply reusable templates that serve as writing prompts. 

### Navigating to Moment Templates

image

### Moment Templates page

image

### Add premade templates

image

image

image

### Create a template

image

### Choose a template in the Moments editor

image

image

image

image

### Moments editor when there are no templates

image

### Legacy Moments editor

image

image

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/425

---


bmwachajr:
This is awesome @julianguyen . I will do a deeper review on Wednesday but the first looks at this feature feels very compelling.

bennpham:
@julianguyen Looks awesome! I'll probably get around to reviewing this hopefully on the weekend. Busy this sprint as bug triager and some milestone at work!

config/routes.rb
```diff
@@ -81,6 +81,12 @@
   match 'care_plan_contacts/update', to: 'care_plan_contacts#update', via: :patch
   match 'care_plan_contacts/destroy', to: 'care_plan_contacts#destroy', via: :delete
 
+  match 'moment_templates', to: 'moment_templates#index', via: :get
+
+  match 'moment_templates/create', to: 'moment_templates#create', via: :post
+  match 'moment_templates/update', to: 'moment_templates#update', via: :patch
+  match 'moment_templates/destroy', to: 'moment_templates#destroy', via: :delete
+
```
bennpham, line 89: Could these also be shortened to:

```suggestion
  resources :moment_templates, only: [:index, :create, :update, :destroy]
```

db/migrate/20210328174852_create_moment_templates.rb
```diff
@@ -0,0 +1,13 @@
+class CreateMomentTemplates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :moment_templates do |t|
+      t.string :name
+      t.text :description
+      t.string :slug
+      t.index :slug, unique: true
+      t.integer :user_id
```
bennpham, line 8: I think this is preferred to set a foreign key on `user_id`. Base on the factory I saw below, I assume that Moment Templates belonging to a user would be optional?

```suggestion
      t.references :user, foreign_key: true, null: true
```

app/models/moment_template.rb
```diff
@@ -0,0 +1,30 @@
...
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+
+class MomentTemplate < ApplicationRecord
+  extend FriendlyId
+
+  friendly_id :name
+  validates :user_id, :name, :description, presence: true
+  belongs_to :user
```
bennpham, line 20: I'm curious, base on the MomentTemplate factory, we could have a MomentTemplate without a user, right? If that's the case, we'd probably need to set the following:

```suggestion
  belongs_to :user, optional: true
```

`belongs_to` since rails 5 requires an ID by default, but if we want this to not be `nil`, disregard this message.

bennpham (commented):
Looks pretty good to me overall 🎉 . Just a few nitpicks in terms of possible shorthand alternatives for arrays.

One overall question I have is in regards to the migration creation for the table and should `user_id` be nullable for `MomentTemplate`. 👍

app/views/moment_templates/index.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= react_component('MomentTemplates', html_options: {}, props: {
+  templates: @templates
+}) %>
```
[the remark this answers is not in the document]
julianguyen: It doesn't matter!

PR #1974, opened 2021-04-16
Refactor pages controller spec to pages request spec

bennpham:
# Description

Refactored `spec/controllers/pages_controller_spec.rb` to `spec/requests/pages_spec.rb`.

## More Details

- moves some expect receive to allow and `have_received` so the `Arrange` and `Assert` part of the code is easier to follow
- I actually decided to leave the `assigns` here

I actually decided to leave the `assigns` here. In this case, we won't be able to remove the `rails-controller-testing` gem, but I'd recommend not to write new `assigns` or `assert_template` in the future. I was stuck on how to rewrite the tests without using `assigns` here, but I can't think of a clean way to do it. If I were to throw some of these in a feature spec for existing tests, that'd be messier, and I feel like we'd be losing some important test functionality. I assume we'd want to know those assigned variables were asserted properly because it looks like to be that they'd be pass over to some of the React compnents to do what it needs to do. 

I found [another open source project where the maintainer decided to not remove `assigns`](https://github.com/bikeindex/bike_index/issues/845#issuecomment-499160942) and just deprecate future controller specs instead. Unless someone knows another way to remove the existing assigns test here cleanly (maybe we'd get a new PR for that if someone have a solution), I'll just leave them here and keep the `rails-controller-testing` gem without having to lose existing functionality. 👍 

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1833

# Screenshots

---


julianguyen (approved):
LGTM! Thank you as always for your refactoring magic Ben 👏  I think the decision to keep `assigns` works and makes sense, so thanks for explaining why.

Issue #1987, opened 2021-05-31
Add the ability to remove formatting when copying and pasting in editor


julianguyen:
# Description

If you copy and paste text from another page/screen into our editor, the formatting of the text is copied like so:

![image](https://user-images.githubusercontent.com/3010728/120220239-4abb4c80-c1f1-11eb-99fe-3b4a14dd2e09.png)

By default, it would be preferrable to automatically remove HTML formatting.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

nad0m:
@julianguyen How should we handle the case where the user copy and pastes a formatted text from the current editor?

julianguyen:
@nad0m Good question! I think it should be copied and pasted as plaintext!

PR #1988, opened 2021-06-01
Fix: Remove HTML formatting when pasting in editor

nad0m:
# Description

Removed formatting when copying and pasting in editor.

## More Details

By default, a container with `contenteditable` set to true will allow HTML formatting to be pasted into its inner content. There are a few ways to prevent this from happening:

1. Instead of a generic `div`, we can use `textarea` or `input` elements for our text field. These elements will sanitize any HTML being pasted in and also serves for good accessibility and HTML semantics. 

2. Intercept the user's clipboard and sanitize it with JavaScript before pasting it.

Due to the nature of how the pell action bar is implemented, option 2 will be the simplest way to achieve this.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1987

# Screenshots

### Before
new-moment-before

### After
new-moment-after

---

### DIFF
--- a/client/app/components/Input/InputTextarea.jsx
+++ b/client/app/components/Input/InputTextarea.jsx
@@ -90,13 +90,13 @@ export function InputTextarea(props: Props) {
     }
   };
 
-  const onPaste = e => {
+  const onPaste = (e) => {
     e.preventDefault();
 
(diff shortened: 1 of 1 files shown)


julianguyen (approved):
Thanks a lot for working on this!

PR #1996, opened 2021-07-02
Fixed #1898 Added white background to Pell editor options.

### LINKED ISSUE #1898: Pell WYSIWYG editor color not accessible
# Description
The active colour for the Pell editor options is not an extreme enough colour change. On the current background I actually needed to squint very hard at the screen in order to tell that the italics button had been engaged successfully. For accessibility purposes, it should be obvious that colour has changed between active and inactive.

In this case, the switch from white to the current green/yellow colour only has a contrast ratio of 1.34:1. For text displayed on a background for bolded text it is required to be at least 3. I know the implementation is different to text displayed on a background here but I think the same basic principle holds. The contrast must be strong enough between the two states so that folks with accessibility issues can engage with the controls properly.

# Reproduction Steps
Go to `/users/edit` and scroll down to the About section textarea. Toggle the Pell editor styling options on and off and observe the difference in colour.

# Screenshots
Inactive: 
image

Active:
image

---

adang48:
# Description

I added a white background to the Pell Editor options to increase visibility.

## More Details

I was having trouble finding a color that passed the accessibility tests for mulberry and white, so I made the options mulberry colored with a white background instead. I wanted to find a variant of key-lime, but none passed the accessibility tests for white. I think an accessible color for mulberry and white is difficult to find because mulberry is a darker color and white is a lighter color, so their contrasting colors blend in with one another. For example, a bright color like yellow contrasts well with mulberry but blends in with white, making it unusable.

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/1898

# Screenshots

  Screenshots (required for user interface work), remove if not applicable
Before:
image
Contrast Ratio 1.34:1

After:
image
Contrast Ratio 12.04:1

### DIFF
--- a/client/app/components/Input/InputTextarea.scss
+++ b/client/app/components/Input/InputTextarea.scss
@@ -57,11 +57,12 @@
     }
 
     &Selected {
-      color: $key-lime;
+      background: white;
+      color: $mulberry;
 
       b,
       strong {
-        color: $key-lime !important;
+        color: $mulberry !important;
       }
     }
   }
--- a/client/app/components/Input/InputTextarea.scss
+++ b/client/app/components/Input/InputTextarea.scss
@@ -57,7 +57,7 @@
     }
 
     &Selected {
-      background: white;
+      background:$white;
       color: $mulberry;
 
       b,


client/app/components/Input/InputTextarea.scss
```diff
@@ -57,11 +57,12 @@
     }
 
     &Selected {
-      color: $key-lime;
+      background: white;
```
julianguyen, line 60: This totally sufficient, but for consistency's sake, let's use [$white](https://github.com/ifmeorg/ifme/blob/main/app/assets/stylesheets/base/_colors.scss#L18)!

julianguyen (commented):
Thanks so much for looking into this and your decision to use the white background is a good call!

client/app/components/Input/InputTextarea.scss
adang48: Great, just changed it!

adang48:
> Thanks so much for looking into this and your decision to use the white background is a good call!

Glad to help!

julianguyen (approved):

PR #2133, opened 2022-10-02
Improve/add appropriate message during workspace setup step

rohandaxini:
# Description

Currently, when we run `bin/rake setup_workspace` [step](https://github.com/ifmeorg/ifme/wiki/Installation#configuration-files), it's not clear (i.e. does not indicate) whether the setup process is completed properly or not. This change adds an appropriate message with a check whether the setup process was successful or not.

---

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -42,6 +42,8 @@ task setup_workspace: :environment do
     end
     File.write(target, content)
   end
+
+  puts "Workspace setup completed."
 end
 
 # Run the secret task, grab the output, strip the new lines
--- a/Rakefile
+++ b/Rakefile
@@ -25,25 +25,30 @@ end
 
 desc 'Automate the Config Setup for New Environments'
 task setup_workspace: :environment do
-  SECRETS = {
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -42,6 +42,8 @@ task setup_workspace: :environment do
     end
     File.write(target, content)
   end
+
+  puts "Workspace setup completed."
```
julianguyen, line 46: I think this will still print when there is an error. I think we should only display this when no errors are thrown, and rephrase it as "Workspace setup completed successfully 🎉".

julianguyen (changes requested):
Good catch, thanks for working on this!

Rakefile
rohandaxini: @julianguyen Agree 👍 
I have improved the exception handling in this task. Could you please review it again?

rohandaxini:
Hi @julianguyen Could you please review this PR and merge if it looks good?

julianguyen (approved):

Issue #2120, opened 2022-10-01
Add ARIA attribute to element that opens the Tooltip component


julianguyen:
# Description

In the [Tooltip component](http://design.if-me.org/?path=/docs/components-tooltip--element-is-text) ([code](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Tooltip/index.jsx)), the element that opens the tooltip has an `aria-labelledby` attribute on it so that it can be associated with the element with `role="tooltip"`.  This element that opens the tooltip does not have a `role` itself. This will reduce the quality of the screen-reader experience.

This is a problem because the `aria-labelledby` attribute cannot be used on a `div` with no valid role attribute.

We don't want to assign a role to this `div` because it could contain an element of a different role. The ideal solution would be to add `aria-labelledby={tooltipId}` to the child that is rendered i.e. `{Utils.renderContent(element)}`.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PaxtonDevs:
Hello, I would like to assign myself this issue

julianguyen:
@PaxtonDevs Added you to our org! After accepting the invitation, you should be able to assign yourself to the issue.

PaxtonDevs:
Awesome, thank you!

andomain:
Hi, if there is no further activity on this I'd like to contribute?

a-camarillo:
Hi, I noticed this issue is still unresolved and would like to tackle it!

I took a look at [#2156] and suggest possibly breaking this issue up into two smaller issues depending on if we want to write custom functionality for rendering HTML as a React Element or use one of the [other libraries](https://github.com/hatashiro/react-render-html#deprecation) recommended by the old react-render-html maintainers.

PR #2156, opened 2022-10-08
Added ARIA Attribute To Element [#2120]

PaxtonDevs:
# Description

## Corresponding Issue

# Screenshots

---

### DIFF
--- a/client/app/components/Tooltip/index.jsx
+++ b/client/app/components/Tooltip/index.jsx
@@ -26,8 +26,8 @@ export const Tooltip = (props: Props): Node => {
   const tooltipId = id || `tooltip${Utils.randomString()}`;
   return (
     <div className={css.tooltipWrapper}>
-      <div className={css.tooltipElement} aria-labelledby={tooltipId}>
-        {Utils.renderContent(element)}
+      <div className={css.tooltipElement}>
+        <div aria-labelledby={tooltipId}>{Utils.renderContent(element)}</div>
       </div>
       <div
         id={tooltipId}


client/app/components/Tooltip/index.jsx
```diff
@@ -26,8 +26,8 @@ export const Tooltip = (props: Props): Node => {
   const tooltipId = id || `tooltip${Utils.randomString()}`;
   return (
     <div className={css.tooltipWrapper}>
-      <div className={css.tooltipElement} aria-labelledby={tooltipId}>
-        {Utils.renderContent(element)}
+      <div className={css.tooltipElement}>
+        <div aria-labelledby={tooltipId}>{Utils.renderContent(element)}</div>
```
julianguyen, line 30: This doesn't fix the problem [outlined in the issue](https://github.com/ifmeorg/ifme/issues/2120) since `aria-labelledby` is still being applied to `div` which has no default value for `role`. We'll want to pass in an optional argument into [Utils.renderContent()](https://github.com/ifmeorg/ifme/blob/8201c2f1923a57e303a7eccb66173c73c5f1052c/client/app/utils/index.js#L36) that lets us apply custom props or HTML attributes to the `element`.

For this util function, we use a third-party dependency called [render-react-html](https://github.com/hatashiro/react-render-html). It looks like it's no longer maintained, so we likely want to write something more custom to be able to render HTML as a React element and ALSO apply custom props / HTML attributes.

Ahh, so I realize this may not be a `beginner friendly` issue as it was originally labelled as 😅  Let me know if you still want to work on it, happy to work on it together as well!

julianguyen (changes requested):
Thanks so much for taking this on!

client/app/components/Tooltip/index.jsx
julianguyen: Looks like the prop type for `element` is `any` (which is not the best decision 😅 ), but that very likely means it could be a React `Node` or a `string`. I think if it's a `Node`, you should be able to do something like `element.aria-labelledby = tooltipId;`. 

If it's a `string`, then the changes I mentioned above is the way to go.

PaxtonDevs: Oh haha, yes, I attempted this issue as a beginner in React and took this on because of the beginner friendly label. 

I'm definitely willing to work on it together. I have sent an email to join the slack channel.

julianguyen: Let's sync on this on Slack then!

Issue #2121, opened 2022-10-01
Improve aria-label values in non-text links in the StoryActions component


julianguyen:
# Description

The [Story component](http://design.if-me.org/?path=/docs/components-story--all-props) ([code](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Story/index.jsx)) appears on many pages, including the signed-in homepage when there are viewable Moments or Strategies and any signed-in page that's displaying a list of items.

The [StoryActions sub-component](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Story/StoryActions.jsx) has non-text links (icons with tooltips) for functionality like editing and deleting. There are existing values for `aria-label` on these non-text links, but they should be improved.

Let's include the name of the story that the action is being applied to. The [change should be made here](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Story/StoryActions.jsx#L120), and we should concatenate the prop `name` from `Story` to this string. 

e.g. "Edit Story-Name-Here"

We'll have to account for internationalization (i18n) for different grammar structures in different languages. Feel free to defer to Google Translate for verification.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #2158, opened 2022-10-08
[UPDATE] added story name to aria label

nikhilbhatt:
# Description
Updated aria label value in story actions with story action & story name

## Corresponding Issue
#2121 

---


client/app/components/Story/StoryActions.jsx
```diff
@@ -41,6 +41,7 @@ export type Props = {
   actions: Actions,
   hasStory?: boolean,
   dark?: boolean,
+  name?: string,
```
julianguyen, line 44: I think to avoid naming collisions, we should rename this prop to `storyName` and this prop should be required, so we can remove the `?`. This also fixes the Flow error in the failed CI build. You should run `yarn lint` to reproduce it locally and verify that it's been fixed.

julianguyen (changes requested):
Thanks so much for taking this on! Looking good, just some feedback and we should be good to go!

client/app/components/Story/StoryActions.jsx
```diff
@@ -40,6 +40,7 @@ export type Actions = {
 
 export type Props = {
   actions: Actions,
+  storyName: string,
```
julianguyen, line 43: You'll want to pass in `storyName` as a prop in all use cases of `StoryActions`. As raised in the Flow failure in the [recent CI build](https://app.circleci.com/pipelines/github/ifmeorg/ifme/1913/workflows/d764a95f-2e6f-4118-a507-c0b114291060/jobs/15457), we're missing the prop in https://github.com/ifmeorg/ifme/blob/main/client/app/widgets/Comments/index.jsx#L113.

doc/pages/blurbs.json
```diff
@@ -682,5 +682,14 @@
     "link_name": "manishsharma.ml",
     "link": "https://manishsharma.ml",
     "social": "globe"
+  },
+  {
+    "name": "Nikhil Bhatt",
+    "image": "/assets/contributors/nikhil_bhatt.jpg",
+    "profile": "Techie, who loves to solve the real world problems using code. Currently working as a Full stack developer in Ruby on Rails and React",
```
julianguyen, line 689: Grammar fix: `Currently, I'm working as a full-stack developer in Ruby on Rails and React.`

julianguyen (changes requested):
Thanks for the changes! Also yay for adding a contributor blurb 🎉 

There's one more change that needs to be done before we're good to merge!

client/app/components/Story/StoryActions.jsx
nikhilbhatt: yes, passed but the problem is we comment prop will be something like `<div>Comment</div>`  so we can't show it inside aria-label component.

julianguyen (changes requested):
Looks like there's some ESLint failures! I recommend running `yarn lint:eslint` locally so you can fix them.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -86,7 +86,7 @@ describe('Comments', () => {
       expect(
         screen.getByRole('button', { name: 'Submit' }),
       ).toBeInTheDocument();
-      expect(screen.getByRole('link', { name: 'Report' })).toBeInTheDocument();
+      expect(screen.getByRole('link', { name: `Report ${value}` })).toBeInTheDocument();
```
julianguyen, line 89: So this test reveals that for the case of the Report action, we want to pass in the user's name instead of the comment.

client/app/widgets/Comments/__tests__/Comments.spec.jsx
```diff
@@ -106,7 +106,7 @@ describe('Comments', () => {
       await waitFor(() => expect(screen.getByRole('article')).toBeInTheDocument());
       expect(screen.getByRole('article')).toHaveTextContent('Hey');
 
-      userEvent.click(screen.getByRole('link', { name: 'Delete' }));
+      userEvent.click(screen.getByRole('link', { name: `Delete ${value}` }));
```
julianguyen, line 109: This test also bring up another edge case, we don't want to pass in the entire comment text here. Here, it makes more sense to have something like "Delete from comment from [the user the comment is by]".

client/app/components/Story/StoryActions.jsx
julianguyen: Right, I think for the comment case we have edge cases since comments don't have a name. I think we want to use the user's name that has made the comment post.

julianguyen (changes requested):
Great work fixing the broken tests 🎉 

We have some edge cases with comments which I forgot about 😆 I think after addressing that, we should be good to go 🚀 

Thanks for all the updates!

app/views/groups/_info.html.erb
```diff
@@ -33,7 +33,8 @@
           dataConfirm: t('common.actions.confirm'),
           dataMethod: 'delete'
         } : nil
-      }
+      },
+      storyName: local_assigns[:group].description
```
julianguyen, line 37: I think we want to use the Group name here instead of the description, which could be long.

```suggestion
      storyName: local_assigns[:group].name
```

client/app/widgets/Comments/index.jsx
```diff
@@ -104,6 +104,7 @@ export const Comments = ({ comments, formProps }: Props): Node => {
       deleteAction,
     } = myComment;
     const author = <a href={`/profile?uid=${commentByUid}`}>{commentByName}</a>;
+    const commentStoryName = `from comment from ${commentByName}`;
```
julianguyen, line 107: We use i18n for text in our app. You'll want to add a key in [this part of en.yml](https://github.com/ifmeorg/ifme/blob/main/config/locales/en.yml#L113) and add it to the other translation files. You can use Google translate for this, and we'll verify the translations later.

I should create two keys like the following:

```
report_comment_by: 'Report comment by %{name}`
delete_comment_by: 'Delete comment by %{name}`
```

julianguyen (changes requested):
Thanks so much for the updates 🎉 

One more change needs to be done!

julianguyen (approved):
Fantastic work! Thanks so much for taking this on 🎉 🎊

Issue #2159, opened 2022-10-08
Fix keyboard navigation for navigation menu in mobile web


julianguyen:
# Description

There are two issues with the keyboard navigation for the navigation menu in mobile web:

1. When a user uses the `Tab` key to navigate through a page, the navigation menu is opened. It should only be opened after the `Tab` key focuses on the hamburger icon and the user hits the `Enter` key to open it.
2. There is no focus trap on the navigation menu when it's opened (as a modal)

The following recent [PR for "[FIX] Keyboard navigation for modals"](https://github.com/ifmeorg/ifme/pull/2127) by @maniSHarma7575 fixed the issue in other areas. So you'll want to dig deeper into this part of the codebase as to why it's not working for the navigation menu in mobile web.

This navigation menu is part of the [Header component](http://design.if-me.org/?path=/docs/components-header--no-active-link).

![Oct-08-2022 08-12-57](https://user-images.githubusercontent.com/3010728/194714471-216410f3-ab78-46cf-9421-43b3bd0054b6.gif)

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

PR #2166, opened 2022-10-11
Keyboard navigation in mobile web (Issue #2159)

LMulvey:
# Description

Few a11y fixes related to the Header component

- Fixed bug where `toggle` was being called _every time_ `onKeyDown` was triggered on the hamburger button

navigation-bug-fixed

- Introduced focus trapping for the mobile navigation bar while open - you can no longer focus out of the nav via the keyboard without closing it

trap-focus

- Upgraded `@testing-library/jest-dom`, `@testing-library/react` and `@testing-library/user-event` - the dependency update was largely to use the updated `userEvent.keyboard` event support from `@testing-library/user-event` but I wanted to ensure the dependencies were in sync. 

## Corresponding Issue

#2159 

---

### DIFF
--- a/client/app/components/Header/__tests__/Header.spec.jsx
+++ b/client/app/components/Header/__tests__/Header.spec.jsx
@@ -4,6 +4,13 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import Header from 'components/Header';
 
+function setup(jsx) {
+  return {
+    user: userEvent.setup(),
+    ...render(jsx),
+  };
+}
+
 const component = (
   <Header
     home={{ name: 'Home', url: '/some-path' }}
@@ -16,27 +23,46 @@ const component = (
 
 describe('Header', () => {
   it('renders correctly', () => {
-    render(component);
+    setup(component);
 
     expect(screen.getByRole('banner')).toBeInTheDocument();
     expect(screen.getByRole('navigation')).toBeInTheDocument();
     expect(screen.getByRole('link', { name: /home/i })).toBeInTheDocument();
     expect(screen.getByRole('link', { name: /link 1/i })).toBeInTheDocument();
     expect(screen.getByRole('link', { name: /link 1/i })).toHaveClass(
-      'headerActiveLink',
+      'headerActiveLink'
     );
     expect(screen.getByRole('link', { name: /link 2/i })).toBeInTheDocument();
   });
 
-  it('toggles hamburger correctly', () => {
-    const { container } = render(component);
+  it('toggles hamburger correctly', async () => {
+    const { container, user } = setup(component);
     const hamburger = container.querySelector('#headerHamburger');
-    userEvent.click(hamburger);
+    await user.click(hamburger);
 
     const mobile = container.querySelector('#headerMobile');
     expect(mobile).toBeInTheDocument();
 
-    userEvent.click(hamburger);
+    await user.click(hamburger);
     expect(mobile).not.toBeInTheDocument();
+  });
+
+  it('toggles hamburger from keyboard input when expected', async () => {
+    const { container, user } = setup(component);
+    const hamburger = container.querySelector('#headerHamburger');
+
+    await user.keyboard('{Tab}{Tab}');
+    expect(hamburger).toHaveFocus();
+    await user.keyboard('{Enter}');
+    expect(container.querySelector('#headerMobile')).toBeInTheDocument();
+    await user.keyboard('{Enter}');
+    expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
+    await user.keyboard('[Space]');
+    expect(container.querySelector('#headerMobile')).toBeInTheDocument();
+    await user.keyboard('[Space]');
+    expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
+
+    await user.keyboard('{Tab}');
+    expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
   });
 });
--- a/client/app/components/Header/__tests__/Header.spec.jsx
+++ b/client/app/components/Header/__tests__/Header.spec.jsx
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, getByRole } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import Header from 'components/Header';
 
@@ -63,6 +63,31 @@ describe('Header', () => {
     expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
 
     await user.keyboard('{Tab}');
+    expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
+  });
+
+  it('traps focus when the mobile navbar is open', async () => {
+    const { container, user } = setup(component);
+    const hamburger = container.querySelector('#headerHamburger');
+
+    await user.keyboard('{Tab}{Tab}');
+    expect(hamburger).toHaveFocus();
+    await user.keyboard('{Enter}');
+    expect(container.querySelector('#headerMobile')).toBeInTheDocument();
+
+    await user.keyboard('{Tab}');
+    expect(screen.getByRole('link', { name: /link 1/i })).toHaveFocus();
+    await user.keyboard('{Tab}');
+    expect(screen.getByRole('link', { name: /link 2/i })).toHaveFocus();
+    await user.keyboard('{Tab}');
+    expect(screen.getByRole('link', { name: /home/i })).toHaveFocus();
+    await user.keyboard('{Tab}');
+    expect(hamburger).toHaveFocus();
+    await user.keyboard('{Tab}');
+    expect(screen.getByRole('link', { name: /link 1/i })).toHaveFocus();
+
+    // Shift-tab back to the hamburger and close the mobile menu
+    await user.keyboard('{Shift>}{Tab}{/Shift}{Enter}');
     expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
   });
 });
--- a/client/app/components/Header/__tests__/Header.spec.jsx
+++ b/client/app/components/Header/__tests__/Header.spec.jsx
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import { render, screen, getByRole } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import Header from 'components/Header';
 
@@ -30,7 +30,7 @@ describe('Header', () => {
     expect(screen.getByRole('link', { name: /home/i })).toBeInTheDocument();
     expect(screen.getByRole('link', { name: /link 1/i })).toBeInTheDocument();
     expect(screen.getByRole('link', { name: /link 1/i })).toHaveClass(
-      'headerActiveLink'
+      'headerActiveLink',
     );
     expect(screen.getByRole('link', { name: /link 2/i })).toBeInTheDocument();
   });
--- a/client/app/components/Header/index.jsx
+++ b/client/app/components/Header/index.jsx
@@ -18,23 +18,29 @@ export type Props = {
 
 export type State = {
   mobileNavOpen: boolean,
-  toggled: boolean,
 };
 
-export const Header = ({
-  home, links, mobileOnly, profile,
-}: Props): Node => {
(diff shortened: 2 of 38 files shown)


LMulvey:
Obviously broke some things with the test upgrades. I'm working through the changes - there were a few API updates with `@testing-library` over some versions.

@julianguyen - happy to update all existing tests to still function. The benefit here is that we get a more realistic testing environment from `user-event` (See: [Differences from fireEvent](https://testing-library.com/docs/user-event/intro#differences-from-fireevent)) but I definitely don't want to disrupt things too much so let me know if you'd like me to step back here and I can update the `Header.spec` tests I've added to try and get them to work with the _older_ APIs.

julianguyen:
@LMulvey Thanks for the summary and for taking this on 🎉 ! Yeah, let's update the existing tests like you said.

LMulvey:
Okay! Tests are updated but there are a few caveats. 

I've had to skip three tests for now - I need to think about an approach a bit. I left a comment on all three tests describing why they were skipped. Still, in summary, `@testing-library/user-event` v14 dropped support for `event.keyCode` (See: https://github.com/testing-library/user-event/issues/842). 

`react-autosuggest` is pretty outdated and [still exclusively looks for](https://github.com/moroshko/react-autosuggest/blob/master/src/Autosuggest.js#L713) `event.keyCode` when detecting whether or not a user hit the Enter key. Because of this, our `Enter` inputs from tests are _not_ working with any components that use `react-autosuggest`.

I've tried manually dispatching an event with JSDom (you can see my scraps in the `InputTag.spec` tests) but it wasn't working. I can try to submit a PR to `react-autosuggest` but it seems to be unmaintained at this point. 

Does anyone have any ideas? 🤔

client/app/components/Input/__tests__/InputTextarea.spec.jsx
```diff
@@ -154,11 +155,13 @@ describe('InputTextarea', () => {
         { title: 'Unordered List', expectedArgs: ['insertUnorderedList'] },
       ];
 
-      buttons.forEach(({ title, expectedArgs }) => {
-        const button = screen.getByTitle(title);
-        userEvent.click(button);
-        expect(pell.exec).toHaveBeenCalledWith(...expectedArgs);
-      });
+      await Promise.all(
```
julianguyen, line 158: Can we use `waitFor` instead?

client/app/components/Input/__tests__/InputTextareaTemplate.spec.jsx
```diff
@@ -161,11 +162,13 @@ describe('InputTextarea', () => {
         { title: 'Unordered List', expectedArgs: ['insertUnorderedList'] },
       ];
 
-      buttons.forEach(({ title, expectedArgs }) => {
-        const button = screen.getByTitle(title);
-        userEvent.click(button);
-        expect(pell.exec).toHaveBeenCalledWith(...expectedArgs);
-      });
+      await Promise.all(
```
julianguyen, line 165: Similar question as above.

client/package.json
```diff
@@ -144,5 +144,6 @@
   },
   "browserslist": [
     "defaults"
-  ]
+  ],
+  "packageManager": "yarn@1.22.19"
```
julianguyen, line 148: What's the rationale for this? Is it necessary?

julianguyen (commented):
Thanks so much for taking this on! Thanks for also clarifying your approach and debugging! 🎉  Hope you're well :)

client/app/components/Header/index.jsx
```diff
@@ -18,23 +18,61 @@ export type Props = {
...
+    if (mobileNavOpen) {
+      // Let's trap the focus
+      if (navigationRef.current) {
+        const focusableElements = navigationRef.current.querySelectorAll(
+          'a, button, input, select, textarea, [tabindex]:not([tabindex="-1"])',
+        );
+        const firstFocusableElement = focusableElements[0];
+        const lastFocusableElement = focusableElements[focusableElements.length - 1];
+
+        if (event.key === 'Tab') {
```
julianguyen, line 41: When I do the following, the focus trap does not work:

1. `Tab` into the hamburger icon
2. Hit `Enter` to open the mobile menu
3. `Tab` through the menu items
4. When the last item of the menu is reached, and I hit `Tab` again, the focus trap does not work and the contents behind the menu are hit.

The expected behaviour is for the focus to go back to logo at the top of the menu.

Oct-11-2022 21-21-15

julianguyen: I think the problem is that `document.activeElement` is not giving you the active element that is being focused on. It's giving the element before.

I think we want to make t[his focus trap code we're using in the Modal component](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Modal/index.jsx#L64) a custom hook that we can reuse here.

client/app/components/Input/__tests__/InputTag.spec.jsx
```diff
@@ -22,90 +29,107 @@ describe('InputTag', () => {
...
-    userEvent.click(option);
+    await user.click(option);
 
     checkbox = screen.getByRole('checkbox', queryOptions);
     expect(checkbox).toBeInTheDocument();
   });
 
-  it('selects a value with keydown without specifying text', () => {
+  /**
+   * Skipping this one for now. react-autosuggest still checks for event.keyCode, which
```
julianguyen, line 78: Have you tried using `fireEvent` instead in these skipped tests? I know it's not ideal since we want to simulate a complete interaction via `userEvent`.

client/app/components/Header/index.jsx
julianguyen: Another issue I see is that `lastFocusableElement` is the "Sign out" button when it should be the "Resources" link.

client/app/components/Header/__tests__/Header.spec.jsx
```diff
@@ -28,15 +35,59 @@ describe('Header', () => {
...
+    await user.keyboard('{Tab}{Tab}');
+    expect(hamburger).toHaveFocus();
+    await user.keyboard('{Enter}');
+    expect(container.querySelector('#headerMobile')).toBeInTheDocument();
+    await user.keyboard('{Enter}');
+    expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
+    await user.keyboard('[Space]');
+    expect(container.querySelector('#headerMobile')).toBeInTheDocument();
+    await user.keyboard('[Space]');
+    expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
LMulvey: Ensuring that it opens AND closes the navbar with the space key!

client/app/components/Input/__tests__/InputTextarea.spec.jsx
LMulvey: What's the rationale for `waitFor` vs `Promise.all`? I'm not sure I follow! My larger gut feeling is to break out of doing loops here entirely and write out all the Button conditions explicitly – just to avoid any hack-y-ness 🤔

client/package.json
LMulvey: 🤦🏻 Not at all. My setup uses Yarn v3 and to avoid changing the lockfile formatting, I ran `yarn set-version classic` - which introduced a bunch of random changes throughout the repo - I thought I had caught them all. Turns out it's not super intuitive to run a separate version of Yarn in a specific workspace. I'll uncommit this and watch for it again!

client/app/components/Input/__tests__/InputTag.spec.jsx
LMulvey: I _did_ try that - turns out it _also_ writes the `keyCode` is `0` so we end up with the same issue 😭

client/app/components/Header/index.jsx
LMulvey: @julianguyen Very interesting! I was testing it in an unauthenticated state on the homepage–I feel like this is a separate state. I'll take a look (and look at lifting the `Modal` component `useEffect` into a reusable hook! 💪🏻 )

LMulvey: This is fixed now! It was related to a `display: none` element in the Links - it isn't technically focusable by the browser but still appeared in our `focusableElements` array. I've added an extra step to filter out non-visible elements here – the Focus Trap now works as expected while authenticated or unauthenticated.

client/app/components/Input/__tests__/InputTag.spec.jsx
```diff
@@ -22,90 +29,107 @@ describe('InputTag', () => {
...
       'true',
     );
-    userEvent.type(input, '{enter}');
-
-    checkbox = screen.getByRole('checkbox', queryOptions);
-    expect(checkbox).toBeInTheDocument();
+    expect(input).toHaveFocus();
+
+    const event = new window.KeyboardEvent('keydown', {
+      keyCode: 13,
```
julianguyen, line 112: I think it would be worth putting a comment here that this `keyCode` refers to the `Enter` key.

client/app/components/Input/__tests__/InputTag.spec.jsx
julianguyen: Thanks for trying all this! Hmm, I'm okay with skipping this test for now, but we should have a fast follow to fix this. Since `react-autosuggest` is deprecated, I think we'll need to migrate to another library. Could you create an issue for this? Are you able to take this on?

I think we can just keep the original contents of this test since it's being skipped anyways.

client/app/components/Input/__tests__/InputSwitch.spec.jsx
```diff
@@ -35,7 +35,7 @@ describe('InputSwitch', () => {
       expect(screen.getByRole('checkbox')).not.toBeChecked();
 
       /**
-       * TODO: Follow up on `userEvent.type(inputSwitch, '{enter}')` in v12.1.7.
+       * TODO: Follow up on `await userEvent.type(inputSwitch, '{enter}')` in v12.1.7.
```
julianguyen, line 38: Let's also make a GitHub issue for this, so it doesn't get lost in comments.

client/app/components/Input/__tests__/InputTextarea.spec.jsx
julianguyen: Could you do something like this instead with [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor)?

```
buttons.forEach(({ title, expectedArgs }) => {
  const button = screen.getByTitle(title);
  userEvent.click(button);
  
  await waitFor(() => expect(pell.exec).toHaveBeenCalledWith(...expectedArgs));
});
```

client/app/components/Modal/index.jsx
```diff
@@ -61,34 +62,12 @@ export const Modal = (props: Props): Node => {
     setOpen(!open);
   };
 
-  useEffect(() => {
-    const focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
-    const modal = modalEl.current;
+  useFocusTrap(modalEl, open);
```
julianguyen, line 65: Woot!

client/app/widgets/QuickCreate/__tests__/QuickCreate.spec.jsx
```diff
@@ -135,7 +147,19 @@ describe('QuickCreate', () => {
   });
 
   describe('when the form', () => {
-    it('is submitted it adds a new checkbox from data', async () => {
+    /**
```
julianguyen, line 150: Similar comment around keeping original test like stated above!

julianguyen (changes requested):
Awesome work! 🎉  Thanks for the updates. Tested things locally, and the keyboard navigation works as expected.

Shared some feedback around moving forward with `react-autosuggest` and also the tests

client/app/components/Input/__tests__/InputTextarea.spec.jsx
LMulvey: Wahoo! Worked like a charm 🎉

client/app/components/Input/__tests__/InputSwitch.spec.jsx
LMulvey: This is an old comment that I was able to just resolve & remove as part of this upgrade 🎉 !

LMulvey:
@julianguyen Thank you for all the feedback! I've pushed up another round of fixes and created #2170 to cover off the `react-autosuggest` changeover task. I should be able to take that on next week once this PR is wrapped up 🤝

julianguyen (approved):
Awesome work! 🎉 Thanks so much for taking this on and navigating all the gotchas in a good way.

Issue #2155, opened 2022-10-08
navbar is not sticky


amitp138:
# Description
navbar of the website is not sticky that means it is not fixed due to which we have to scroll up for selecting any link from the navbar 

# Reproduction Steps

I can make the navbar sticky with responsiveness

# Screenshots

![image](https://user-images.githubusercontent.com/86116926/194696081-8ad77579-d3a5-40e9-854f-4fc798cb2321.png)

above we can see that navbar is not visible once we scroll down

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

julianguyen:
Thanks for creating an issue with us! I think this could be a nice improvement. I think we'll want to talk through the design of this more. 

The different experiences we'll want to consider:
* Mobile web
* Signed-in
* Signed-out


The background of the sticky nav bar could have the same one as the mobile experience:

![image](https://user-images.githubusercontent.com/3010728/194713797-f1d22294-e2c2-418d-bcb0-c81d6f2910e4.png)

saminarp:
I am going to try fixing this issue, so I am assigning this to myself. :)

amitp138:
Then will I get the hactoberfest tag ??

amitp138:
so can you please give me the hactoberfest tag atleast

amitp138:
Hello @saminarp  can you please give me the hactoberfest tag please since I have raised this issue so I should be getting the chance right ??

amitp138:
Thank you sir for the consideration and also how can we discuss about this
like through the mail or anything else

amitp138:
hello @saminarp i am having issue while setting this up so if you want then you can continue with project

SanjoyPator1:
Is this issue fixed or Do you still need help?

saminarp:
I am still looking at it, and will keep it posted after I have updates

amitp138:
Ohk

sfayyad:
I would love to work on this! Can someone please assign it to me?

Demandtech:
I just check your website and I see the issue is still there, I will like to work on it.

MuhammadAakash:
@Demandtech Sure go ahead.

Demandtech:
[...]

PR #2171, opened 2022-10-13
Issue#2155 ux

saminarp:
# Description
With the following changes in `client/app/components/Header/Header.scss`, the navbar will be sticky.

## More Details
The fixed navbar is not supported in mobile view. I intentionally left it out because the background is transparent in the navbar (i.e, background: transparent;) Giving the navbar a subtle color would prevent any clashes from occurring as a result of scrolling.

## Corresponding Issue

This enhancement addresses the issue https://github.com/ifmeorg/ifme/issues/2155. 

# Screenshots

#### Current behavior: navbar is in fixed position even with the scoll.
CleanShot 2022-10-13 at 05 05 29@2x

#### The screenshot below is the background color I would propose using but haven't implemented yet.
This would prevent the clashes with logo and content on the scroll 
CleanShot 2022-10-13 at 05 02 41@2x

---

### DIFF
--- a/client/app/components/Header/Header.scss
+++ b/client/app/components/Header/Header.scss
@@ -97,6 +97,12 @@
   }
 
   &Desktop {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    z-index: 1000;
+
     width: 100%;
     padding: 25px $size-34;
     box-sizing: border-box;
@@ -108,11 +114,11 @@
 
     @media screen and (max-width: $medium) {
       padding: 15px $size-10;
+      position: relative;
     }
 
     &Home {
       @include linkStyle();
-
       font-size: $size-30;
     }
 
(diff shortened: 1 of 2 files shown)


client/app/components/Header/Header.scss
```diff
@@ -97,6 +97,12 @@
   }
 
   &Desktop {
+    position: fixed;
```
julianguyen, line 100: With the header being fixed, you'll want to add some kind of `margin-top` value to the content of the page so that the header doesn't cover things. It should be height of the header + the existing margin or padding top value.

In addition, I think we should have a transparent background colour, I recommend using `$black-80` or `$black-90`.

julianguyen (changes requested):
Thanks so much for taking this on 🎉

client/app/components/Header/Header.scss
saminarp: Sure, I will be working on it! I will reach out via Slack even if I have questions.

saminarp:
Hello, I made a mistake force pushing the changes, please do not merge it. I will be creating a separate branch and push the updated PR

saminarp:
Please refer to this [PR](https://github.com/ifmeorg/ifme/compare/main...saminarp:ifme:UX-Issue-2155?expand=1)  and ignore the previous

PR #2176, opened 2022-10-16
UX: Navigation bar fixed

saminarp:
# Description
This is my follow up PR from https://github.com/ifmeorg/ifme/pull/2171 as I mistakenly force-pushed my commit.

With the following changes in `client/app/components/Header/Header.scss` and `app/assets/stylesheets/static/static_content.scss`, the navbar will be sticky.  

I have added margin-top to the main content of the page as requested in https://github.com/ifmeorg/ifme/pull/2171#discussion_r994865709

## More Details

As. @julianguyen pointed out in https://github.com/ifmeorg/ifme/pull/2171 the background color `$black-80` The result will look like this. However, I haven't made this change yet and wanted to ask if this is the desired result. :) If so, please let me know and I will make the necessary changes.

CleanShot 2022-10-16 at 03 30 21@2x

## Corresponding Issue

https://github.com/ifmeorg/ifme/issues/2155

---


app/assets/stylesheets/static/static_content.scss
```diff
@@ -5,7 +5,7 @@
   box-sizing: border-box;
   padding: $size-40 34px;
   scroll-behavior: smooth;
-  margin: 0 auto;
+  margin: 50px auto;
```
julianguyen, line 8: We'll want to apply similar logic in [/app/assets/stylesheets/dashboard/dashboard_content.scss](https://github.com/ifmeorg/ifme/blob/main/app/assets/stylesheets/dashboard/dashboard_content.scss).

Or else this is happening on the signed in pages:

image

client/app/components/Header/Header.scss
```diff
@@ -97,6 +97,12 @@
   }
 
   &Desktop {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    z-index: 1000;
```
julianguyen, line 104: It would be a better design to have this background only show up once the user starts scrolling in the page. We can do that by applying the transparent black background through React JS in [/client/app/components/Header/index.jsx](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Header/index.jsx)

julianguyen (changes requested):
Thanks for the updates 🎉

julianguyen:
Any updates here?

Issue #2128, opened 2022-10-02
Write unit tests for client/app/utils/index.js


julianguyen:
# Description

Write unit tests for client/app/utils/index.js

Functions include:
* `randomString`
* `setCsrfToken`
* `getPusher`
* `renderContent`

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

akk312000:
I would like to give this a try @julianguyen

julianguyen:
That would be great, thank you! Feel free to assign yourself to the issue @akk312000 :)

akk312000:
@julianguyen  I'm not able assign the issue to myself, can you guide me how should I do that?

julianguyen:
@akk312000 Ahh looks like you're not added to the project org. I just sent you an invitation and assigned you.

saminarp:
@julianguyen  I am investigating this and I wanted to ask if I am supposed to create `__tests__` folder in utils to write the corresponding tests for index.js. Please let me know :)

saminarp:
Hello @julianguyen,
 I noticed that in `package.json jest configurations` there's  `"testRegex": "./app/.*.spec\\.jsx?$"` meaning that this regex will only capture the test files which includes `spec` in the filename. I am writing my tests under the name `index.test.js`. Do you suggest updating the regexp? Or do the unit test files need to be somewhere else with a different naming convention?

lucas-codes-stuff:
@julianguyen What else needs to be done on this issue? Unit tests for setCsrfToken and renderContent?

julianguyen:
@lucas-codes-stuff that is correct!

camillof:
Looks like this one is already covered here? https://github.com/ifmeorg/ifme/commit/893d43301601d7fa9b35b62f23ddb4c7f9fcc53d 🤔 ?

https://github.com/ifmeorg/ifme/blob/main/client/app/utils/__tests__/index.spec.jsx

I can write better tests for `randomString` but otherwise, tests looks good.

cc @julianguyen

PR #2181, opened 2022-10-19
added tests for randomString and getPusher

saminarp:
# Description

I have written unit tests  for the `client/app/utils`. The coverage report for the `index.js` has gone up to 65% with the corresponding tests I have written in `index.test.js`

## More Details

While I was able to cover all the lines for the function `randomString`,  the functions `setCsrfToken`, `getPusher` uses DOM implementation which makes it very tricky to mock it in Jest. Thus, they remain uncovered.

## Corresponding Issue
https://github.com/ifmeorg/ifme/issues/2128

# Screenshots

CleanShot 2022-10-19 at 03 55 51@2x

---


client/app/utils/__tests__/index.test.js
```diff
@@ -0,0 +1,35 @@
+import { randomString, setCsrfToken, getPusher } from '../index';
+import { JSDOM } from 'jsdom';
+
+
+describe('randomString, setCsrfToken, getPusher tests ', () => {
```
julianguyen, line 5: I think a better way to organize this is to structure the tests like so:

```
describe('Utils', () => {
  describe('randomString()`, () => {
    // etc.
  });
  
  describe('setCsrfToken()`, () => {
    // etc.
  });
  
  describe('getPusher()`, () => {
    // etc.
  });
  
   describe('renderContent()`, () => {
    // etc.
  });
});
```

client/app/utils/__tests__/index.test.js
```diff
@@ -0,0 +1,35 @@
...
+  it('should return random string', () => {
+    expect(typeof randomString()).toBe('string');
+  });
+
+
+  it('should be equal to pusher key and cluster when getPusher is called', () => {
```
julianguyen, line 11: I think you'll also want to test the `return null` case as well, when `window.pusher` does into exist.

julianguyen (changes requested):
Great work so far! Thanks for taking this on.

Looks like there's some ESLint failures in the CI build. You'll want to fix them. I recommend running `yarn lint:eslint` locally to reproduce the failures and verify when you've fixed them.

saminarp:
Hi @julianguyen  I've done the changes, please review ☺️

client/app/utils/index.js
```diff
@@ -3,22 +3,18 @@ import axios from 'axios';
 import renderHTML from 'react-render-html';
 import { sanitize } from 'dompurify';
 
-const randomString = (): string => Math.random()
-  .toString(36)
-  .substring(2, 15)
-  + Math.random()
-    .toString(36)
-    .substring(2, 15);
+export const randomString = (): string => Math.random().toString(36).substring(2, 15)
```
julianguyen, line 6: I don't think it's necessary to export these functions individually in order to be able to test them. You should be able to import the `Utils` object and call the functions as stated in the previous comment.

client/app/utils/__tests__/index.test.js
```diff
@@ -0,0 +1,21 @@
+const { randomString, getPusher } = require('../index');
```
julianguyen, line 1: Let's use imports here instead, I recommend just pulling in the Utils object instead of adding new exports for each function.

```suggestion
import Utils from '../index';
```

Then you can call `Utils.randomString()` and `Utils.getPusher()`.

julianguyen (changes requested):
Thanks for the updates, the tests looks good! 🎉 

Just a few comments!

julianguyen (approved):
Great work! Thanks for working on this 🎊

Issue #2141, opened 2022-10-05
Support ability to embed external images in Moments and Strategies editor


julianguyen:
# Description

We originally had [an issue in progress](https://github.com/ifmeorg/ifme/issues/1481) to support image uploads in these editors. Due to limited funds, we realistically can't afford to pay for potentially, costly Cloudinary (image upload service we use) usage. As a bare minimum, we should support being able to embed external images in Moments and Strategies posts.

We'll want to add this feature to our [InputTextarea component](http://design.if-me.org/?path=/docs/components-input--input-textarea) ([code](https://github.com/ifmeorg/ifme/blob/main/client/app/components/Input/InputTextarea.jsx)). We use an open source library called [pell](https://github.com/jaredreich/pell) which should have that feature.

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

Syed-Ansar:
Hey @julianguyen Can you help me out to self assign issues. I am not able to get self assigned, as new to open-source.

julianguyen:
@Syed-Ansar Looks like you've been added as a member of org, so you should be able to see the following in the right-hand navigation bar

![image](https://user-images.githubusercontent.com/3010728/194091608-40aaefb7-4ef1-4bcc-a749-1f087de45f15.png)

Let me know if you can't!

PR #2205, opened 2022-11-19
Embed external images in Moments and Strategies editor

Rxthew:
# Description

Implemented changes to allow editor to embed an image after the user provides an appropriate link via prompt.

## More Details

This was implemented to conform with the rest of the workflow which uses [pell](https://github.com/jaredreich/pell). 

One concern I have is that behind the scenes  pell relies heavily on the [document.execCommand ](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) property which is deprecated.

This is also highlighted as an [issue](https://github.com/jaredreich/pell/issues/199) in the repository itself.

## Corresponding Issue

[Issue 2141](https://github.com/ifmeorg/ifme/issues/2141)

# Screenshots

Peek Issue 2141

---

### DIFF
--- a/app/assets/stylesheets/core/containers.scss
+++ b/app/assets/stylesheets/core/containers.scss
@@ -11,3 +11,9 @@
   width: 100%;
   margin: 0;
 }
+
+.imageContainer {
+  img {
+    max-width: 100%;
+  }
+}
--- a/app/views/moments/show.html.erb
+++ b/app/views/moments/show.html.erb
@@ -20,7 +20,7 @@
 </div>
 
 <% if @moment.why.present? %>
-  <div class="smallMarginTop">
+  <div class="smallMarginTop imageContainer">
     <% if @moment.fix.present? %>
       <div class="label">
         <%= label_tag t('moments.form.why_legacy') %>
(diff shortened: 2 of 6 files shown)


julianguyen (changes requested):
Great start, thanks so much for working on this 🎉  

I pulled down your branch and the functionality works as expected!

A few things we need to address are making sure very large images are scaled to fix the container or else this will happen:

In the editor:

image

On the post page:

image

Through CSS, we want to set `max-width: 100%;`.

Rxthew:
> Great start, thanks so much for working on this 🎉
> 
> I pulled down your branch and the functionality works as expected!
> 
> A few things we need to address are making sure very large images are scaled to fix the container or else this will happen:
> 
> In the editor:
> 
> image
> 
> On the post page:
> 
> image
> 
> Through CSS, we want to set `max-width: 100%;`.

Hello,

Sorry this took so long, I've completed it this a while ago but I'm still experiencing issues with my dev environment.

I spent some time getting more familiar with the repository and I realised there was no dedicated component for posts so I added a class in the ERB files. This involved importing the containers from the core assets module into the global scss file, and I was a little uncomfortable with that so If there is a less disruptive approach I may have overlooked I could go back and make changes.

julianguyen (approved):
Thanks the updates!

PR #2241, opened 2023-07-24
Make the header bar component stick to top

thatsthat:
# Description

1. Created two additional div nodes in order to apply CSS position: sticky property to the header component. 
2. Added some extra css properties to the header component div container.
3. Created an additional hook named stickyHeader which is invoked from the header component and contains the logic to detect scrolling to change the background color of the header to black when user scrolls down and restores its original color if user scrolls up to the top.

## Corresponding Issue

[Issue 2155](https://github.com/ifmeorg/ifme/issues/2155)

# Screenshots
chrome-capture-2023-6-24

---

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -54,6 +54,16 @@ html {
   }
 }
 
+#headerContainer {
+  position: sticky;
+  top: 0;
+  z-index: 1000;
+}
+
+.stickyHeader {
+  background-color: $black-90;
+}
+
 .content {
   margin: 0 auto;
   padding: 0;
--- a/client/app/hooks/stickyHeader.js
+++ b/client/app/hooks/stickyHeader.js
@@ -1,0 +1,28 @@
+// @flow
+export const stickyHeader = () => {
+  // Check when element gets position sticky
+  var observer = new IntersectionObserver(
+    function (entries) {
+      // no intersection
+      if (
+        entries[0].intersectionRatio === 0 &&
+        // Filter false positives due to page not rendered
+        entries[0].rootBounds !== null
+      ) {
+        console.log(entries[0]);
+        document
+          .querySelector('#headerContainer')
+          .classList.add('stickyHeader');
+      }
+      // fully intersects
+      else if (entries[0].intersectionRatio === 1) {
+        console.log(entries[0]);
+        document
+          .querySelector('#headerContainer')
+          .classList.remove('stickyHeader');
+      }
+    },
+    { threshold: [0, 1] }
+  );
+  observer.observe(document.querySelector('#headerContainerTop'));
+};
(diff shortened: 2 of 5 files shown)


julianguyen:
Hey @thatsthat can you rebase from main? I recently merged in [some changes](https://github.com/ifmeorg/ifme/pull/2243) to fix the broken CI build! Thanks!

app/assets/stylesheets/application.scss
```diff
@@ -54,6 +54,16 @@ html {
   }
 }
 
+#headerContainer {
+  position: sticky;
+  top: 0;
+  z-index: 1000;
```
julianguyen, line 60: We want to be deliberate with the `z-index` values we pass in! Let's put the value in `app/assets/stylesheets/base/_values.scss` and use a variable instead! Thanks!

app/assets/stylesheets/application.scss
```diff
@@ -54,6 +54,16 @@ html {
   }
 }
 
+#headerContainer {
```
julianguyen, line 57: These style changes should be made inside `client/app/components/Header/Header.scss` instead, and should make these changes from inside of `Header` component in the React front-end!

client/app/hooks/stickyHeader.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+export const stickyHeader = () => {
```
julianguyen, line 2: If we implement this sticky header completely in React, then we don't need to be making these DOM manipulations to toggle the sticky header.

julianguyen (changes requested):

client/app/hooks/stickyHeader.js
```diff
@@ -0,0 +1,28 @@
...
+  // Check when element gets position sticky
+  var observer = new IntersectionObserver(
+    function (entries) {
+      // no intersection
+      if (
+        entries[0].intersectionRatio === 0 &&
+        // Filter false positives due to page not rendered
+        entries[0].rootBounds !== null
+      ) {
+        console.log(entries[0]);
```
julianguyen, line 12: We should remove this since it's used for debugging purposes!

julianguyen:
Thanks so much for taking this on 🎉 Let me know if you have any questions :)

app/assets/stylesheets/application.scss
thatsthat: I originally tried doing just that, but found out that in order for css position: sticky to work, it needs an parent element to the Header component to whom it can 'stick to', and therefore some changes going up in the DOM tree are needed. I chose 'app/views/layouts/application.html.erb' and 'app/assets/stylesheets/application.scss' but it could be any ancestor of the Header component. What do you think?

julianguyen: I think you can just setup this container element from inside the React component `client/app/components/Header/index.jsx` as a parent container, instead of doing it on the Rails view side.

client/app/hooks/stickyHeader.js
thatsthat: I tried just doing that, and didn't manage to make it work. The reason, I think, is that React on Rails inserts each react component as an independent div node, preventing its CSS style sheet any access to its ancestor elements in the DOM. Moreover, the intersection observability api which is used to change the color when scrolling down, also needs access to an ancestor element to the header.

I am open to any suggestion that would allow only modifying the react header component code, but I'm afraid that would require a different approach.

PR #2315, opened 2024-09-16
Fix user seeds not populating from `pwned_password` validation

bennpham:
# Description

Ignore validations when creating `User` in the seed file to match the sample users on https://github.com/ifmeorg/ifme/wiki/Installation#testing-accounts 

## More Details

Since my implementation of https://github.com/ifmeorg/ifme/pull/1935 , when the pwned password list database updates, more passwords will be checked to the data breach dump which prevents users from creating a password that's already been found in a breach.

Since these are just test users which you can find on a public Wiki page in plaintext, it doesn't matter that the password is `password` so I decide to ignore the validations so we can seed those users for testing purposes.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@
 .idea/
 .tags
 *.swp
+.tool-versions
 
 # Ignore coverage results
 /coverage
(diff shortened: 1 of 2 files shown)


julianguyen (approved):
Thank you Ben!!

PR #2325, opened 2024-10-15
Update `renderContent` to accept attributes

### LINKED ISSUE #2120: Add ARIA attribute to element that opens the Tooltip component

camillof:
# Description

Updates `renderContent` utils method, so it can now accept (optional) attributes, that will be attached to the provided React element

## More Details

This doesn't include `string` content, because the result of `renderHTML(sanitize(content));` could be an array of elements, so in that case we have a few options:
A. apply the attributes to all the elements
B. wrap the elements in a `div` and apply the attributes to that div (but again, we would be in the scenario of applying invalid attributes)
C. apply the attributes to the first element

## Corresponding Issue

Resolves #2120

# Screenshots

image
image

---

### DIFF
--- a/client/app/components/Tooltip/index.jsx
+++ b/client/app/components/Tooltip/index.jsx
@@ -26,8 +26,8 @@ export const Tooltip = (props: Props): Node => {
   const tooltipId = id || `tooltip${Utils.randomString()}`;
   return (
     <div className={css.tooltipWrapper}>
-      <div className={css.tooltipElement} aria-labelledby={tooltipId}>
-        {Utils.renderContent(element)}
+      <div className={css.tooltipElement}>
+        {Utils.renderContent(element, { 'aria-labelledby': tooltipId })}
       </div>
       <div
         id={tooltipId}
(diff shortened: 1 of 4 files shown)


julianguyen (approved):
Looks good, thank you!

Issue #2300, opened 2024-07-07
[FEAT]: Add the ability to create a new Strategy or Moment from an individual Category page


julianguyen:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is there a pain point in the app driving this feature request?

Adds another way to be able to quickly tag a Category in a Strategy or Moment post.

### Desired Behavior

1. Go to the `https://www.if-me.org/categories/category-name-slug` page
2. Add two buttons "New Moment" and "New Strategy" aligned right of the Category name heading 

Similar to `/moments` page:
![image](https://github.com/ifmeorg/ifme/assets/3010728/bc8eac89-cd21-4d7a-b2b2-098eee5da7de)

3. When clicking these buttons, the Category that was entered from will be automatically tagged in the form.

![image](https://github.com/ifmeorg/ifme/assets/3010728/59a459e3-4b3a-4483-b3e9-9afcf3f65270)

![image](https://github.com/ifmeorg/ifme/assets/3010728/1bd02b61-f94d-45e0-94fb-77160eef18bb)

It's recommend that you pass the category slug through url params.

### Criteria for Completion

- [ ] Use components from http://design.if-me.org
- [ ] Responsive - works in mobile, tablet, and desktop breakpoints

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the if-me community

yumi520:
Hi @julianguyen ! I would like to be assigned this task! Thank you!

Hachondeoro:
@julianguyen I would like to complete this task. Please assign it to me as I see it is still open.

julianguyen:
Feel free to take it on @Hachondeoro

Jagss24:
@Hachondeoro have you completed these task yet or can I pick the task?

patrikkr858:
Hi @julianguyen, is this task completed? I would like to take a look at it, if it's possible. 😏

LucasKay64:
According to our conversation via Email and Slack with @julianguyen, i am giving this issue a go.

Ghanasree:
Hi @julianguyen. Can I work on this issue?

LucasKay64:
Hi @Ghanasree, I am actively working on the issue. Got postponed a little but its in the process.

Ishpreet01:
Hi, Can you assign this issue to me? I  am new to open source and would like to work on this.

Issue #2115, opened 2022-09-30
Update to the latest version of React (v18)


julianguyen:
# Description

This requires updates to the following packages:

* Gems: react_on_rails
* npm packages: React (and related dependencies) and react-on-rails 

---

Please assign yourself (via the Assignees dropdown), if you do want to work on this issue. Can't find yourself? You need to [join our organization](https://github.com/ifmeorg/ifme/wiki/Join-Our-Slack).

Check out our [Picking Up Issues](https://github.com/ifmeorg/ifme/wiki/Picking-Up-Issues) guide if you haven't already!

Im-Kunal-13:
@julianguyen can you assign me this issue? I would like to work on it.

julianguyen:
Hey @Im-Kunal-13! I just sent you an invitation to join the GitHub org, so you can assign yourself to the ticket. If you want to join the Slack, please follow that "next steps" link in the welcome bot message above. Thanks!

toshitapandey:
@julianguyen - Was this task done? If not, I'd like to pick it up.

sbpipb:
yes you can work on this @toshitapandey

julianguyen:
@toshitapandey Feel free to take this on!

toshitapandey:
@julianguyen @sbpipb - Thank you! Picking it.

PR #2330, opened 2024-10-16
WIP: Add the ability to create a new Strategy or Moment from an individual Category page

lenikadali:
# Description

Adds the ability to create a new Strategy or Moment from an individual Category page.

## Corresponding Issue

#2300

### DIFF
--- a/app/helpers/categories_helper.rb
+++ b/app/helpers/categories_helper.rb
@@ -56,6 +56,11 @@ module CategoriesHelper
     }
   end
 
+  def category_input_props(field, type, label, group = false)
+    { id: "moment_#{field}", type:,
+      name: "moment[#{field}]#{group ? '[]' : ''}", label: t(label) }
+  end
+
   def category_form_inputs
     [
       category_name(@category&.name),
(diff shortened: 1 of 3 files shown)


julianguyen:
Hey @lenikadali do you need any help here?

lenikadali:
> Hey @lenikadali do you need any help here?

Hi 👋🏿 

Thanks for checking in here 🙌🏿  The plan is to try out your suggestions either today or tomorrow and see how it goes 🤞🏿

lenikadali:
> So the view is being rendered https://github.com/ifmeorg/ifme/blob/main/app/views/categories/new.html.erb. 
> So you’ll want to do it in the form that is being rendered with React. So you’ll need to update the value of `new_category_props` on the Rails side.
I've attempted to do that in this commit [8b9f8e2](https://github.com/ifmeorg/ifme/pull/2330/commits/8b9f8e2030992aefe6dacfedb30744e991457759)

> You’ll want to do something like the Moments form where you can add a new Strategy from in there.
> This isn’t about adding a link to the index page.

Oh. Then what did this statement mean in the issue description 🤔?

> Desired Behavior
>
> 1. Go to the https://www.if-me.org/categories/category-name-slug page
> 2. Add two buttons "New Moment" and "New Strategy" aligned right of the Category name heading

Also, I noticed that the documentation for the version used of React on Rails is now [not easily accessible](https://github.com/shakacode/react_on_rails#news). Maybe it could be worth it to first upgrade React on Rails as per [this ticket](https://github.com/ifmeorg/ifme/issues/2115) and then attempt this one since it would be easier to reference the documentation as we work on this.

julianguyen:
Ohhh sorry I think I misunderstood / forgot about this ticket was asking for even though I wrote it haha.

Yes so what you want to do is modify the form for the `show.html.erb` page. When you click "New Moment" or "New Strategy" it will automatically have that category pre-checked in their respective `new.html.erb` forms.

julianguyen:
Yeah, we should upgrade React on Rails! It doesn't block this work, but I agree! There's a ticket to do that already with the latest version of React. https://github.com/ifmeorg/ifme/issues/2115

julianguyen:
Closing this since this was worked on in https://github.com/ifmeorg/ifme/pull/2337

PR #2337, opened 2024-11-25
2300 tagging strategy or moment from category page

LucasKay64:
# Description

Added the ability to create a new strategy or moment from a concrete category page with tagging of the related category in the form for creating that new strategy or moment.

## More Details

Added buttons aligned right to the title. They take the user to the form for creating a new strategy or moment. The tagged category from which the user clicked the buttons is stored in the URL query string and is read by the corresponding form. Also added some styling for mobile responsiveness.

The new @page_new_buttons handles multiple buttons for any title on any page for further use in the future although if one adds too many buttons, new CSS media query will be required to add as well.

# Screenshots

image
image
image

# Corresponding Issue
#2300

### DIFF
--- a/app/helpers/moments_form_helper.rb
+++ b/app/helpers/moments_form_helper.rb
@@ -122,7 +122,12 @@ module MomentsFormHelper
   def data_for(item)
     case item.class.name
     when 'Category'
-      @moment.categories.pluck(:id)
+      ids = @moment.categories.pluck(:id)
+      if params[:category].present? && ids.empty?
+        category = Category.friendly.find_by(slug: params[:category])
+        ids << category.id if category
+      end
+      ids
     when 'Mood'
       @moment.moods.pluck(:id)
     when 'Strategy'
--- a/app/helpers/strategies_form_helper.rb
+++ b/app/helpers/strategies_form_helper.rb
@@ -106,11 +106,17 @@ module StrategiesFormHelper
   def category_checkboxes
     checkboxes = []
     @categories.each do |item|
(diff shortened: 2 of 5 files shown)


app/helpers/moments_form_helper.rb
```diff
@@ -122,7 +122,12 @@ def checkboxes_for(data)
   def data_for(item)
     case item.class.name
     when 'Category'
-      @moment.categories.pluck(:id)
+      ids = @moment.categories.pluck(:id)
+      if params[:category].present? && ids.empty?
+        category = Category.friendly.find_by(slug: params[:category])
+        ids << category.id if category
+      end
+      ids
```
julianguyen, line 130: Can you write unit tests for this?

app/helpers/strategies_form_helper.rb
```diff
@@ -106,11 +106,17 @@ def build_switch_input(value, checked, unchecked_value)
   def category_checkboxes
     checkboxes = []
     @categories.each do |item|
+      checked = @strategy.categories.include?(item)
+
+      if params[:category].present? && @strategy.new_record? && !checked
+        checked = item.slug == params[:category]
```
julianguyen, line 112: Can you write unit tests for this?

julianguyen (changes requested):
Great work! I think the last thing we want to do before we merge is write tests! I pointed places you can write unit tests in the helpers. And I also think we could write E2E tests as well.

julianguyen (approved):
The next PR will address tests.

PR #2338, opened 2024-11-26
refactor allies controller to use AR query methods

hendrixfan:
# Description

In my opinion, `.sort_by!` outsources the work of sorting the dataset unnecessarily to ruby. It might make more sense to keep the ActiveRecord context so that `.order` and `.exists?` can be performed on the DB-Level.


hendrixfan:
Sorry my mistake. It turns out `allies_by_status` is a method that returns an array.

Issue #2357, opened 2025-03-12
[FEAT]: Convert all JPG/PNG images to WebP


julianguyen:
### Is there an existing request for this feature?

- [x] I have searched the existing issues

### Is there a pain point in the app driving this feature request?

In an effort to reduce asset sizes and improve runtime performance, let's convert all the JPG/PNG images in the app to WebP.

### Code of Conduct

- [x] I've read the Code of Conduct and understand my responsibilities as a member of the if-me community

Sriram-7451:
Hey @julianguyen 
if this issue is not yet fixed then can i take up this and do this for you?

julianguyen:
@Sriram-7451 Go right ahead!

Sriram-7451:
@julianguyen Can you give access to me?

Sriram-7451:
@julianguyen can you please give access to me?
I'm done with the resolve for the issue I just need to push the code

Sub-codes:
If this issue is not fixed I would love to contribute

Sub-codes:
@julianguyen is this still open

pr-iti:
Is this still ope... I would love to  solve this issue

Nithin-cholleti:
is this still open

julianguyen:
A [PR](https://github.com/ifmeorg/ifme/pull/2360) was open here, @Nithin-cholleti or @pr-iti feel free to help out with it. The contributor who put it out hasn't responded to my feedback request!

MiguelNavarrete-dev:
Continue open this [PR](https://github.com/ifmeorg/ifme/issues/2357l)?

PR #2360, opened 2025-03-15
feat: Changed images from jpg/png to webp

Sriram-7451:
PR is raised for the resolve for the issue #2357 

Changed the Images from JPG/PNG to WebP

Conversion was done through conversion tool available in the internet and was able to reduce the asset storage.


julianguyen:
Ooh so exciting, thanks for the fast follow!

I would keep our favicon image as an `.ico` file. I think it can be in `.webp` but not all browsers may support it. So you can at an HTML tag in our header for the `.webp` version but also keep the current `.ico` version.

Sriram-7451:
@julianguyen Added both images actually ico and webp for favicon in header

julianguyen:
Looks like there's remaining updates that need to be made:

```
/home/circleci/ifmeorg/ifme/client/app/stories/Avatar.stories.jsx
  5:19  error  Unable to resolve path to module 'app/assets/images/contributors/ingrid_garcia.jpg'  import/no-unresolved

/home/circleci/ifmeorg/ifme/client/app/stories/Modal.stories.jsx
  5:19  error  Unable to resolve path to module 'app/assets/images/contributors/ingrid_garcia.jpg'  import/no-unresolved
  ```

PR #2368, opened 2025-04-24
Implementing eager loading on the MomentsController

mirjharr2:
# Description
This change will implement eager loading into the `set_association_variables!` function in `MomentsController` to reduce queries when creating and editing a moment. 

**Key Changes**
- Implementing eager loading using `.includes(...)` to preload moments associations when retrieving associated categories and moods.
- Addition of `visible_categories` and `extra_categories` to avoid a false.include error in the case of a null response.
- Addition of `visible_moods` and `extra_moods` to avoid a false.include error in the case of a null response.

## More Details

This refactor is important because:
- The current code results in an **N+1 query issue** which can create performance issues under heavy load because the database will be overloaded with queries when retrieving associated categories and moods.
- This change will make the database more scalable than it is now.
- Significantly reduces the number of queries to two queries per category and mood that is fetched, opposed to N+1.

---


julianguyen (approved):
Solid work, Miranda!

PR #2445, opened 2026-05-13
chore: remove redundant polyfills from webpack entry point

renanfochetto:
## Description
Remove redundant polyfills (`@babel/polyfill` and `es5-shim`) from the webpack entry point and from `package.json`.

These polyfills were redundant because `@babel/preset-env` with `useBuiltIns: 'usage'` and `core-js@3` already handles polyfills automatically — injecting only what is needed based on the target browsers (`defaults` browserslist).

Additionally, `@babel/polyfill` has been officially deprecated in favor of this approach, and `es5-shim` was designed for IE8/IE9 support which is not required given the current browserslist configuration.

This change reduces the JavaScript bundle size by approximately 1.3 MiB (~10% reduction).

## Type of Change
Improvement

## Related to
Not related to any issues

## How to test
Run the test suite and verify all tests pass:
`yarn test`

Verify the bundle size reduction by running:
`yarn build:production`

## Checklist:
- [x] Tested locally
- [ ] Added tests
- [x] Lint passed
- [x] Followed style guide

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,9 @@ node_modules
 
 # Ignore gems installed locally
 /vendor
+
+# Contributing RoadMap Portuguese
+CONTRIBUTION_ROADMAP-pt.md
+
+# Package lock (project uses yarn)
+client/package-lock.json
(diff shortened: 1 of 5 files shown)


julianguyen (approved):
Solid work!

PR #2451, opened 2026-05-25
Add Traditional Chinese zh-TW locale

PeterDaveHello:
# Description

This pull request adds support for Traditional Chinese (`zh-TW`) as a new locale throughout the application :tada:

### DIFF
--- a/config/locales/ca.yml
+++ b/config/locales/ca.yml
@@ -30,6 +30,7 @@ ca:
     fr: Francès
     hi: Hindi
     zh-CN: Xinès
+    zh-TW: Xinès tradicional
     ko: Coreà
     id: Indonesi
     ar: Àrab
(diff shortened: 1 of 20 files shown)


config/locales/ca.yml
```diff
@@ -30,6 +30,7 @@ ca:
     fr: Francès
     hi: Hindi
     zh-CN: Xinès
+    zh-TW: Xinès tradicional
```
julianguyen, line 33: Is the tradicional part correct?

julianguyen:
Thanks so much for adding Traditional Chinese!

config/locales/ca.yml
PeterDaveHello: Thanks for the review!

I actually don't speak Catalan, but I double-checked the translation with multiple frontier AIs (GPT, Claude, Grok, etc.) and they all confirm that **"Xinès tradicional"** is the standard and correct term for Traditional Chinese.

Let me know if you'd like me to change anything!

PR #2457, opened 2026-06-03
set up devcontainer

geppy:
# Description

Configure devcontainer using existing Docker Compose.

## More Details

This is intended to make it easier to contribute to the project.  With a few changes, this'd make it possible to run in GitHub Codespaces, simplifying contribution even further.

---

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,10 +18,6 @@
     }
   },
 
-  "features": {
-    "ghcr.io/devcontainers/features/github-cli:1": {}
-  },
-
   "postCreateCommand": "bash -lc 'set -e; for f in development test; do src=config/env/${f}.example.env; dst=config/env/${f}.env; [ -f \"$dst\" ] || cp \"$src\" \"$dst\"; done; bundle check || bundle install; (cd client && yarn install); bundle exec rake db:prepare; if [ -z \"${DEVCONTAINER_SKIP_SEED:-}\" ]; then bundle exec rake db:seed || true; fi'",
 
   "customizations": {
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,68 @@
+{
+  "name": "ifme (Rails 7 + React on Rails)",
+
+  "dockerComposeFile": [
+    "../docker-compose.yml",
+    "docker-compose.devcontainer.yml"
+  ],
+  "service": "app",
+  "workspaceFolder": "/app",
+  "remoteUser": "root",
+  "shutdownAction": "stopCompose",
+
+  "forwardPorts": [3000],
+  "portsAttributes": {
+    "3000": {
+      "label": "Rails (puma)",
+      "onAutoForward": "notify"
+    }
+  },
+
+  "features": {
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,68 @@
...
+    "3000": {
+      "label": "Rails (puma)",
+      "onAutoForward": "notify"
+    }
+  },
+
+  "features": {
+    "ghcr.io/devcontainers/features/github-cli:1": {}
+  },
+
```
geppy: Allowing the `.app.github.dev` domain would unblock GitHub Codespaces hosting.
```suggestion
  "containerEnv": {
    "RAILS_DEVELOPMENT_HOSTS": ".app.github.dev"
  },
  
```

This change is enough on its own, but doesn't address the use of `localhost:3000`  in URLs.

julianguyen: What's the action here? Should we have both?

geppy: I just meant to here specify that setting `RAILS_DEVELOPMENT_HOSTS` isn't enough without updating how
[`primary_domain`](https://github.com/ifmeorg/ifme/blob/2ad1ea6748bedfe5ecd4cb4a2130e58f18ca0b6d/config/environments/development.rb#L4) is used.  If you think that's worth doing I'll expand the PR.

Now that you mention it I might actually strip out the `gh` bits since my lockfile was getting mutated.  If someone really wants it they can add it.

geppy: A local devcontainer works without any changes, but running on GitHub's servers it won't be at `localhost:3000`, it'll be at something-something-blah`-3000.app.github.dev` and (e.g.) asset URLs will be broken, as below:

image

geppy: If you want to try this on your own fork, add this branch to your fork and open it in a codespace.  Then you can run it with:
```sh
RAILS_DEVELOPMENT_HOSTS=".app.github.dev" bundle exec foreman start -f Procfile.dev
```

julianguyen: Ahh so adding a value for `primary_domain` will allow the bundles to include the compiles SCSS?

geppy: Yes.  I added some commits showing one way we could do it.

config/environments/development.rb
```diff
@@ -98,6 +105,5 @@
 
   config.force_ssl = false
 
-  config.action_controller.default_url_options = { host: primary_domain }
-  config.action_controller.asset_host = primary_domain
```
geppy, line 102: Removing `asset_host` causes root-relative URLs to be used.

config/environments/development.rb
```diff
@@ -1,7 +1,14 @@
 # frozen_string_literal: true
 require "active_support/core_ext/integer/time"
 
-primary_domain = 'localhost:3000'
```
geppy, line 4: This isn't as readable as it was before, but makes it work out of the box in codespaces.  Up to you if you think it's worth it.

julianguyen (approved):
Looks great to me, thanks for looking into this! I didn't know about devcontainer!

PR #2474, opened 2026-06-22
[#2322] Add language selector to navbar

### LINKED ISSUE #2322: Change the location of  language menu
image
image

I think the standard practice is to place the language menu on the right side of the navbar. This positioning aligns with common design patterns, making it easier for visitors to locate and switch languages quickly.

kenlacroix:
# Description

Surfaces the language switcher in the navbar so users no longer have to scroll to the footer to change languages. The existing `ToggleLocale` widget is reused as-is and rendered in two new places via the `Header` component:

- **Desktop:** right-aligned in the nav bar (after the links).
- **Mobile:** inside the hamburger menu.

## More Details

`locale`/`locales` are threaded through `header_props`; they were already assigned for every request in `ApplicationController` (the footer uses the same values), so no controller or routing changes are needed. The selector only renders when a current locale and available locales are present, so existing `Header` usages are unaffected. A `WithLanguageSelector` story is added to the existing Header stories so the new state is visible in Storybook.

No new gems/npm packages are introduced — this reuses the existing `ToggleLocale` widget.

The footer toggle is kept in place as well. Happy to **remove it from the footer** if you'd prefer the navbar to be the single location — that felt like a product call for you to make rather than assume.

**Testing:** `Header.spec.jsx` passes 7/7 (4 existing + 3 new, covering desktop render, no-render without locales, and render inside the open mobile menu). ESLint, Flow, Prettier, and Stylelint are clean on the changed files. I was unable to run the full Ruby/Rails suite or RuboCop locally (no Ruby toolchain in my environment) — happy to address anything CI flags.

## Corresponding Issue

Closes #2322

# Screenshots

Captured from the `Components/Header` Storybook story (mulberry background).

**Desktop** — selector right-aligned after the nav links:

Language selector in desktop navbar

**Mobile** — selector inside the open hamburger menu:

Language selector in mobile menu

### DIFF
--- a/client/app/components/Header/Header.scss
+++ b/client/app/components/Header/Header.scss
@@ -57,6 +57,24 @@
     }
   }
 
+  &Locale {
+    @include setMargin($size-0, $size-0, $size-0, $size-26);
+
+    display: flex;
+    align-items: center;
+
+    @media screen and (max-width: $medium) {
+      display: none;
+    }
+  }
+
+  &MobileLocale {
+    @include setMargin($size-20, $size-0, $size-0, $size-0);
+
+    display: flex;
+    justify-content: center;
+  }
+
   &Scroll {
     background-color: $black-80;
 
@@ -125,6 +143,10 @@
     }
 
     &Nav {
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+
       &Links {
         @include setFontSize($size-18);
 
--- a/client/app/components/Header/Header.scss
+++ b/client/app/components/Header/Header.scss
@@ -66,6 +66,10 @@
     @media screen and (max-width: $medium) {
       display: none;
     }
+
+    select {
+      @include setPadding($size-12, $size-0, $size-12, $size-24);
+    }
   }
 
   &MobileLocale {
@@ -73,6 +77,10 @@
 
     display: flex;
     justify-content: center;
+
+    select {
+      text-align: center;
+    }
   }
 
   &Scroll {
--- a/client/app/components/Header/__tests__/Header.spec.jsx
+++ b/client/app/components/Header/__tests__/Header.spec.jsx
@@ -117,7 +117,7 @@ describe('Header', () => {
     await user.click(container.querySelector('#headerHamburger'));
 
     expect(
-      container.querySelector('#headerMobile #locale'),
+      container.querySelector('#headerMobile #mobileLocale'),
     ).toBeInTheDocument();
   });
 });
--- a/client/app/components/Header/__tests__/Header.spec.jsx
+++ b/client/app/components/Header/__tests__/Header.spec.jsx
@@ -90,4 +90,34 @@ describe('Header', () => {
     await user.keyboard('{Shift>}{Tab}{/Shift}{Enter}');
     expect(container.querySelector('#headerMobile')).not.toBeInTheDocument();
   });
+
+  const componentWithLocales = (
+    <Header
+      home={{ name: 'Home', url: '/some-path' }}
+      links={[{ name: 'Link 1', url: '/some-path-one' }]}
+      locale="en"
+      locales={['en', 'es']}
+    />
+  );
+
+  it('renders the language selector when locales are provided', () => {
+    const { container } = setup(componentWithLocales);
+
+    expect(container.querySelector('#locale')).toBeInTheDocument();
+  });
+
+  it('does not render the language selector without locales', () => {
+    const { container } = setup(component);
+
+    expect(container.querySelector('#locale')).not.toBeInTheDocument();
+  });
+
+  it('renders the language selector inside the open mobile menu', async () => {
+    const { container, user } = setup(componentWithLocales);
+    await user.click(container.querySelector('#headerHamburger'));
+
+    expect(
+      container.querySelector('#headerMobile #locale'),
+    ).toBeInTheDocument();
+  });
 });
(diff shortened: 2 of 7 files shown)


julianguyen:
Make sure to update the tests: https://app.circleci.com/pipelines/github/ifmeorg/ifme/3012/workflows/41e5e902-4eaf-4532-8d56-3af44b35d6fe/jobs/18102

client/app/components/Header/Header.scss
```diff
@@ -57,6 +57,24 @@
     }
   }
 
+  &Locale {
```
julianguyen, line 60: We now want the height of the desktop navigation header to be the same as the one when you scroll down the page


Image

Image

client/app/components/Header/Header.scss
```diff
@@ -57,6 +57,24 @@
...
+
+    display: flex;
+    align-items: center;
+
+    @media screen and (max-width: $medium) {
+      display: none;
+    }
+  }
+
+  &MobileLocale {
```
julianguyen, line 75: Let's also center align the locale dropdown text as well!

Image

julianguyen (changes requested):
Thanks for working on this!

client/app/components/Header/__tests__/Header.spec.jsx
```diff
@@ -90,4 +90,34 @@ describe('Header', () => {
...
+      links={[{ name: 'Link 1', url: '/some-path-one' }]}
+      locale="en"
+      locales={['en', 'es']}
+    />
+  );
+
+  it('renders the language selector when locales are provided', () => {
+    const { container } = setup(componentWithLocales);
+
+    expect(container.querySelector('#locale')).toBeInTheDocument();
```
julianguyen: Because there are now two locale dropdowns on the page, there are two values for `id="locale"`. You can only have one unique value per page; you will need to give each of these dropdowns a different value for id. And update the tests accordingly.

kenlacroix: Good catch — gave the two selectors distinct ids: the desktop one keeps `id="locale"` and the mobile one is now `id="mobileLocale"`, via a new optional `id` prop on `ToggleLocale` (defaulting to `locale`, so other usages are unaffected). Updated the mobile test to assert on `#headerMobile #mobileLocale`. Pushed in 1dc95407.

client/app/components/Header/Header.scss
kenlacroix: Trimmed the in-navbar locale `select`'s vertical padding so the desktop header keeps its normal height instead of growing to fit the dropdown (pushed in 1dc95407). Let me know if you want it tighter — happy to dial the height in further to match exactly.

client/app/components/Header/Header.scss
kenlacroix: Done — center-aligned the locale dropdown text (pushed in 1dc95407).

client/app/components/Header/__tests__/Header.spec.jsx
kenlacroix: Follow-up (`9682105d`): the failing CircleCI job was the **RSpec** suite, and digging in showed the footer already renders `ToggleLocale` (`id`/`name="locale"`), so the navbar desktop instance was still colliding with it. Fixed properly now — `ToggleLocale` derives `name` from the `id` prop, the desktop navbar instance is `navbarLocale` (mobile stays `mobileLocale`), so footer `locale` / navbar `navbarLocale` / mobile `mobileLocale` are all unique. This also un-breaks the `persisting_browser_locale_after_sign_in` feature spec (its `within 'select[name=locale]'` now matches only the footer again). Also updated `header_helper_spec` for the new `locale`/`locales` props.

kenlacroix:
Update: pushed `9682105d` addressing all three review threads and the CI failures.

The CircleCI **Run Rails RSpec Tests** job was failing for two reasons, both now fixed:
- `header_helper_spec.rb` needed the new `locale`/`locales` keys in its expected `header_props` hashes.
- The footer already renders `ToggleLocale` (`id`/`name="locale"`), so the navbar added a second `select[name=locale]`, which made the `change_language` feature helper's `within 'select[name=locale]'` ambiguous (and left a duplicate `id="locale"`). `ToggleLocale` now derives `name` from its `id` prop, and the instances use unique ids — footer `locale`, navbar `navbarLocale`, mobile `mobileLocale` — so the footer select is unambiguous again.

The latest run ([18127](https://circleci.com/gh/ifmeorg/ifme/18127)) is now down to a **single, unrelated failure**: `user_creates_a_published_strategy_spec.rb:16` expects `"Ally 0, Ally 1, and Ally 2"` but got `"Ally 2, Ally 0, and Ally 1"` — a non-deterministic ally-ordering assertion (seed 3799) untouched by this PR. Could you give CI a re-run to clear that flake? Happy to help if it turns out to be more than flakiness.

client/app/components/Header/__tests__/Header.spec.jsx
julianguyen: Was able to run again, and it is a flake unrelated to your changes! I'll file a separate issue to handle this flake!

client/app/components/Header/Header.scss
julianguyen: Thanks for the update, it looks good!

julianguyen (approved):
Thanks for working on this! It looks good to me!

PR #2489, opened 2026-07-15
Handle missing Pusher meta tags

adamjpena:
# Description

Updates `Utils.getPusher` to return `null` when the required Pusher meta tags are missing.

## More Details

The utils test file is also renamed from `.test.jsx` to `.spec.jsx` so it matches the project's Jest test pattern. This brings the existing coverage for the missing-meta case into the test suite and avoids an error when `window.Pusher` is present but the page has no Pusher configuration tags.

# Screenshots

Not applicable; utility/test change.


julianguyen (approved):
Thanks for this!

PR #2490, opened 2026-07-15
Remove legacy ESLint config

adamjpena:
# Description

Removes the legacy `.eslintrc` now that the client uses the ESLint flat config.

## More Details

The active config lives in `client/eslint.config.mjs`. This also drops the unused `eslint-plugin-flowtype` dev dependency and lockfile entry that were only referenced by the old config.

# Screenshots

Not applicable; config-only change.


julianguyen (approved):
Good cleanup!

PR #2512, opened 2026-09-19
Add skeleton loader for lazy-loaded charts

### LINKED ISSUE #2484: Improve page loading experience with skeleton loaders
### Is there an existing request for this feature?

- [x] I have searched the existing issues

### Is there a pain point in the app driving this feature request?

When navigating between pages or refreshing the application, the current loading state is represented only by a thin progress bar at the top of the page.

On slower network connections, this provides limited visual feedback and users may not immediately understand that the next page is still loading, especially when the page content takes a moment to appear.

### Desired Behavior

Provide a more informative loading experience by displaying skeleton loaders (or another appropriate loading state) for pages that load asynchronous content.

This would give users a visual preview of the page layout while content is being fetched and improve the perceived responsiveness of the application.

### Criteria for Completion

- [ ] Identify pages where content is loaded asynchronously.
- [ ] Display an appropriate loading state while data is being fetched.
- [ ] Replace the loading state with the actual content once loading is complete.
- [ ] Ensure the loading indicator follows the existing design system and is accessible.
- [ ] Verify there are no layout shifts or visual glitches during loading.

### Anything else?

Skeleton loaders are commonly used to improve perceived performance in modern web applications because they communicate that content is actively loading rather than leaving users with minimal feedback.

I'm happy to work on this enhancement if it aligns with the project's design goals.

### Code of Conduct

- [x] I've read the Code of Conduct and understand my responsibilities as a member of the if-me community

jamalkamaladdin:
# Description

Adds a `Skeleton` component and uses it as the `Suspense` fallback of the lazy-loaded `Chart`. The page no longer shifts when the chart code arrives.

## More Details

Since #2475 the chart loads with `fallback={null}`: it takes 0px, then pushes the content below it down 300px. The skeleton holds the same 300px (the react-chartkick default height). Measured layout shift with the chart chunk delayed: 0.030 before, 0 after.

The other async components get their data as server props, fetch on click, or show the result in a modal. They do not shift the page.

`aria-busy` on the wrapper, `aria-hidden` on the block, no shimmer under `prefers-reduced-motion: reduce`. No new i18n keys. Includes specs and a story.

## Corresponding Issue

Closes #2484

# Screenshots

Loading before this change, loading after, and the loaded chart.


julianguyen:
Thanks for improving this!

